PR #8, opened 2019-07-24
User Menu component

### LINKED ISSUE #4: Component: User Menu
Main UI is an avatars (using Gravatar, with https://ui-avatars.com/ or somehing similar as a fallback)

When clicking it, opens something dropdown-ish containing:

- Login/Logout (just show one for now and keep the other one hidden - we'll update that once the login code is ready)
- Your full name/email?

---

Or do we want to show something completely different while not logged-in? Probably something to see later; for now having a Login action in there is the easiest option.

mic4ael:
closes #4


newdle/client/package.json
```diff
@@ -3,11 +3,17 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "chai": "^4.2.0",
+    "enzyme": "^3.10.0",
```
ThiefMaster: that's pretty certainly a dev dependency

newdle/client/src/UserMenu.jsx
```diff
@@ -0,0 +1,59 @@
...
+  '8b5d5d', // color
+  2, // length
+  0.5, // font size
+  true, // rounded
+];
+
+export default function UserMenu({user}) {
+  const {email, fullName} = user;
+  const gravatarFallback = fullName
+    ? encodeURI(`//ui-avatars.com/api/${fullName}/${defaultUIAvatarsArgs.join('/')}`)
```
ThiefMaster: I think it's enough to pass the first character of the first/last name here; No need to expose more data than needed to an external service.

newdle/client/package.json
mic4ael: Yeah, indeed. I was to do it when cleaning up the code ;)

newdle/client/src/UserMenu.jsx
mic4ael: Yep, definitely ;)

newdle/client/package.json
```diff
@@ -5,13 +5,19 @@
...
-    "react-scripts": "3.0.1"
+    "react-gravatar": "^2.6.3",
+    "react-scripts": "3.0.1",
+    "semantic-ui-react": "^0.87.3"
   },
   "devDependencies": {
     "flask-urls.macro": "0.0.1",
-    "node-sass": "^4.12.0"
+    "node-sass": "^4.12.0",
+    "chai": "^4.2.0",
```
ThiefMaster: why do we need to explicitly depend on it? i can't see it referenced anywhere in our code. is it a peer dependency?

newdle/client/src/UserMenu.jsx
```diff
@@ -0,0 +1,66 @@
...
+  );
+}
+
+UserMenu.propTypes = {
+  user: PropTypes.shape({
+    email: PropTypes.string,
+    fullName: PropTypes.string,
+    firstName: PropTypes.string,
+    lastName: PropTypes.string,
+  }).isRequired,
```
ThiefMaster: Won't this fail for `user={null}`?

newdle/client/src/UserMenu.test.jsx
```diff
@@ -0,0 +1,26 @@
...
+  it('renders UserMenu successfully for anonymous users', () => {
+    const anonymous = mount(<UserMenu />);
+    expect(anonymous.contains(Icon)).to.equal(true);
+    expect(anonymous.exists('.user-gravatar')).to.equal(false);
+    anonymous.unmount();
+  });
+
+  it('renders UserMenu successfully for users with Gravatar', () => {
+    const userData = {email: 'test@test.email', fullName: 'T T', firstName: 'T', lastName: 'T'};
+    const anonymous = mount(<UserMenu user={userData} />);
```
ThiefMaster: this isn't *anonymous* ;)

newdle/client/package.json
mic4ael: There is one reference. `UserMenu.test.js`

newdle/client/src/UserMenu.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
```
ThiefMaster: why the empty line here?

newdle/client/src/UserMenu.js
```diff
@@ -0,0 +1,64 @@
...
+        <span>{fullName || email}</span>{' '}
+        <Gravatar email={email} default={gravatarFallback} size={40} />
+      </div>
+    )
+  );
+}
+
+export default function UserMenu() {
+  const user = useSelector(getUserInfo);
+  const trigger = user ? renderGravatar(user) : <NoUserAvatar />;
```
ThiefMaster: I'd move this after all the hooks

newdle/client/src/UserMenu.js
```diff
@@ -0,0 +1,64 @@
...
+function renderGravatar({email, first_name: firstName, last_name: lastName, full_name: fullName}) {
+  const gravatarFallback =
+    firstName && lastName
+      ? encodeURI(
+          `//ui-avatars.com/api/${firstName[0]} ${lastName[0]}/${defaultUIAvatarsArgs.join('/')}`
+        )
+      : 'mp';
+
+  return (
+    email && (
```
ThiefMaster: `if (!email) return null;` at the beginning.. the fallback defined above isn't even used then.

However, I don't think we actually need it: If we have a user we always have an email as well.

PR #10, opened 2019-07-25
Add Navigator Component

### LINKED ISSUE #3: Component: Navigator
- Shows in which creation step you are
- Shows a title/description of that step

---

Could either be generic (props: steps, current, title, description) or specific to the steps we have (props: step). Generic is probably cleaner since that way the parent component handles the data shown in it.

panagiotappl:
closes #3


newdle/client/src/components/Navigator.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Container} from 'semantic-ui-react';
+
+import {Step, StepBody} from './Step.js';
```
mvidalgarcia: `.js` again

newdle/client/src/App.js
```diff
@@ -1,22 +1,19 @@
 import React from 'react';
 import {useSelector} from 'react-redux';
-import {Header, Modal} from 'semantic-ui-react';
+import {Modal} from 'semantic-ui-react';
 import AuthTest from './AuthTest';
 import {isRefreshingToken} from './selectors';
 import {useAuthentication} from './auth';
-import styles from './App.module.scss';
+
+import TopHeader from './components/TopHeader.js';
```
mvidalgarcia: You don't need the `.js`

newdle/client/src/components/Step.scss
```diff
@@ -0,0 +1,36 @@
...
+.step.current{
+    background: $purple;
+    color: white;
+    border: none;
+    border-radius: 0;
+    margin-bottom: 5px;
+    transform: translate(0, -5px);
+}
+
+.step-body{
```
mvidalgarcia: also here and a couple of times below

newdle/client/src/styles/palette.scss
```diff
@@ -1,2 +1,4 @@
 $coral: #ff8383;
 $pink: #ff92c3;
+$purple: #6101b3;
+$grey: grey;
```
mvidalgarcia: I wonder if we need this. IMO it's better to have all the colors in the same place. If you we go for this, I think we're missing `white` that's being used in other files.

newdle/client/src/components/TopHeader.js
```diff
@@ -0,0 +1,53 @@
...
+    description: 'Which will be presented as options to the participants',
+  },
+  {
+    number: 3,
+    title: 'Finalize',
+    description: 'Do something final here',
+  },
+];
+
+class TopHeader extends React.Component {
```
mvidalgarcia: Let's go for 100% functional components? :wink: I guess it's just a matter of preference but it's a good moment to decide cc/ @indico/development-team

ThiefMaster: yes, no classes please

newdle/client/src/styles/palette.scss
meluru: Yes, let's have them all in one place

newdle/client/src/components/TopHeader.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import {Grid, Header} from 'semantic-ui-react';
+
+import styles from '../App.module.scss';
+
+import Navigator from './Navigator.js';
+
+const steps = [
```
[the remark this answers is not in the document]
panagiotappl: How can I only render the active step's body card that way without having everything appear in the wrong place?
The three step numbers on top should always be rendered next to each other and then only one component with details should appear underneath them.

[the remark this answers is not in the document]
ThiefMaster: If you use the first version then you simply don't render `children` unless the `active` prop is set.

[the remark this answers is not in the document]
panagiotappl: I have to render the numbers no matter what, though. Should those number boxes be outside of Navigator?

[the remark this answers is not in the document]
ThiefMaster: ah damn, didn't think of those... maybe there's some react magic that can be used  to know the index...  anyway, the easiest option for now is just to have `n={1}`, `n={2}` and `n={3}` on the `<Navigator.Step />` components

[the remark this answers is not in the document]
ThiefMaster: actually CSS can display the numbers, no need to know them in the JS code at all: https://codepen.io/anon/pen/VojNrj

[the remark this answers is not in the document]
panagiotappl: Shouldn't the numbered boxes be rendered first and then whichever one is active also has its title/description part rendered? That has the logic of the numbered box and the title/description separately, kind of like how Tabs work.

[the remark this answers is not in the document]
ThiefMaster: you're right.. I just checked how SUI does it for their tabs and turns out they also use a mapping array instead of something nicer with react components..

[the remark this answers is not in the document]
ThiefMaster: Maybe let's stick with the array then and just get rid of the counter in there (since that is really not needed as the array index + 1 is already a good counter)...

newdle/client/src/styles/palette.scss
panagiotappl: Do you think I should declare all defaults in that file in the end?

newdle/client/src/components/Navigator.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
```
ThiefMaster: `npm i --save prop-types` inside newdle/client/ - it just works right now because prop-types is installed as a dependency of something; but if we use a library it should be in our package.json

newdle/client/src/components/Navigator.js
```diff
@@ -0,0 +1,36 @@
...
+import {Container} from 'semantic-ui-react';
+import {Step, StepBody} from './Step';
```
ThiefMaster: Just wondering.. do we need a separate file for this component? It's somewhat small and only useful within the context of `Navigator`, so I'd rather define it in the same file (and not export it)

newdle/client/src/components/Step.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Header} from 'semantic-ui-react';
+
+import styles from './Step.module.scss';
+
+export function Step({active, children}) {
+  return (
+    <>
```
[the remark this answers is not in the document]
panagiotappl: No, it was leftover after refactoring.

newdle/client/src/components/TopHeader.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import {Grid, Header} from 'semantic-ui-react';
+
+import styles from '../App.module.scss';
+
```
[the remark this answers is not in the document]
panagiotappl: Shouldn't the one between the module and the components stay?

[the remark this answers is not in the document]
ThiefMaster: no strong opinion... i got annoyed a few times by my editor adding new imports after the last one :D 

since we usually have only one css import maybe there's not much benefit in separating it from the rest?

ThiefMaster:
I think the navigator should only be shown on `/new`. It should be easily possible using the Router.

newdle/client/src/components/Navigator.js
```diff
@@ -0,0 +1,55 @@
...
+
+export default function Navigator({steps, active}) {
+  return (
+    <Container>
+      {steps.map((step, index) => (
+        <Step key={index} active={index + 1 === active}>
+          {index + 1}
+        </Step>
+      ))}
+      {steps.map((step, index) => {
```
ThiefMaster: This whole block can be replaced with `<StepBody title={steps[active-1].title} description={steps[active-1].description} />`

Or, to make it prettier, assign `const activeStep = steps[active - 1];` at the beginning of the function and then use `activeStep` in the snippet above

newdle/client/src/components/Navigator.js
```diff
@@ -0,0 +1,55 @@
...
+
+function Step({active, children}) {
+  return <div className={`${styles.step} ${active ? styles.active : ''}`}>{children}</div>;
+}
+
+Step.propTypes = {
+  active: PropTypes.bool,
+};
+
+export function StepBody({title, description}) {
```
ThiefMaster: no need to `export` it

newdle/client/src/components/TopHeader.js
```diff
@@ -0,0 +1,54 @@
...
+          <UserMenu />
+        </Grid.Column>
+        <Grid.Row centered>
+          <Grid.Column />
+          <Grid.Column>
+            <Switch>
+              <Route
+                exact
+                path="/new"
+                component={() => <Navigator steps={steps} active={active} />}
```
ThiefMaster: That's not the correct way to use `component`. It should be something like `component={Navigatorr}`, but since you need to specify props here you want `render={() => <Navigator .. />}`

newdle/client/src/components/Navigator.js
```diff
@@ -0,0 +1,55 @@
...
+  ),
+  active: PropTypes.number,
+};
+
+function Step({active, children}) {
+  return <div className={`${styles.step} ${active ? styles.active : ''}`}>{children}</div>;
+}
+
+Step.propTypes = {
+  active: PropTypes.bool,
```
ThiefMaster: `children` is missing in the propTypes

newdle/client/src/styles/palette.scss
mvidalgarcia: I'd declare in the palette the ones that we use. If later on we decide to change e.g the grey, we just have to modify this file.

PR #15, opened 2019-07-25
Add create newdle page layout

### LINKED ISSUE #11: Component: Create Newdle
This is the page for creating a new...dle. While we don't have the components yet that should go there, we can put placeholders for them and setup the CSS to arrange everything nicely.

mvidalgarcia:
close #11

### DIFF
--- a/newdle/client/src/App.js
+++ b/newdle/client/src/App.js
@@ -4,7 +4,6 @@ import {Modal} from 'semantic-ui-react';
 import {BrowserRouter as Router, Route, Switch} from 'react-router-dom';
 import Home from './Home';
 import CreateNewdle from './CreateNewdle';
-import TimelineExample from './TimelineExample';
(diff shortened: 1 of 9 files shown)


panagiotappl:
Very nice PR :clap: 

Just one comment about the spacing between elements in the `TimeSlotsPage`.
Would it be better if the elements were a bit "closer" together? Having the timeline a bit more centered and the calendar closer to and not on the left side of the page. 

Example below:
createnewdle-new

mvidalgarcia:
> Very nice PR 
> 
> Just one comment about the spacing between elements in the `TimeSlotsPage`.
> Would it be better if the elements were a bit "closer" together? Having the timeline a bit more centered and the calendar closer to and not on the left side of the page.

Thanks!
Totally agree! Doesn't look so pretty on big screens

PR #14, opened 2019-07-25
Add user search API

ThiefMaster:
Return dummy data for now. Name and email should be enough - I don't think affiliation is needed here.

Example code I used to test it (no need for redux in the real user search code later, except when adding someone to the participants):

```diff
diff --git a/newdle/client/src/CreateNewdle.js b/newdle/client/src/CreateNewdle.js
index a1cc87e..c87bc0b 100644
--- a/newdle/client/src/CreateNewdle.js
+++ b/newdle/client/src/CreateNewdle.js
@@ -1,5 +1,36 @@
-import React from 'react';
+import React, {useState} from 'react';
+import {Input} from 'semantic-ui-react';
+import client from './client';

 export default function CreateNewdle() {
-  return <h2>New newdle</h2>;
+  const [query, setQuery] = useState('');
+  const [searching, setSearching] = useState(false);
+
+  const handleSearch = async evt => {
+    evt.preventDefault();
+    setSearching(true);
+    let result;
+    try {
+      result = await client.searchUsers(query);
+    } catch (err) {
+      console.log(err);
+      console.log(err.data);
+    }
+    setSearching(false);
+    console.log(result);
+  };
+
+  return (
+    <>
+      <h2>New newdle</h2>
+      <form onSubmit={handleSearch}>
+        <Input
+          icon="search"
+          placeholder="Search user..."
+          onChange={evt => setQuery(evt.target.value)}
+          disabled={searching}
+        />
+      </form>
+    </>
+  );
 }
```

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -1,8 +1,11 @@
+from faker import Faker
 from flask import Blueprint, g, jsonify, request
 from itsdangerous import BadData, SignatureExpired
+from marshmallow import fields
 
 from .core.auth import user_info_from_app_token
-from .schemas import UserSchema
+from .core.webargs import use_kwargs
+from .schemas import UserSchema, UserSearchResultSchema
 
 
 api = Blueprint('api', __name__, url_prefix='/api')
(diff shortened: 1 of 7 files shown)


newdle/api.py
```diff
@@ -29,3 +41,40 @@ def require_token():
...
+    def _generate_fake_user():
+        first_name = f.first_name()
+        last_name = f.last_name()
+        email = f'{first_name}.{last_name}@{f.domain_name()}'.lower()
+        return {'first_name': first_name, 'last_name': last_name, 'email': email}
+
+    return [_generate_fake_user() for _ in range(100)] + [
+        {
+            'first_name': g.user['first_name'],
+            'last_name': g.user['last_name'],
```
lepture: I think `first_name` and `last_name` are bad naming. In Asia countries, their family name come at first.

ThiefMaster: While our userbase is *mostly* western, that's a valid point - luckily our UI uses primarily the `name` for display purposes, which is currently built from first+last in the schema but could just as well be taken directly from whatever "name" the SSO provider sends (but I have the feeling over there it's also just concatenating first and last name, since at least in our case that data is coming from the HR database where I'm pretty sure it's already split into first and last name)

newdle/schemas.py
```diff
@@ -8,3 +8,14 @@ class UserSchema(mm.Schema):
     name = fields.Function(lambda u: f'{u["first_name"]} {u["last_name"]}')
     initials = fields.Function(lambda u: f'{u["first_name"][0]} {u["last_name"][0]}')
     uid = fields.String()
+
+
+class UserSearchResultSchema(UserSchema):
+    class Meta:
+        fields = ('email', 'name', 'initials')
```
mic4ael: Since you added this here, maybe you can use that in the `UserMenu` component?

ThiefMaster: 982b14c ;)

mic4ael: Ohh, didn't notice that one :P

PR #16, opened 2019-07-26
Calendar component

### LINKED ISSUE #13: Component: Calendar
This is the calendar (`react-dates`?) where you select a day on which to see participant availability & add time slots.

Days that have time slots should be highlighted so it's clearly visible which days are used. Removing the last time slot from a day should remove this highlight.

The active day should be highlighted differently regardless of whether it has time slots or not.

mic4ael:
closes #13

### DIFF
--- a/newdle/client/src/Calendar.js
+++ b/newdle/client/src/Calendar.js
@@ -1,0 +1,46 @@
+import 'react-dates/initialize';
+import React, {useCallback, useEffect} from 'react';
+import {HTML5_FMT} from 'moment';
+import {useDispatch, useSelector} from 'react-redux';
+import {DayPickerSingleDateController as DayPicker} from 'react-dates';
+import {getCalendarDates, getCalendarActiveDate} from './selectors';
+import {setActiveDate, clearCalendarActiveDate} from './actions';
+import {serializeDate, toMoment} from './util/date';
+
+import 'react-dates/lib/css/_datepicker.css';
+import styles from './Calendar.module.scss';
+
+export default function Calendar() {
+  const calendarDates = useSelector(getCalendarDates);
+  const activeDate = useSelector(getCalendarActiveDate);
+  const dispatch = useDispatch();
+
+  const handleDateChange = useCallback(date => dispatch(setActiveDate(serializeDate(date))), [
+    dispatch,
+  ]);
+
+  const isDayHighlighted = useCallback(date => calendarDates.includes(serializeDate(date)), [
+    calendarDates,
+  ]);
+
+  useEffect(() => {
+    return () => {
+      dispatch(clearCalendarActiveDate());
+    };
+  }, [dispatch]);
+
+  return (
+    <div className={styles.calendar}>
+      <DayPicker
+        date={toMoment(activeDate, HTML5_FMT.DATE)}
+        firstDayOfWeek={1}
+        onDateChange={handleDateChange}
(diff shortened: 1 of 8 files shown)


newdle/client/src/actions.js
```diff
@@ -27,3 +30,15 @@ export function loadUser() {
 export function tokenExpired() {
   return {type: TOKEN_EXPIRED};
 }
+
+export function setActiveDate(date) {
+  return {type: SET_ACTIVE_DATE, date};
+}
+
+export function setCalendarDate(dates) {
```
ThiefMaster: Isn't this implicit (ie something for a selector) based on which dates have time slots?

newdle/client/src/Calendar.js
```diff
@@ -0,0 +1,46 @@
...
+
+import 'react-dates/lib/css/_datepicker.css';
+import styles from './Calendar.module.scss';
+
+export default function Calendar() {
+  const calendarDates = useSelector(getCalendarDates);
+  const activeDate = useSelector(getCalendarActiveDate);
+  const dispatch = useDispatch();
+
+  const onDateChange = useCallback(date => dispatch(setActiveDate(serializeDate(date))), [
```
ThiefMaster: If this data is only used inside this component, does it really belong in the redux store?

I'd only put things in the store that have to persist between unmounts of whatever component deals with the data. So for example, the participants, time slots and duration, since switching between steps shouldn't lose that data.

mic4ael: I assumed this will be somehow shared between the `Timeline` component and `Calendar` hence I put it in the store. I might be missing some details since I wasn't there during the design stage.

newdle/client/src/actions.js
mic4ael: I wasn't aware of the specific data model thus I assumed we will do it somehow manually.

newdle/client/src/Calendar.js
mic4ael: And apart from that, I think that selecting a date will have impact on the `Timeline` component.

ThiefMaster: Yes you are right, the timeline needs to know the selected date.

newdle/client/src/actions.js
ThiefMaster: from the notes in the codimd ;)

> Calendar highlights days which have time slots - removing the last time slot removes the whole (empty) day;

and in the issue:

> Days that have time slots should be highlighted so it's clearly visible which days are used. Removing the last time slot from a day should remove this highlight.

By not having this data in the store the logic remains easier since we just need a selector that returns the dates which have time slots

mic4ael: Hm, then for now I will just create a selector that will return an empty array. And as soon as we add more stuff we will adapt it.

ThiefMaster: or some random dates in the current month so it looks better until we have it?

mic4ael: Yeah, fine as well ;)

newdle/client/src/Calendar.js
ThiefMaster: i think the standard naming style for event handlers in react is `handle...`, i.e. `handleDateChange`

mic4ael:
> Could we show a different highlight if the day is both selected and has time slots?

Yeah, it is just a matter of adding one selector.

Issue #6, opened 2019-07-24
Component: Timeline


ThiefMaster:
Displays the timeline listing all participants and their availability. Each participant will be another component (Participant) that's shown inside the Timeline.

PR #17, opened 2019-07-26
Timeline

meluru:
Closes: #6


newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,148 @@
...
+        {date.format('D MMM YYYY')}
+      </Header>
+      {renderTimelineHeader(hourSeries, hourSpan, hourStep)}
+      {renderTimelineContent(busySlots)}
+    </div>
+  );
+}
+
+Timeline.propTypes = {
+  date: PropTypes.object.isRequired,
```
ThiefMaster: momentjs object I guess? I think it'd be nicer to pass around strings and only convert to moment internally when needed.

newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,148 @@
...
+  );
+}
+
+function renderBusySlot({width, pos}, key) {
+  return (
+    <div
+      className={styles['busy-item']}
+      key={`timeline-busy-${key}`}
+      style={{left: `${pos}%`, width: `${width}%`}}
+    ></div>
```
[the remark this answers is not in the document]
meluru: It didn't, but yeah, that would be much better

newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,145 @@
+import React from 'react';
+import Participant from './Participant';
```
ThiefMaster: let's keep relative imports after non-relative ones (ie stuff coming from node_modules)? no empty line between, just some sorting

newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,145 @@
...
+export default function Timeline({date, availability, minHour, maxHour, hourStep}) {
+  const hourSeries = _.range(minHour, maxHour + hourStep, hourStep);
+  const hourSpan = maxHour - minHour;
+  const busySlots = calculateBusyPositions(availability, minHour, maxHour);
+  return (
+    <div className={styles['timeline']}>
+      <Header as="h2" className={styles['timeline-date']}>
+        {moment(date, 'YYYY-MM-DD').format('D MMM YYYY')}
+      </Header>
+      {renderTimelineHeader(hourSeries, hourSpan, hourStep)}
```
ThiefMaster: I think some of these `render` functions could rather be components themselves. This would keep things more structured and easier to cover with tests. For example, I'd use a `TimelineRow` component for each row.

newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,145 @@
...
+}
+
+function calculatePosition(start, minHour, maxHour) {
+  const spanMins = (maxHour - minHour) * 60;
+  let startMins = start.hours() * 60 + start.minutes() - minHour * 60;
+
+  if (startMins < 0) {
+    startMins = 0;
+  } else if (startMins >= spanMins) {
+    startMins = spanMins - 5;
```
mic4ael: Why are you subtracting 5 here? Is it because of some margin/padding?

newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,145 @@
...
+      participant,
+      busySlots: slots,
+    };
+  });
+}
+
+function renderParticipant(participant) {
+  return (
+    <span className={styles['timeline-row-label']}>
+      {<Participant name={`${participant.firstName} ${participant.lastName}`} />}
```
mic4ael: Unnecessary outer `{}`?

newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,145 @@
...
+        {busySlots.map((slot, i) => renderBusySlot(slot, i))}
+      </div>
+    </div>
+  );
+}
+
+function renderTimelineContent(availability) {
+  return (
+    <div className={styles['timeline-rows']}>
+      {availability.map((avail, i) => renderTimelineRow(avail, i))}
```
mic4ael: `availability.map(renderTimelineRow)`

newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,145 @@
...
+      {availability.map(participant => {
```
mic4ael:
```suggestion
      {availability.map(participant => participant.busySlots.map(slot => renderBusyColumn(slot)))}
```

newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,145 @@
...
+function renderBusyColumn({width, pos}) {
+  return <div className={styles['busy-column']} style={{left: `${pos}%`, width: `${width}%`}} />;
+}
+
+function renderTimelineRow({participant, busySlots}, key) {
+  return (
+    <div className={styles['timeline-row']} key={`row-${key}`}>
+      {renderParticipant(participant)}
+      <div className={styles['timeline-busy']}>
+        {busySlots.map((slot, i) => renderBusySlot(slot, i))}
```
mic4ael: `{busySlots.map(renderBusySlot)}`

newdle/client/src/Timeline.js
```diff
@@ -0,0 +1,145 @@
...
+    <div
+      className={styles['busy-item']}
+      key={`timeline-busy-${key}`}
+      style={{left: `${pos}%`, width: `${width}%`}}
+    />
+  );
+}
+
+function renderBusyColumn({width, pos}) {
+  return <div className={styles['busy-column']} style={{left: `${pos}%`, width: `${width}%`}} />;
```
pferreir: There's a missing `key=` here:

I tried this out and it seems to work:
```js
function renderBusyColumn({ width, pos, startTime, endTime }) {
  return <div
    key={`${startTime}-${endTime}`}
    className={styles['busy-column']}
    style={{ left: `${pos}%`, width: `${width}%` }}
  />;
}
```

newdle/client/src/Timeline.js
meluru: It's the case when start time is later than the end time of the timeline. We are subtracting 5 from end time so that the row is visible

mic4ael: Ok, but does it have to be 5?

meluru: I understand your point now, I've changed the way we render the overflowing slots

newdle/client/src/Timeline.js
meluru: Yes, thanks for pointing out. :) I took your idea and added participant id as well, since columns from different participants can have the same times.

i still get the warnings in the console though

meluru: Solved

newdle/client/src/components/Timeline.js
```diff
@@ -0,0 +1,169 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
+import {Header} from 'semantic-ui-react';
+import styles from './Timeline.module.scss';
```
mvidalgarcia: We normally place `.scss` imports at the last

newdle/client/src/components/Timeline.js
```diff
@@ -0,0 +1,169 @@
...
+import moment from 'moment';
+import Participant from './Participant';
+
+const overflowWidth = 0.5;
```
mvidalgarcia: What about using `OVERFLOW_WIDTH` since it's a constant?

newdle/client/src/components/Timeline.js
```diff
@@ -0,0 +1,169 @@
...
+          </span>
+        </div>
+      ))}
+    </div>
+  );
+}
+
+TimelineHeader.propTypes = {
+  hourSeries: PropTypes.array.isRequired,
+  hourSpan: PropTypes.number,
```
mvidalgarcia: I think you're missing some `defaultProps` here.

newdle/client/src/TimelineExample.js
```diff
@@ -0,0 +1,56 @@
...
+    {
+      participant: {
+        id: 2,
+        lastName: 'Person',
+        firstName: 'Second',
+      },
+      busySlots: [
+        {
+          startTime: '14:00',
+          endTime: '15:00:',
```
mvidalgarcia: Extra colon?

PR #23, opened 2019-08-01
Add DurationPicker

### LINKED ISSUE #12: Component: Duration Picker
Displays the field for the duration with some text around it.

panagiotappl:
closes #12

### DIFF
--- a/newdle/client/src/components/DurationPicker.js
+++ b/newdle/client/src/components/DurationPicker.js
@@ -1,7 +1,10 @@
-import React, {useState} from 'react';
+import React from 'react';
+import {useDispatch, useSelector} from 'react-redux';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
 import {Dropdown} from 'semantic-ui-react';
+import {setDuration} from '../actions';
+import {getDuration} from '../selectors';
 import styles from './DurationPicker.module.scss';
 
 function getDurationOptions(min, max, interval) {
@@ -9,8 +12,9 @@ function getDurationOptions(min, max, interval) {
 }
 
 export default function DurationPicker({minDuration, maxDuration, interval}) {
-  const [durationValue, setDurationValue] = useState(minDuration);
+  const duration = useSelector(getDuration);
   const durationOptions = getDurationOptions(minDuration, maxDuration, interval);
+  const dispatch = useDispatch();
 
   return (
     <div className={styles['duration-picker']}>
@@ -19,8 +23,8 @@ export default function DurationPicker({minDuration, maxDuration, interval}) {
         className={styles['dropdown']}
         options={durationOptions}
         selection
-        value={durationValue}
-        onChange={(_, {value}) => setDurationValue(value)}
+        value={duration}
+        onChange={(_, {value}) => dispatch(setDuration(value))}
       />
       <span> minutes</span>
     </div>
--- a/newdle/client/src/components/DurationPicker.js
+++ b/newdle/client/src/components/DurationPicker.js
@@ -1,0 +1,40 @@
+import React, {useState} from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
+import {Dropdown} from 'semantic-ui-react';
+import styles from './DurationPicker.module.scss';
+
+function getDurationOptions(min, max, interval) {
+  return _.range(min, max + interval, interval).map(i => ({key: i, value: i, text: i}));
+}
+
+export default function DurationPicker({minDuration, maxDuration, interval}) {
+  const [durationValue, setDurationValue] = useState(minDuration);
+  const durationOptions = getDurationOptions(minDuration, maxDuration, interval);
+
+  return (
+    <div className={styles['duration-picker']}>
+      <span>Meeting time: </span>
+      <Dropdown
+        className={styles['dropdown']}
+        options={durationOptions}
+        selection
+        value={durationValue}
+        onChange={(_, {value}) => setDurationValue(value)}
+      />
+      <span> minutes</span>
+    </div>
+  );
+}
+
+DurationPicker.propTypes = {
+  minDuration: PropTypes.number,
+  maxDuration: PropTypes.number,
+  interval: PropTypes.number,
+};
+
+DurationPicker.defaultProps = {
+  minDuration: 15,
+  maxDuration: 90,
+  interval: 15,
+};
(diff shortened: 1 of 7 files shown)


newdle/client/src/components/DurationPicker.module.scss
```diff
@@ -0,0 +1,13 @@
+@import '../styles/palette.scss';
+
+.duration-picker {
+  background-color: $pink;
+  display: inline-block;
+  padding: 5px;
+  color: white;
```
mvidalgarcia: `$white`

newdle/client/src/components/DurationPicker.js
```diff
@@ -0,0 +1,34 @@
+import React, {useState} from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
+import {Dropdown} from 'semantic-ui-react';
+import styles from './DurationPicker.module.scss';
+
+function renderDurationOptions(min, max, interval) {
```
ThiefMaster: I don't think `render` in the function name makes much sense here

panagiotappl: True. Would a name like `calculate` be more appropriate?

ThiefMaster: not a fan of `calculate` unless math is involved. `getOptions` or `buildOptions` maybe?

panagiotappl: I'll go with `getDurationOptions` then. Sounds better indeed.

newdle/client/src/components/DurationPicker.js
```diff
@@ -0,0 +1,34 @@
...
+      />
+      <span> minutes</span>
+    </div>
+  );
+}
+
+DurationPicker.propTypes = {
+  minDuration: PropTypes.number.isRequired,
+  maxDuration: PropTypes.number.isRequired,
+  interval: PropTypes.number.isRequired,
```
meluru: Not a must, but maybe those props could have some default values? Most likely we will use it with the same props everywhere.

panagiotappl: Good idea. Since I'm kind of hardcoding them as I'm passing them from the parent anyway.

newdle/client/src/components/DurationPicker.module.scss
```diff
@@ -0,0 +1,13 @@
...
+
+.duration-picker {
+  background-color: $pink;
+  display: inline-block;
+  padding: 5px;
+  color: $white;
+
+  .dropdown {
+    min-width: 0 !important;
+    border-radius: 0 !important;
```
pferreir: Are you sure there's no alternative to these `!important`?

newdle/client/src/components/DurationPicker.js
```diff
@@ -0,0 +1,40 @@
...
+import {Dropdown} from 'semantic-ui-react';
+import styles from './DurationPicker.module.scss';
+
+function getDurationOptions(min, max, interval) {
+  return _.range(min, max + interval, interval).map(i => ({key: i, value: i, text: i}));
+}
+
+export default function DurationPicker({minDuration, maxDuration, interval}) {
+  const [durationValue, setDurationValue] = useState(minDuration);
+  const durationOptions = getDurationOptions(minDuration, maxDuration, interval);
```
pferreir: This should probably be a fully controlled component. The state is not of much use inside the current component as we will have to use it further into the process.

newdle/client/src/components/DurationPicker.module.scss
panagiotappl: Do we need an alternative in this case? It's only class that's going to be used inside the duration-picker.

newdle/client/src/components/DurationPicker.js
ThiefMaster: Indeed! IIRC. I didn't comment on that one since until we integrate it with the rest having local state is probably easier.

panagiotappl: The `Timeline` is already merged and since this will be integrated with the `Timeline` I can change it now.

newdle/client/src/components/TopHeader.module.scss
```diff
@@ -1,7 +1,7 @@
 @import '../styles/palette.scss';
 
 .header {
-  background: linear-gradient($coral, $pink);
+  background: $coral;
```
mvidalgarcia: Why have you removed the gradient?

newdle/core/util.py
```diff
@@ -1,11 +1,12 @@
 from datetime import datetime
 
-import pytz
 from sqlalchemy import func, types
 from sqlalchemy.sql import operators
 from sqlalchemy.sql.sqltypes import Interval
 from sqlalchemy.util import memoized_property
 
+import pytz
```
mvidalgarcia: What's the reason of this change? I think it was OK.

newdle/models.py
```diff
@@ -1,9 +1,9 @@
-from pytz import timezone, utc
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.schema import CheckConstraint
 
 from newdle.core.db import db
 from newdle.core.util import UTCDateTime, format_dt, parse_dt
+from pytz import timezone, utc
```
mvidalgarcia: And this? `black` messing around?

newdle/schemas.py
```diff
@@ -1,4 +1,5 @@
 from marshmallow import fields, post_dump
+
```
mvidalgarcia: Same.

newdle/client/src/components/Timeline.js
```diff
@@ -143,9 +144,14 @@ export default function Timeline({date, availability, minHour, maxHour, hourStep
...
   return (
     <div className={styles['timeline']}>
-      <Header as="h2" className={styles['timeline-date']}>
-        {moment(date, 'YYYY-MM-DD').format('D MMM YYYY')}
-      </Header>
+      <div className={styles['timeline-title']}>
+        <Header as="h2" className={styles['timeline-date']}>
+          {moment(date, 'YYYY-MM-DD').format('D MMM YYYY')}
+        </Header>
+        <div>
```
mvidalgarcia: I think you can get rid of this `div`

newdle/core/util.py
ThiefMaster: That looks like someone ran isort without the package being properly installed

newdle/models.py
ThiefMaster: same here (isort, not black)

Issue #22, opened 2019-08-01
Add newdle creation API


pferreir:
Basically something like `POST /newdle/new`, which creates a new newdle with given participants (if any).

ThiefMaster:
or just `POST /api/newdle/` since the verb implies creation ;)

pferreir:
> or just `POST /api/newdle/` since the verb implies creation ;)

Yeah, that makes more sense.

PR #24, opened 2019-08-01
Add DB models and tests, POST for newdle creation

### LINKED ISSUE #19: Add database models
- `Newdle` (or some more generic name): stores information such as creator, duration, timezone and final date/time. Time slots can probably be stored in JSONB.
- `Participant`: linked to a `Newdle`, stores name etc. and has a JSONB column for the user's answers.

---

Why JSONB for time slots / answers?

- Modifying a Newdle is not planned for now, and it would be an atomic operation anyway. So there is no need to edit/add/remove individual time slots.
- When a user answers, the availability is for a given date/time (which would not change even if that time slot was moved), so storing it as such is more natural than a time slot ID.

pferreir:
This closes #19 and #22 and adds as well the possibility to create Python unit tests.

When it comes to testing, you should have an empty `newdle_tests` DB in your machine. It would be great if we could reuse https://github.com/indico/indico/blob/master/indico/testing/fixtures/database.py and https://github.com/indico/indico/blob/master/indico/core/db/sqlalchemy/util/management.py, maybe moving them to a 3rd party lib?

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,41 @@
+import os
+import pytest
+
+from newdle.core.app import create_app
+from newdle.core.db import db
+
+
+TEST_DATABASE_URI = 'postgresql:///newdle_tests'
+
+
+@pytest.fixture(scope='session')
+def app(request):
+    """Session-wide test `Flask` application."""
+    config_override = {'TESTING': True, 'SQLALCHEMY_DATABASE_URI': TEST_DATABASE_URI}
+    app = create_app(config_override, use_env_config=False)
+    ctx = app.app_context()
+    ctx.push()
+    yield app
+    ctx.pop()
+
+
+@pytest.fixture(scope='function')
+def db_session(app, request):
+    """Create a new database session."""
+    connection = db.engine.connect()
+    transaction = connection.begin()
+
+    options = dict(bind=connection, binds={})
+    session = db.create_scoped_session(options=options)
+
+    db.session = session
+    db.app = app
+    db.create_all()
+
+    def teardown():
+        transaction.rollback()
+        connection.close()
+        session.remove()
+
+    request.addfinalizer(teardown)
+    return session
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -5,8 +5,10 @@ from marshmallow import fields
 from werkzeug.exceptions import UnprocessableEntity
 
 from .core.auth import user_info_from_app_token
+from .core.db import db
 from .core.webargs import use_kwargs
-from .schemas import UserSchema, UserSearchResultSchema
+from .models import Newdle, Participant
+from .schemas import NewdleSchema, NewNewdleSchema, UserSchema, UserSearchResultSchema
 
 
 api = Blueprint('api', __name__, url_prefix='/api')
@@ -78,3 +80,19 @@ def users(q):
         'total': len(data),
         'users': UserSearchResultSchema(many=True).dump(data[:10]),
     }
+
+
+@api.route('/newdle/', methods=('POST',))
+@use_kwargs(NewNewdleSchema(), locations=('json',))
+def newdle(title, duration, timezone, time_slots, participants):
+    newdle = Newdle(
+        title=title,
+        creator_uid=g.user['uid'],
+        duration=duration,
+        timezone=timezone,
+        time_slots=time_slots,
+        participants={Participant(**p, newdle=newdle) for p in participants},
+    )
+    db.session.add(newdle)
+    db.session.commit()
+    return NewdleSchema().jsonify(newdle)
--- a/newdle/models.py
+++ b/newdle/models.py
@@ -1,0 +1,76 @@
+from pytz import timezone, utc
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.schema import CheckConstraint
+
+from newdle.core.db import db
+from newdle.core.util import UTCDateTime, format_dt, parse_dt
+
+
+class Newdle(db.Model):
+    __tablename__ = 'newdles'
+
+    id = db.Column(db.Integer, primary_key=True)
+    creator_uid = db.Column(db.String, nullable=False, index=True)
+    title = db.Column(db.String, nullable=False)
+    duration = db.Column(db.Integer, nullable=False)
+    _timezone = db.Column('timezone', db.String, nullable=False)
+    _time_slots = db.Column('time_slots', JSONB, nullable=False)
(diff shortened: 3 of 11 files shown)


requirements.txt
```diff
@@ -12,3 +12,6 @@ marshmallow==3.0.0rc8
 marshmallow-sqlalchemy==0.17.0
 faker==2.0.0
 webargs==5.4.0
+pytz==2019.2
+pytest==5.0.1
```
ThiefMaster: those are dev requirements

pferreir: true!

conftest.py
```diff
@@ -0,0 +1,45 @@
...
+    db.session = session
+    db.app = app
+    db.create_all()
+
+    def teardown():
+        transaction.rollback()
+        connection.close()
+        session.remove()
+
+    request.addfinalizer(teardown)
```
ThiefMaster: why not use a yield-style fixture?

newdle/api.py
```diff
@@ -78,3 +80,20 @@ def users(q):
...
+        title=title,
+        creator_uid=g.user['uid'],
+        duration=duration,
+        timezone=timezone,
+        time_slots=time_slots,
+        participants={Participant(**p, newdle=newdle) for p in participants},
+    )
+    db.session.add(newdle)
+    db.session.commit()
+    schema = NewdleSchema()
```
ThiefMaster: return NewdleSchema().jsonify(newdle)

newdle/models.py
```diff
@@ -0,0 +1,79 @@
...
+
+    id = db.Column(db.Integer, primary_key=True)
+    creator_uid = db.Column(db.String, nullable=False, index=True)
+    title = db.Column(db.String, nullable=False)
+    duration = db.Column(db.Integer, nullable=False)
+    _timezone = db.Column('timezone', db.String, nullable=False)
+    _time_slots = db.Column('time_slots', JSONB, nullable=False)
+    final_dt = db.Column(UTCDateTime, nullable=True)
+
+    # relationship backrefs:
```
ThiefMaster: what about using explicit relationships on both sites (using `back_populates`) instead?

newdle/models.py
```diff
@@ -0,0 +1,79 @@
...
+        ]
+
+    def __repr__(self):
+        return "<Newdle {} {}>".format(self.id, 'F' if self.final_dt else '')
+
+
+class Participant(db.Model):
+    __tablename__ = 'participants'
+    __table_args__ = (
+        CheckConstraint('email IS NULL OR auth_uid IS NULL', 'email_or_uid_null'),
```
ThiefMaster: don't you want xor logic so one of them has to be set?

newdle/schemas.py
```diff
@@ -19,3 +21,30 @@ def sort_users(self, data, many, **kwargs):
...
+class SlotSchema(mm.Schema):
+    start = fields.DateTime(required=True, format=DATETIME_FORMAT)
+    end = fields.DateTime(required=True, format=DATETIME_FORMAT)
+
+
+class NewNewdleSchema(mm.Schema):
+    title = fields.String(validate=lambda x: len(x) >= 3, required=True)
+    duration = fields.Int(required=True, validate=lambda x: x % 15 == 0)
+    timezone = fields.String(
+        validate=lambda x: x in common_timezones_set, required=True
```
ThiefMaster: `validate=common_timezones_set.__contains__`? :D (maybe a bit too magic though)

newdle/schemas.py
```diff
@@ -19,3 +21,30 @@ def sort_users(self, data, many, **kwargs):
...
+    )
+    time_slots = fields.List(
+        fields.Nested(SlotSchema), validate=lambda x: len(x) > 0, required=True
```
ThiefMaster: `validate=bool` should do the job

pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+norecursedirs =
```
ThiefMaster: the client folder can probably excluded as well

requirements.txt
```diff
@@ -12,3 +12,4 @@ marshmallow==3.0.0rc8
 marshmallow-sqlalchemy==0.17.0
 faker==2.0.0
 webargs==5.4.0
+pytz==2019.2
```
ThiefMaster: pytz version shouldn't be pinned

newdle/schemas.py
pferreir: argh. no. I really don't see a problem with using lambda.

tests/models_test.py
```diff
@@ -0,0 +1,68 @@
...
+    db_session.flush()
+
+    newdle = Newdle.query.get(dummy_newdle.id)
+    assert newdle.time_slots == [{
+        'start': utc.localize(datetime(2012, 11, 20, 10, 00)),
+        'end': utc.localize(datetime(2012, 11, 20, 11, 00))
+    }]
+
+def test_participant_email_uid(dummy_newdle, db_session):
+    participant = Participant(name="John Doe", email='john@example.com', newdle=dummy_newdle)
```
mic4ael: out of curiosity, why mixing `"` with `'`? :P

newdle/schemas.py
pferreir: (it's also more readable)

ThiefMaster: yeah, i didn't expect anyone to change this one :D

newdle/models.py
pferreir: Had forgotten that exists, thanks!

newdle/models.py
pferreir: That was my initial approach, but then I thought about it and the truth table is something like

|---+---+---|
| email | auth_uid |  |
|----|----|----|
|0|0| Ad-hoc user |
|0|1| Directory user |
|1|0| Ad-hoc user with e-mail |
|1|1| Disallowed |

Actually, now that I think of it, should we allow for the third case at all? And how?

conftest.py
pferreir: This was sto... borrowed from an example I found online. I'm fixing it.

tests/models_test.py
pferreir: Just happened. I'll fix it.

newdle/models.py
pferreir: OK, I've just changed it to `(email IS NULL) = (auth_uid IS NULL)`. I think it makes more sense given the use cases we're targeting right now (either ad-hoc or a directory user, which includes the e-mail).

PR #25, opened 2019-08-02
User search

### LINKED ISSUE #18: Component: User Search
We can probably use a simplified version of the search dialog we have in Indico.

mic4ael:
closes #18

### DIFF
--- a/newdle/client/src/components/UserAvatar.js
+++ b/newdle/client/src/components/UserAvatar.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Gravatar from 'react-gravatar';
+
+const DEFAULT_AVATAR_SIZE = 40;
+const defaultArgs = [
+  'f0e9e9', // background
+  '8b5d5d', // color
+  2, // length
+  0.5, // font size
+  true, // rounded
+].join('/');
+
+function UserAvatar({user: {email, name, initials}, className, withLabel, size}) {
+  initials = initials.replace(/[/?#]/g, '');
+  const uri = `https://ui-avatars.com/api/${initials}/${size}/${defaultArgs}`;
+
+  return (
+    <div className={className}>
+      {withLabel && <span>{name}</span>}{' '}
+      <Gravatar email={email} default={encodeURI(uri)} size={size} />
+    </div>
+  );
+}
+
+UserAvatar.propTypes = {
+  user: PropTypes.object.isRequired,
+  withLabel: PropTypes.bool,
+  size: PropTypes.number,
+};
+
+UserAvatar.defaultProps = {
+  withLabel: false,
+  size: DEFAULT_AVATAR_SIZE,
+};
+
+export default React.memo(UserAvatar);
--- a/newdle/client/src/components/UserSearch.js
+++ b/newdle/client/src/components/UserSearch.js
@@ -1,8 +1,11 @@
 import React, {useCallback, useState} from 'react';
+import {useDispatch, useSelector} from 'react-redux';
 import {Button, Container, Icon, Label, List, Modal, Segment} from 'semantic-ui-react';
 import UserSearchForm from './UserSearchForm';
 import UserSearchResults from './UserSearchResults';
 import UserAvatar from './UserAvatar';
+import {updateParticipantList, removeParticipant} from '../actions';
+import {getMeetingParticipants} from '../selectors';
 import client from '../client';
 
 import styles from './UserSearch.module.scss';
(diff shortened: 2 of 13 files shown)


mic4ael:
Oh, actually I forgot one thing here. I should probably store those participants in the redux store so that we can access them in the last step.

ThiefMaster:
Yeah, but that doesn't need to be part of this PR since I guess this one is mostly about actually being able to search for them

mic4ael:
But I think it is a pretty small change which will not take a lot of time to implement :P

newdle/client/src/components/UserSearch.js
```diff
@@ -46,9 +50,9 @@ export default function UserSearch() {
   );
 
   const handleModalConfirm = useCallback(() => {
-    setParticipants([...participants, ...stagedParticipants]);
+    dispatch(updateParticipantList([...participants, ...stagedParticipants]));
```
ThiefMaster: I think this should just pass the staged ones and leave it up to the reducer to merge the old/new list

mic4ael: But then I would also have to handle it differently when removing a participant from the list. If I do it like that then I can handle all cases with one action.

ThiefMaster: I think having separate actions to add/remove participants is cleaner...

newdle/client/src/components/UserSearch.js
```diff
@@ -21,16 +24,15 @@ async function searchUsers(data, setResults) {
...
   const [searchResults, setSearchResults] = useState(null);
-  const [participants, setParticipants] = useState([]);
   const [stagedParticipants, setStagedParticipants] = useState([]);
 
-  const removeParticipant = useCallback(
-    participant => {
-      setParticipants(participants.filter(p => p.email !== participant.email));
-    },
-    [participants]
+  const handleRemoveParticipant = useCallback(
```
mvidalgarcia: Just curious as I'm not totally familiar with redux hooks, what's the advantage of using `useCallback` here instead of dispatching the action directly?

ThiefMaster: It keeps the identity of `handleRemoveParticipant` consistent between renders so it avoids re-renders of child components it's passed to.

mvidalgarcia: In this particular case, would it avoid re-renders of the `<Icon />` component?

newdle/client/src/components/UserAvatar.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Gravatar from 'react-gravatar';
+
+const DEFAULT_AVATAR_SIZE = 40;
+const defaultArgs = [
+  'f0e9e9', // background
```
mvidalgarcia: Pity we cannot import this from our palette :disappointed:

newdle/client/src/components/UserSearch.js
mic4ael: I actually thoought about the same. And I think I can remove it.

newdle/client/src/components/UserSearch.js
```diff
@@ -0,0 +1,136 @@
...
+  const [searchResults, setSearchResults] = useState(null);
+  const [stagedParticipants, setStagedParticipants] = useState([]);
+
+  const handleRemoveParticipant = participant => dispatch(removeParticipant(participant));
+
+  const handleModalClose = useCallback(() => {
+    setUserModalOpen(false);
+    setSearchResults(null);
+    setStagedParticipants([]);
+  }, [setUserModalOpen, setSearchResults]);
```
mvidalgarcia: Isn't `setStagedParticipants` missing here?

mic4ael: Yeah, good catch; )

mic4ael: Don't know why the linter didn't complain about this :P

newdle/client/src/components/UserSearchResults.js
```diff
@@ -0,0 +1,38 @@
...
+          <List.Item key={user.email} className={styles['result-item']}>
+            <List.Content floated="right">
+              {isAdded(user) ? (
+                <Icon name="checkmark" color="green" size="large" />
+              ) : (
+                <Icon name="plus" size="large" onClick={() => onAdd(user)} />
+              )}
+            </List.Content>
+            <List.Icon size="large" name="user" verticalAlign="middle" />
+            <List.Content>{user.name}</List.Content>
```
meluru: What do you think about displaying user's email in the results as well? We allow searching by the email address so I think it would be good to display it. (I can also imagine a situation when we search by name and we get the same person twice because s/he is has registered twice with different email addresses. Then it would be also good to display the address to know which one we are choosing).

PR #39, opened 2019-08-28
Run pycodestyle & apply black/isort to tests

ThiefMaster:


setup.cfg
```diff
@@ -0,0 +1,3 @@
+[pycodestyle]
+exclude=newdle/client,__pycache__
+max-line-length=89
```
pferreir: Aren't we using 120?

ThiefMaster: no currently we're on the 89 default from black

PR #41, opened 2019-08-28
Add newdle codes

### LINKED ISSUE #32: Add logic to generate newdle URL
This means generating some random ID at newdle creation time. For now we can just use a UUID, but maybe for the production service we should consider the [same approach we used for ursh](https://github.com/indico/ursh/blob/master/ursh/models.py#L50).

pferreir:
Closes #32

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -2,7 +2,7 @@ from faker import Faker
 from flask import Blueprint, g, jsonify, request
 from itsdangerous import BadData, SignatureExpired
 from marshmallow import fields
-from werkzeug.exceptions import UnprocessableEntity
+from werkzeug.exceptions import Forbidden, NotFound, UnprocessableEntity
 
 from .core.auth import user_info_from_app_token
 from .core.db import db
@@ -84,7 +84,7 @@ def users(q):
 
 @api.route('/newdle/', methods=('POST',))
 @use_kwargs(NewNewdleSchema(), locations=('json',))
-def newdle(title, duration, timezone, time_slots, participants):
+def create_newdle(title, duration, timezone, time_slots, participants):
     newdle = Newdle(
         title=title,
         creator_uid=g.user['uid'],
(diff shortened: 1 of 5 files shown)


newdle/api.py
```diff
@@ -96,3 +96,13 @@ def newdle(title, duration, timezone, time_slots, participants):
     db.session.add(newdle)
     db.session.commit()
     return NewdleSchema().jsonify(newdle)
+
+
+@api.route('/newdle/<code>')
+def get_newdle(code):
+    newdle = Newdle.query.filter(Newdle.code == code).first()
```
ThiefMaster: `.filter_by(code=code)`

ThiefMaster: `.one_or_404()`?

newdle/models.py
```diff
@@ -6,6 +9,19 @@
 from newdle.core.util import UTCDateTime, format_dt, parse_dt
 
 
+CODE_ALPHABET = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-'
+
+
+def generate_random_code():
+    """Generate a random newdle code, based on a restricted alphabet."""
+    code_length = current_app.config.get('NEWDLE_CODE_LENGTH', 8)
```
ThiefMaster: since we load the example config before the normal config, using `[]` here is safe (and avoids a default value that is never used)

newdle/models.py
```diff
@@ -16,6 +32,9 @@ class Newdle(db.Model):
     _timezone = db.Column('timezone', db.String, nullable=False)
     _time_slots = db.Column('time_slots', JSONB, nullable=False)
     final_dt = db.Column(UTCDateTime, nullable=True)
+    code = db.Column(
+        db.String, nullable=False, index=True, default=generate_random_code
```
ThiefMaster: `unique=True`

pferreir: Good catch, thanks!

newdle/api.py
pferreir: I didn't realize `.one_or_404()` comes from `flask-sqlalchemy`. I thought it was an Indico thing. Thanks!

ThiefMaster: There's also `first_or_404`, which is fine here since the column is unique

PR #47, opened 2019-08-29
Link calendar with the timeline

### LINKED ISSUE #35: Step 2: link calendar with timeline view
We should be able to navigate through the days by clicking on it.

mic4ael:
closes #35

### DIFF
--- a/newdle/client/src/TimelineExample.js
+++ b/newdle/client/src/TimelineExample.js
@@ -1,10 +1,21 @@
 import React from 'react';
-import Timeline from './components/Timeline';
 import moment from 'moment';
+import {useSelector} from 'react-redux';
+import {getCalendarActiveDate, getCalendarDates} from './selectors';
+import {serializeDate} from './util/date';
+import Timeline from './components/Timeline';
 
 export default function TimelineExample({participants}) {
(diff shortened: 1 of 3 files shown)


newdle/client/src/TimelineExample.js
```diff
@@ -1,10 +1,21 @@
...
 
 export default function TimelineExample({participants}) {
   const availability = generateParticipantAvailability(participants);
-  return <Timeline date="2019-07-29" availability={availability} />;
+  const dates = useSelector(getCalendarDates);
+  const activeDate = useSelector(getCalendarActiveDate);
+
+  return (
+    <Timeline
+      date={activeDate ? serializeDate(activeDate) : dates[0]}
```
ThiefMaster: What if `dates` is empty? Right now we have dummy dates, but once timeslots are in Redux I think we'll only have the dates where timeslots exist, so we should probably default to today or tomorrow there.

ThiefMaster: Also, why is this still called `TimelineExample`? :D

mic4ael: Ohh, right. I was thinking that we would always have some dates there. Anyway I will default to today and ofc will change the name :P

PR #48, opened 2019-08-29
Final step

### LINKED ISSUE #26: Step 3 - Finalize newdle
At this point we should ask the user for the event's title. We should also tell the user that the participants will receive an e-mail once they confirm the creation.

`POST /api/newdle` should be called with the filled in info (also from the previous steps) and then the page should change slightly to reflect that fact that the newdle has been created.

image
image

panagiotappl:
closes #26 

TODO
- [x] - Add real data for `timeslot` coming from redux store in POST request

### DIFF
--- a/newdle/client/src/client.js
+++ b/newdle/client/src/client.js
@@ -32,7 +32,11 @@ class Client {
     return this._request(flask`api.users`({q}));
   }
 
-  createNewdle(params) {
+  createNewdle(title, duration, timezone, timeslots) {
+    const params = {
+      method: 'POST',
+      body: JSON.stringify({title, duration, timezone, timeslots}),
+    };
     return this._request(flask`api.create_newdle`(), params);
   }
 
--- a/newdle/client/src/client.js
+++ b/newdle/client/src/client.js
@@ -32,6 +32,10 @@ class Client {
     return this._request(flask`api.users`({q}));
   }
 
+  createNewdle(params) {
+    return this._request(flask`api.create_newdle`(), params);
+  }
+
   async _request(url, options = {}, withStatus = false, isRetry = false) {
     const headers = {Accept: 'application/json'};
     const requestOptions = {headers, ...options};
--- a/newdle/client/src/client.js
+++ b/newdle/client/src/client.js
@@ -32,10 +32,10 @@ class Client {
     return this._request(flask`api.users`({q}));
   }
 
-  createNewdle(title, duration, timezone, timeslots) {
+  createNewdle(title, duration, timezone, timeslots, participants) {
     const params = {
       method: 'POST',
-      body: JSON.stringify({title, duration, timezone, timeslots}),
+      body: JSON.stringify({title, duration, timezone, timeslots, participants}),
     };
     return this._request(flask`api.create_newdle`(), params);
   }
--- a/newdle/client/src/components/CreateNewdle.js
+++ b/newdle/client/src/components/CreateNewdle.js
@@ -1,18 +1,22 @@
-import React, {useEffect} from 'react';
+import React, {useState, useEffect} from 'react';
 import {Button, Container, Grid, Header, Icon, Input} from 'semantic-ui-react';
 import {useDispatch, useSelector} from 'react-redux';
 import {
   getStep,
+  getTitle,
+  getDuration,
   areParticipantsDefined,
   getMeetingParticipants,
   shouldConfirmAbortCreation,
+  getFullTimeslots,
 } from '../selectors';
-import {abortCreation, setStep} from '../actions';
+import {abortCreation, setStep, setTitle} from '../actions';
 import Calendar from './Calendar';
 import Availability from './Availability';
 import UserSearch from './UserSearch';
 import UnloadPrompt from './UnloadPrompt';
 import styles from './CreateNewdle.module.scss';
+import client from '../client';
 
 export default function CreateNewdle() {
   const step = useSelector(getStep);
@@ -85,30 +89,77 @@ function TimeSlotsPage() {
 }
 
 function FinalizePage() {
+  const [success, setSuccess] = useState(false);
+  const [newdle, setNewdle] = useState(null);
+  const title = useSelector(getTitle);
+  const duration = useSelector(getDuration);
+  const timeslots = useSelector(getFullTimeslots);
+  const timezone = 'UTC';
+  const dispatch = useDispatch();
+
+  async function createNewdle() {
+    const params = {
+      method: 'POST',
+      body: JSON.stringify({title, duration, timezone, time_slots: timeslots}),
+    };
+    const results = await client.createNewdle(params);
+    setNewdle(results);
+    setSuccess(true);
+  }
+
   return (
     <Container text>
-      <Header as="h2" className={styles['input-title']}>
-        Title of your meeting
-      </Header>
-      <Input className={styles['title-input']} focus />
-      <div className={styles['attention-message']}>
-        <Header as="h3" className={styles['header']}>
-          Attention
-        </Header>
-        <p>
-          Your participants will receive an e-mail asking them to register to their preference. Once
-          the Newdle is created, you will be shown a link you can share with anyone else you wish to
-          invite.
-        </p>
-      </div>
-      <div className={styles['create-button']}>
-        <Button color="violet">
-          Create your Newdle!{' '}
-          <span role="img" aria-label="Steaming Bowl Emoji">
-            &#x1f35c;
-          </span>
-        </Button>
-      </div>
+      {!success && !newdle ? (
+        <>
+          <Header as="h2" className={styles['input-title']}>
+            Title of your event
+          </Header>
+          <Input
+            className={styles['title-input']}
+            focus
+            onChange={(_, data) => dispatch(setTitle(data.value))}
+          />
+          <div className={styles['attention-message']}>
+            <Header as="h3" className={styles['header']}>
+              Attention
+            </Header>
+            <p>
+              Your participants will receive an e-mail asking them to register to their preference.
+              Once the Newdle is created, you will be shown a link you can share with anyone else
+              you wish to invite.
+            </p>
+          </div>
+          <div className={styles['create-button']}>
+            <Button
+              color="violet"
+              type="submit"
+              disabled={title.length < 3}
+              onClick={async () => await createNewdle()}
+            >
+              Create your Newdle! <span role="img">🍜</span>
+            </Button>
+          </div>
+        </>
+      ) : (
+        <>
+          <Header as="h1" className={styles['newdle-title']}>
+            {newdle.title}
+          </Header>
+          <div className={styles['success-message']}>
+            <Header as="h3" className={styles['header']}>
+              Done!
+            </Header>
+            <p>
+              Your Newdle was created and invite e-mails have been sent. You can send the following
+              link to everyone you'd like to invite:
+            </p>
+            <div className={styles['newdle-link']}>{newdle.url}</div>
+          </div>
+          <div className={styles['summary-button']}>
+            <Button color="teal">Go to Newdle summary!</Button>
+          </div>
+        </>
+      )}
     </Container>
   );
 }
--- a/newdle/client/src/components/CreateNewdle.js
+++ b/newdle/client/src/components/CreateNewdle.js
@@ -111,12 +111,11 @@ function FinalizePage() {
     <Container text>
       {!success && !newdle ? (
         <>
-          <Header as="h2" className={styles['input-title']}>
-            Title of your event
-          </Header>
           <Input
+            autoFocus
+            transparent
             className={styles['title-input']}
-            focus
+            placeholder="Please enter a title for your event..."
             onChange={(_, data) => dispatch(setTitle(data.value))}
           />
           <div className={styles['attention-message']}>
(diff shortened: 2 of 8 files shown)


newdle/client/src/client.js
```diff
@@ -32,15 +33,33 @@ class Client {
     return this._request(flask`api.users`({q}));
   }
 
-  async _request(url, withStatus = false, isRetry = false) {
-    const headers = {Accept: 'application/json'};
+  createNewdle() {
+    return this._request(flask`api.create_newdle`(), {
+      method: 'POST',
+      body: JSON.stringify({
+        title: getTitle(this.store.getState()),
```
ThiefMaster: the client should not access the store - the only reason it has access to it is for the token refresh logic. regular client function should receive everything as params instead.

newdle/client/src/client.js
```diff
@@ -32,15 +33,33 @@ class Client {
...
+          start: moment().format('YYYY-MM-DDThh:mm'),
+          end: moment()
+            .add(15, 'minutes')
+            .format('YYYY-MM-DDThh:mm'),
+        },
+      }),
+    });
+  }
+
+  async _request(url, options = {}, withStatus = false, isRetry = false) {
```
ThiefMaster: there's a recursive call inside this function where you don't pass `options` yet

newdle/client/src/client.js
```diff
@@ -32,15 +33,33 @@ class Client {
...
+    const headers = {Accept: 'application/json', ...options};
```
ThiefMaster: why are you unpacking options into headers?

newdle/client/src/components/CreateNewdle.js
```diff
@@ -52,29 +43,104 @@ function TimeSlotsPage() {
...
+  );
+}
+
+async function createNewdle() {
+  const results = await client.createNewdle();
+  return results;
+}
+
+function FinalizePage() {
+  const [successMessage, setSuccessMessage] = useState(false);
```
ThiefMaster: I'd go for `success` / `setSuccess` - in `somethingMessage` you usually expect an actual message, not a bool.

newdle/client/src/components/CreateNewdle.js
```diff
@@ -52,29 +43,104 @@ function TimeSlotsPage() {
...
+  const [newdle, setNewdle] = useState(null);
+  const dispatch = useDispatch();
+  return (
+    <Container text>
+      {!successMessage && !newdle ? (
+        <>
+          <Header as="h2" className={styles['input-title']}>
+            Title of your meeting
```
ThiefMaster: s/meeting/event/?

newdle/client/src/components/CreateNewdle.module.scss
```diff
@@ -1,5 +1,65 @@
...
+.newdle-title {
+  display: flex;
+  justify-content: center;
+  color: $purple !important;
+}
+
+.newdle-link {
+  display: flex;
+  justify-content: center;
+  padding: 0.67857143em 1em;
```
ThiefMaster: where is this weird value coming from?

newdle/client/src/components/Home.js
```diff
@@ -1,15 +1,21 @@
...
 export default function Home() {
+  const dispatch = useDispatch();
   const isUserLoggedIn = useSelector(isLoggedIn);
 
   return (
     <>
       <h2>Home</h2>
-      {isUserLoggedIn && <Link to="/new">Create newdle</Link>}
+      {isUserLoggedIn && (
+        <Link to="/new" onClick={() => dispatch(setStep(1))}>
```
ThiefMaster: Not sure this belongs here. What about dispatching an action that clears *all* creation-related data from the state instead (I think right now the participant list etc. isn't reset), and, since resetting the step would show the wrong page, redirect to something like `/created` after creation and show the success message etc. there? (in that case it now does make sense to put some data about to the newly created newdle in the redux store)

newdle/client/src/components/CreateNewdle.js
```diff
@@ -52,29 +43,104 @@ function TimeSlotsPage() {
...
+              you wish to invite.
+            </p>
+          </div>
+          <div className={styles['create-button']}>
+            <Button
+              color="violet"
+              type="submit"
+              onClick={async () => {
+                const results = await createNewdle();
+                setNewdle(results);
```
ThiefMaster: shouldn't this stuff be inside `createNewdle`?

newdle/client/src/components/CreateNewdle.js
```diff
@@ -52,29 +43,104 @@ function TimeSlotsPage() {
...
+            <p>
+              Your Newdle was created and invite e-mails have been sent. You can send the following
+              link to everyon you'd like to invite:
+            </p>
+            <div className={styles['newdle-link']}>{newdle.url}</div>
+          </div>
+          <div className={styles['summary-button']}>
+            <Button color="teal">
+              Go to Newdle summary{' '}
+              <span role="img" aria-label="Steaming Bowl Emoji">
```
ThiefMaster: Do we really need that stuff instead of just putting the literal emoji there? It doesn't even reduce a11y since someone not seeing the emoji would not lose any relevant information.

newdle/client/src/components/CreateNewdle.module.scss
```diff
@@ -1,5 +1,65 @@
...
+  border: 1px solid rgba(34, 36, 38, 0.15);
+  border-radius: 0.28571429rem;
```
ThiefMaster: another weird value

newdle/client/src/components/Home.js
panagiotappl: If I set the step after successful creation wouldn't the app transition from step 3 (where are are looking at the success page with the event url) to step 1?

ThiefMaster: > since resetting the step would show the wrong page, redirect to something like /created after creation and show the success message etc. there

newdle/client/src/components/CreateNewdle.module.scss
panagiotappl: I wanted to use exactly the same padding semantic uses for inputs but I can change it to a less weird value.

newdle/client/src/client.js
panagiotappl: Somehow rebased the wrong code. It shouldn't be here. I'll fix it.

newdle/client/src/client.js
panagiotappl: I didn't notice, thanks!

newdle/client/src/components/Home.js
panagiotappl: I see you already replied in that message

newdle/client/src/components/CreateNewdle.module.scss
ThiefMaster: no in that case it's fine (but maybe add a comment mentioning that)

newdle/client/src/components/CreateNewdle.js
```diff
@@ -52,29 +43,104 @@ function TimeSlotsPage() {
...
+          </Button>
+        </div>
+      </Grid.Row>
+    </Grid>
+  );
+}
+
+async function createNewdle() {
+  const results = await client.createNewdle();
+  return results;
```
mic4ael: `return await client.createNewdle()`?

newdle/client/src/client.js
```diff
@@ -32,15 +33,33 @@ class Client {
...
+    return this._request(flask`api.create_newdle`(), {
+      method: 'POST',
+      body: JSON.stringify({
+        title: getTitle(this.store.getState()),
+        duration: getDuration(this.store.getState()),
+        timezone: 'UTC',
+        time_slots: {
+          start: moment().format('YYYY-MM-DDThh:mm'),
+          end: moment()
+            .add(15, 'minutes')
```
pferreir: Shouldn't `time_slots` be generated from the state?

newdle/client/src/components/CreateNewdle.js
```diff
@@ -52,29 +43,104 @@ function TimeSlotsPage() {
...
+          <Header as="h1" className={styles['newdle-title']}>
+            {newdle.title}
+          </Header>
+          <div className={styles['success-message']}>
+            <Header as="h3" className={styles['header']}>
+              Done!
+            </Header>
+            <p>
+              Your Newdle was created and invite e-mails have been sent. You can send the following
+              link to everyon you'd like to invite:
```
pferreir: Typo: "everyone".

newdle/client/src/client.js
panagiotappl: There was no time slots in the state when I was working on this. It's mentioned in my todos.

newdle/client/src/components/CreateNewdle.js
```diff
@@ -93,3 +94,86 @@ function TimeSlotsPage() {
...
   );
 }
+
+function FinalizePage() {
+  const [success, setSuccess] = useState(false);
+  const [newdle, setNewdle] = useState(null);
+  const title = useSelector(getTitle);
+  const duration = useSelector(getDuration);
+  const timeslots = useSelector(getFullTimeslots);
+  const timezone = 'UTC';
```
ThiefMaster: while we'll eventually add a dropdown to select one, what about using `Europe/Zurich` for now?

newdle/client/src/components/CreateNewdle.js
```diff
@@ -93,3 +94,86 @@ function TimeSlotsPage() {
...
+              Once the Newdle is created, you will be shown a link you can share with anyone else
+              you wish to invite.
+            </p>
+          </div>
+          <div className={styles['create-button']}>
+            <Button
+              color="violet"
+              type="submit"
+              disabled={title.length < 3}
+              onClick={async () => await createNewdle()}
```
ThiefMaster: `onClick={createNewdle}`

newdle/client/src/selectors.js
```diff
@@ -23,3 +23,10 @@ export const shouldConfirmAbortCreation = createSelector(
   areParticipantsDefined,
   (timeslots, hasParticipants) => Object.keys(timeslots).length || hasParticipants
 );
+export const getTitle = state => state.title;
+export const getFullTimeslots = state =>
+  [].concat(
+    ...Object.keys(state.timeslots).map(date =>
+      state.timeslots[date].map(time => `${date}T${time}`)
+    )
+  );
```
ThiefMaster: this can be simplified:

```js
export const getFullTimeslots = state =>
  [].concat(
    ...Object.entries(state.timeslots).map(([date, slots]) =>
      slots.map(slot => `${date}T${slot}`)
    )
  );
```

newdle/migrations/versions/20190913_1031_000000000000_create_initial_tables.py
```diff
@@ -1,7 +1,7 @@
 """Create initial tables
 
 Revision ID: 000000000000
-Create Date: 2019-09-12 16:30:45.370144
+Create Date: 2019-09-13 10:31:54.565733
```
ThiefMaster: nothing changed in the structure so it shouldn't be necessary to regenerate the revision

newdle/client/src/components/CreateNewdle.js
```diff
@@ -93,3 +94,77 @@ function TimeSlotsPage() {
...
+          <Header as="h1" className={styles['newdle-title']}>
+            {newdle.title}
+          </Header>
+          <div className={styles['success-message']}>
+            <Header as="h3" className={styles['header']}>
+              Done!
+            </Header>
+            <p>
+              Your Newdle was created and invite e-mails have been sent. You can send the following
+              link to everyone you'd like to invite:
```
meluru: `you would like to invite`?

newdle/client/src/components/CreateNewdle.js
```diff
@@ -93,3 +94,77 @@ function TimeSlotsPage() {
...
+            </Header>
+            <p>
+              Your participants will receive an e-mail asking them to register to their preference.
+              Once the Newdle is created, you will be shown a link you can share with anyone else
+              you wish to invite.
+            </p>
+          </div>
+          <div className={styles['create-button']}>
+            <Button color="violet" type="submit" disabled={title.length < 3} onClick={createNewdle}>
+              Create your Newdle! <span role="img">🍜</span>
```
meluru: Missing `aria-label` (check the console)

newdle/client/src/components/CreateNewdle.js
```diff
@@ -66,8 +60,9 @@ function ParticipantsPage() {
 function TimeSlotsPage() {
   const dispatch = useDispatch();
   const participants = useSelector(getMeetingParticipants);
+  const timeslots = useSelector(getFullTimeslots);
   return (
-    <div className={styles['time-slots-grid']}>
+    <div>
```
mic4ael: Do we need that div in the end?

newdle/client/src/components/CreateNewdle.module.scss
```diff
@@ -1,5 +1,88 @@
...
   display: flex;
   justify-content: flex-end;
   width: 100%;
 }
+
+.title-input {
+  width: 100%;
+  border-bottom: 1px solid $purple !important;
+  padding-bottom: 10px !important;
+  input {
```
mic4ael: Maybe separate the `input` rule with the attribute in the previous line with an empty line?

newdle/client/src/components/CreateNewdle.js
panagiotappl: Yes, I mentioned it once in the channel. Adding the `div` adds some extra padding to the entire component that otherwise is not there because of this css:
```
main {
  > * {
    padding: 3em;
  }
}
```
inside `App.module.scss`

mic4ael: Ok, good to know ;)

PR #49, opened 2019-09-16
Add backend for responding to a newdle

ThiefMaster:


newdle/models.py
```diff
@@ -59,14 +77,35 @@ class Participant(db.Model):
...
     def __repr__(self):
         return '<Participant {}: {}{}>'.format(
             self.id, self.name, ' ({})'.format(self.email) if self.email else ''
         )
+
+
+class Availability(AutoNameEnum):
+    unavailable = auto()
+    available = auto()
+    ifneedbe = auto()
```
mic4ael: `ifneeded` maybe?

newdle/schemas.py
```diff
@@ -24,12 +26,24 @@ def sort_users(self, data, many, **kwargs):
...
     auth_uid = fields.String()
 
 
+class ParticipantSchema(NewParticipantSchema):
+    answers = fields.Mapping(
+        fields.DateTime(format=DATETIME_FORMAT), EnumField(Availability)
+    )
+
+
+class UpdateParticipantSchema(mm.Schema):
```
mic4ael: Why do we need two identical schemas?

newdle/models.py
ThiefMaster: I stuck with the term Doodle uses ;)

PR #55, opened 2019-09-19
Add Timezone selector

### LINKED ISSUE #45: Add support for timezones
Dates/times should be already timezone-aware in the backend, so not much should need to be done.

 * We should add a timezone selector in step 2, which could default to the user's client timezone;
 * When creating the newdle, we should send the selected timezone along with it.

pferreir:
image

I have also improved a bit the `DurationPicker`, to make it more user-friendly.

Closes #45


newdle/client/src/components/TimezonePicker.js
```diff
@@ -0,0 +1,31 @@
...
+import {getTimezone} from '../selectors';
+import {setTimezone} from '../actions';
+
+import {commonTimezones} from '../util/timezones';
+import styles from './TimezonePicker.module.scss';
+
+export default function TimezonePicker() {
+  const dispatch = useDispatch();
+  const timezone = useSelector(getTimezone);
+  const options = commonTimezones.map(({name, caption}) => {
```
ThiefMaster: this one could be wrapped in `useMemo()`

newdle/client/src/util/timezones.js
```diff
@@ -0,0 +1,43 @@
...
+      accum[tz] = true;
+    })
+    return accum;
+  }, {});
+
+  // add some common timezones which are not official in any country
+  [
+    // UTC/GMT
+    'UTC',
+    'GMT',
```
ThiefMaster: is GMT still relevant?

newdle/client/src/components/TimezonePicker.js
```diff
@@ -0,0 +1,31 @@
...
+    return {key: name, value: name, text: name, description: caption, active: name === timezone};
+  });
+  return (
+    <div>
+      <span>Timezone:</span>
+      <Dropdown
```
ThiefMaster: add `selectOnBlur={false}` to avoid selecting a value when clicking outside

newdle/client/src/util/timezones.js
```diff
@@ -0,0 +1,43 @@
...
+    zones.forEach(tz => {
+      accum[tz] = true;
+    })
+    return accum;
+  }, {});
+
+  // add some common timezones which are not official in any country
+  [
+    // UTC/GMT
+    'UTC',
```
ThiefMaster: I don't see UTC in the dropdown (nor any of the other timezones from this list)

newdle/client/src/util/timezones.js
```diff
@@ -0,0 +1,43 @@
...
+  ].forEach(k => {
+    commonTzSet[k] = true;
+  });
+
+  const timezones = Object.keys(commonTzSet).map(k => {
+    const z = moment.tz.zone(k);
+    return {
+      name: z.name,
+      offset: z.utcOffset(+ new Date()),
+      caption: moment.tz(k).format('z (Z)')
```
ThiefMaster: the zone caption looks pretty strange for some timezones:

image

newdle/client/src/components/TimezonePicker.js
ThiefMaster: the dropdown has really awful performance here (and slows down the page render by ~150ms which is very noticable).

Semantic-Org/Semantic-UI-React#1674 might be related

newdle/client/src/util/timezones.js
pferreir: What do you mean?

newdle/client/src/util/timezones.js
pferreir: Weird. I can see them.

pferreir: Maybe it's because I force-pushed them a few minutes after the PR was opened?

newdle/client/src/util/timezones.js
pferreir: I copied that [from here](https://github.com/stub42/pytz/blob/master/gen_tzinfo.py#L111).
Not sure whether everyone knows that UTC is.
https://www.timeanddate.com/time/gmt-utc-time.html

Maybe we should just remove both of them? Why would you want to hold a meeting in UTC?

newdle/client/src/util/timezones.js
ThiefMaster: `+14 (+14:00)` looks pretty strange to me

pferreir: Ah yes. Maybe I can remove the first part when it's just numbers/signs.

newdle/client/src/components/TimezonePicker.js
pferreir: Doesn't feel that heavy here. But there isn't much we can do.

newdle/client/src/util/timezones.js
ThiefMaster: OK maybe the babel cache is to blame... maybe it didn't rebuild the file after I checked out your updated branch... after having been on a different branch for a while and switching back to yours it worked

newdle/client/src/components/DurationPicker.js
```diff
@@ -17,16 +32,14 @@ export default function DurationPicker({minDuration, maxDuration, interval}) {
   const dispatch = useDispatch();
 
   return (
-    <div className={styles['duration-picker']}>
+    <div>
       <span>Meeting time: </span>
```
ThiefMaster: i think it might look better without the `:`

pferreir: Indeed.

PR #56, opened 2019-09-19
Participant table

### LINKED ISSUE #36: "Summary page" - participant table widget
If someone wants to get started on this before #27 is done, they can just use mock data for now.

image

mic4ael:
closes #36

### DIFF
--- a/newdle/client/src/actions.js
+++ b/newdle/client/src/actions.js
@@ -16,7 +16,8 @@ export const ADD_TIMESLOT = 'Add new timeslot';
 export const REMOVE_TIMESLOT = 'Remove a timeslot';
 export const SET_TITLE = 'Set title for Newdle';
 export const SET_TIMEZONE = 'Set the meeting timezone';
-export const FETCH_NEWDLE = 'Fetch newdle data';
+export const NEWDLE_RECEIVED = 'Received newdle data';
+export const CLEAR_NEWDLE = 'Clear newdle data';
 
 export function userLogin(token) {
   return async dispatch => {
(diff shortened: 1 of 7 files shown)


newdle/client/src/reducers.js
```diff
@@ -125,4 +126,12 @@ export default combineReducers({
         return state;
     }
   },
+  newdle: (state = {}, action) => {
```
ThiefMaster: wouldn't `null` be a better default here?

newdle/client/src/actions.js
```diff
@@ -76,3 +77,10 @@ export function removeTimeslot(date, time) {
 export function setTitle(title) {
   return {type: SET_TITLE, title};
 }
+
+export function fetchNewdle(code, fullDetails) {
```
panagiotappl: What if you don't want to pass `fullDetails` as a parameter here? I noticed that in `client.js` it's set to `false` by default.

ThiefMaster: it would be undefined, ie still falsy. but it's better to add `= false` as a default argument here as well...

newdle/client/src/components/ParticipantTable.js
```diff
@@ -0,0 +1,120 @@
...
+const MAX_PARTICIPANTS_SHOWN = 4;
+
+function formatMeetingTime(startTime, duration) {
+  const endTime = startTime.add(duration, 'm').format('HH:mm');
+  return `${serializeDate(startTime, 'HH:mm')} - ${endTime}`;
+}
+
+function renderParticipantNames(availableParticipants) {
+  if (availableParticipants.length === 0) {
+    return <Message warning>Nobody is available at this time</Message>;
```
meluru: Maybe this could be a regular div? The Message box doesn't really work here visually

newdle/client/src/actions.js
mic4ael: The default arg to `client.getNewdle` is `false` anyway so it wouldn't be a problem. But anyway I will add the default value here as well.

newdle/client/src/selectors.js
```diff
@@ -31,3 +31,22 @@ export const getFullTimeslots = state =>
...
+  getNewdleParticipants,
+  participants => participants.length
+);
+export const getParticipantAvailability = createSelector(
+  getNewdleTimeslots,
+  getNewdleParticipants,
+  (timeslots, participants) => {
+    return timeslots.map(timeslot => {
+      const available = participants.filter(
+        participant => participant.answers[timeslot] === 'available'
```
ThiefMaster: shouldn't `ifneedbe` be considered available as well? (at least I wouldn't consider those unavailable.. maybe yellow in the circular progress thing if it supports more than 2 states?)

PR #59, opened 2019-09-20
Make login process more user-friendly

ThiefMaster:
- Make "create newdle" button always visible but automatically start the login process when clicking while logged out
- Show a page dimmer + loader while the login window is open


newdle/client/src/auth.js
```diff
@@ -14,6 +17,8 @@ export function useAuthentication() {
     if (popup.current) {
       popup.current.close();
     }
+    popupId.current = Date.now();
```
pferreir: What about users opening windows super-fast? 😉

newdle/client/src/auth.js
```diff
@@ -63,6 +69,25 @@ export function useAuthentication() {
     };
   }, [dispatch]);
 
+  useEffect(() => {
+    if (isUserLoggedIn) {
+      return;
+    }
+    const interval = window.setInterval(() => {
```
pferreir: Maybe we don't need to do this? Is it really a big deal if we leave the `windowId` in the state when the user closes the window? What's the worst thing that can happen? The user opening another window and the value being rewritten? Or maybe there's something I'm not seeing...

ThiefMaster: `onunload` for windows is apparently not very reliable (at least based on google/SO results, didn't test it myself), but I need to know when the window closed (e.g. without logging in) so I can dispatch the action to hide the "logging in" dimmer/loader

newdle/client/src/auth.js
```diff
@@ -14,6 +17,8 @@ export function useAuthentication() {
     if (popup.current) {
       popup.current.close();
```
mic4ael: Shouldn't you also dispatch `loginWindowClosed` here if `popupId` has been set? Or is it handled by the `setInterval` callback?

ThiefMaster: We're immediately opening a new window here, so there's no need to dispatch that here. Also, it should be very unlikely that `login()` gets called while another window is open (since we now disable the page with the page dimmer while it's open)

ThiefMaster: FWIW, `login()` can ONLY be called explicitly through user interaction, since any other way of calling it would likely trigger the browser's popup blocker.

newdle/client/src/auth.js
ThiefMaster: btw i just gave this a try, and it's as awful as people on SO said: `onunload` triggers once on load (for `about:blank` unloading), and even if i register onunload again while handling that event, it won't run when the window is actually closed :man_facepalming:

PR #61, opened 2019-09-20
Answer page

### LINKED ISSUE #37: "Answer page" - basic layout

meluru:
Closes #37


newdle/client/src/components/Answer.js
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div>
+      <Grid container>
+        <Grid.Row columns={2}>
+          <Grid.Column width={5}>
+            <Calendar />
+            <Header as="h3" className={styles['options-msg']}>
+              4 out of 7 options chosen
+            </Header>
+            <Checkbox toggle label="Accept all options when I'm available" />
```
ThiefMaster: "when" or "where"?

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  const position = (startMins / spanMins) * 100;
+  return position < 100 ? position : 100 - OVERFLOW_HEIGHT;
+}
+
+function groupOverlaps(timeSlots) {
+  const sortedTimeSlots = _.orderBy(timeSlots, ['pos'], ['asc']);
+  // marks items that overlap with its successors
+  let cluster_id = 0;
```
ThiefMaster: `clusterId`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+  const clusteredCandidates = sortedTimeSlots.map((candidate, index, sortedTimeSlots) => {
+    if (
+      sortedTimeSlots[index - 1] &&
+      moment(candidate.startTime, 'HH:mm').isSameOrAfter(
+        moment(sortedTimeSlots[index - 1].endTime, 'HH:mm')
+      )
+    ) {
+      cluster_id += 1;
+    }
+    return {...candidate, cluster_id: cluster_id};
```
ThiefMaster: `{...candidate, cluster_id}`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  return _.chain(clusteredCandidates)
```
ThiefMaster: this whole expression looks like it can be replaced by `Object.values(_.groupBy(clusteredCandidates, c => c.cluster_id))`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+  const start = toMoment(slot, 'YYYY-MM-DDTHH:mm');
+  const end = moment(start).add(duration, 'm');
+
+  return {
+    slot,
+    startTime: serializeDate(start, 'HH:mm'),
+    endTime: serializeDate(end, 'HH:mm'),
+    height: calculateHeight(start, end, minHour, maxHour),
+    pos: calculatePosition(start, minHour, maxHour),
+    key: `${slot}`,
```
ThiefMaster: `key: slot,`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+  };
+}
+
+function getDate(timeslot) {
+  return serializeDate(toMoment(timeslot, 'YYYY-MM-DDTHH:mm'), 'YYYY-MM-DD');
+}
+
+function calculateSlotsPositions(slots, duration, minHour, maxHour) {
+  const slotsByDate = _.chain(slots)
+    .map(slot => getSlotProps(slot, duration, minHour, maxHour))
```
ThiefMaster: Do we need lodash for all this? Looks like the only lodashy thing we really need is `groupBy`; the rest is already available in JS itself..

`_.groupBy(slots.map(...), ...)`

newdle/client/src/styles/palette.scss
```diff
@@ -7,3 +7,5 @@ $busy: #ffb3b8;
 $green: #62f46f;
 $light-teal: #00caa8;
 $blue: #2185d0;
+$candidate: rgba(130, 130, 130, 0.3);
```
ThiefMaster: pretty sure scss lets you define rgba colors in hex notation, e.g. `rgba(#rrggbb, 0.3)` or something like that

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+    .groupBy(slot => getDate(slot.slot))
+    .value();
+
+  return Object.entries(slotsByDate).map(([date, slots]) => {
+    return {date: date, slotGroups: groupOverlaps(slots)};
```
mic4ael: `return {date, slotGroups...`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+    <Grid>
+      <Grid.Row>
+        <Grid.Column width={1}>
+          <Hours minHour={minHour} maxHour={maxHour} />
+        </Grid.Column>
+        <Grid.Column width={5}>
+          <AnswerCalendarDay timeSlots={slotsByDay[0]} />
+        </Grid.Column>
+        <Grid.Column width={5}>
+          <AnswerCalendarDay timeSlots={slotsByDay[1]} />
```
mic4ael: This will fail if user defined only slots for one day. What about sth like:
```javascript

slotsByDay.slice(0, 3).map(daySlots => ...

```

newdle/client/src/components/AnswerCalendarDay.js
```diff
@@ -0,0 +1,94 @@
+import {Checkbox, Header} from 'semantic-ui-react';
+import moment from 'moment';
+import PropTypes from 'prop-types';
+import React from 'react';
+import styles from './Answer.module.scss';
+
+function CandidateOption({startTime, endTime}) {
```
mic4ael: Maybe clicking the whole `CandidateOption` should check/uncheck the `Checkbox`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+  let endMins = end.hours() * 60 + end.minutes();
+
+  if (startMins < minHour * 60) {
+    startMins = minHour * 60;
+  }
+  if (endMins > maxHour * 60) {
+    endMins = maxHour * 60;
+  }
+  const height = ((endMins - startMins) / ((maxHour - minHour) * 60)) * 100;
+  return height > 0 ? height : OVERFLOW_HEIGHT;
```
mic4ael: `return height || OVERFLOW_HEIGHT;`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+  return _.chain(clusteredCandidates)
+    .groupBy(cand => {
+      return cand.cluster_id;
+    })
+    .values()
+    .value();
+}
+
+function getSlotProps(slot, duration, minHour, maxHour) {
+  const start = toMoment(slot, 'YYYY-MM-DDTHH:mm');
```
mic4ael: I don't know if it is worth it but `moment` has a constant for it `moment.HTML5_FMT.DATETIME_LOCAL`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,159 @@
...
+    startTime: serializeDate(start, 'HH:mm'),
+    endTime: serializeDate(end, 'HH:mm'),
+    height: calculateHeight(start, end, minHour, maxHour),
+    pos: calculatePosition(start, minHour, maxHour),
+    key: `${slot}`,
+  };
+}
+
+function getDate(timeslot) {
+  return serializeDate(toMoment(timeslot, 'YYYY-MM-DDTHH:mm'), 'YYYY-MM-DD');
```
mic4ael: you don't have to pass the format here since `YYYY-MM-DD` is the default value

newdle/client/src/components/AnswerCalendar.js
meluru: Yeah, this is only temporary :) but thx

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,175 @@
...
+  }
+  const height = ((endMins - startMins) / ((maxHour - minHour) * 60)) * 100;
+  return height > MIN_HEIGHT ? height : MIN_HEIGHT;
+}
+
+function calculatePosition(start, minHour, maxHour) {
+  const spanMins = (maxHour - minHour) * 60;
+  let startMins = start.hours() * 60 + start.minutes() - minHour * 60;
+
+  if (startMins < 0) {
```
ThiefMaster: This doesn't look ends-at-midnight-safe

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,175 @@
...
+  // mark items that overlap with their successors
+  let clusterId = 0;
+  const clusteredOptions = sortedOptions.map((option, index, sortedOptions) => {
+    if (
+      sortedOptions[index - 1] &&
+      toMoment(option.startTime, 'HH:mm').isSameOrAfter(
+        toMoment(sortedOptions[index - 1].endTime, 'HH:mm')
+      )
+    ) {
+      clusterId += 1;
```
ThiefMaster: `clusterId++;`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,175 @@
...
+    }
+    return {...option, clusterId};
+  });
+  // group overlapping options
+  return Object.values(_.groupBy(clusteredOptions, c => c.clusterId));
```
ThiefMaster: I think you can pass `'clusterId'` as a string to groupBy instead of a function

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,175 @@
...
+}
+
+function getAnswerProps(slot, answer) {
+  if (answer === 'available') {
+    return {
+      icon: 'check square outline',
+      action: addAnswer(slot, 'ifneedbe'),
```
ThiefMaster: I had to look at this twice to realize that it's not a bug... I'd rather make `action` an arrow function returning this and then `dispatch(option.action())` instead of `dispatch(option.action)`

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,175 @@
...
+};
+
+export default function AnswerCalendar({minHour, maxHour}) {
+  const answers = useSelector(getAnswers);
+  const timeSlots = useSelector(getNewdleTimeslots);
+  const duration = useSelector(getNewdleDuration);
+  const optionsByDay = calculateOptionsPositions(timeSlots, duration, minHour, maxHour, answers);
+
+  if (optionsByDay.length === 0) {
+    return <div>No data</div>;
```
ThiefMaster: do we need this div or can we just return a string? (but I guess it'll change eventually anyway since it might not look great to have an unformatted 'no data' string)

newdle/client/src/components/AnswerCalendar.js
```diff
@@ -0,0 +1,175 @@
...
+            <AnswerCalendarDay options={options} key={options.date} />
+          </Grid.Column>
+        ))}
+      </Grid.Row>
+    </Grid>
+  );
+}
+
+AnswerCalendar.propTypes = {
+  minHour: PropTypes.number,
```
ThiefMaster: I think this component needs the same shift/expand logic we added on the creation page if the default times don't cover everything. i.e. if shift the min/max hours to cover everything, or if that's not enough extend it to a longer range

newdle/client/src/components/AnswerCalendarDay.js
```diff
@@ -0,0 +1,38 @@
...
+import {serializeDate, toMoment} from '../util/date';
+import AnswerSlot from './AnswerSlot';
+import AnswerMultipleSlot from './AnswerMultipleSlot';
+import styles from './Answer.module.scss';
+
+export default function AnswerCalendarDay({options}) {
+  const date = serializeDate(toMoment(options.date, 'YYYY-MM-DD'), 'dddd D MMM');
+  return (
+    <>
+      <Header as="h3" className={styles['date']}>
```
ThiefMaster: `styles.date`

newdle/client/src/components/AnswerOption.js
```diff
@@ -0,0 +1,31 @@
+import {Icon} from 'semantic-ui-react';
+import {serializeDate, toMoment} from '../util/date';
+import PropTypes from 'prop-types';
+import React from 'react';
+import styles from './Answer.module.scss';
+
+export default function AnswerOption({startTime, endTime, icon, onClick, style}) {
```
ThiefMaster: I'd call the `style` prop `className` instead.

newdle/client/src/components/Title.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import {Container, Header} from 'semantic-ui-react';
+import styles from './Title.module.scss';
+export default function Title() {
```
ThiefMaster: I think if we go for a component name like this we absolutely need to start moving stuff to a subdirectory, e.g. `components/answer/Title.js`.

newdle/client/src/components/Title.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import {Container, Header} from 'semantic-ui-react';
+import styles from './Title.module.scss';
```
ThiefMaster: Missing linebreak after this line.

newdle/client/src/reducers/answer.js
```diff
@@ -0,0 +1,26 @@
+import {combineReducers} from 'redux';
+import {ADD_ANSWER, REMOVE_ANSWER, SET_ANSWER_ACTIVE_DATE} from '../actions';
+
+export default combineReducers({
+  answers: (state = {}, action) => {
+    switch (action.type) {
+      case ADD_ANSWER:
+        return {...state, [action.timeslot]: action.answer};
+      case REMOVE_ANSWER: {
+        const {[action.timeslot]: value, ...answers} = state;
```
ThiefMaster: this looks broken, what's `value`?

newdle/client/src/components/AnswerMultipleSlot.js
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react';
+import {useDispatch} from 'react-redux';
+import AnswerOption from './AnswerOption';
+import styles from './Answer.module.scss';
+
+export default function MultipleAnswerSlot({height, pos, options}) {
+  const dispatch = useDispatch();
+  return (
+    <div className={styles['multiple-answer-slot']} style={{top: `${pos}%`, height: `${height}%`}}>
+      {options.map(option => {
```
mic4ael: `{options.map(option => <AnswerOption ...`

newdle/client/src/reducers/answer.js
meluru: It works, I'm basically removing one option from answers this way. But probably there's a better way to do that

ThiefMaster: Oh right, you are destructuring here and not returning the object.

ThiefMaster: But what does "removing an answer" mean? Isn't this the same as being unavailable?

meluru: Yep, I simply wasn't sure which way to go with unavailable answers in the redux state. 
Right now I'm storing only accepted and ifneedbe and omitting the unavailable. I could store the unavailable as well, but then I would always need to fill the state with 'unavailable' for each newdle time slot at the begining.

newdle/client/src/components/AnswerCalendar.js
meluru: Yes, it will change I believe.

newdle/client/src/reducers/answer.js
ThiefMaster: Just filling anything missing with unavailable when saving the answers would work as well - but when you edit your existing response, then unavailable slots *will* be explicitly set to unavailable as well, so any display logic needs to be able to handle explicitly-unavailable slots.

meluru: I didn't take editing into account, I see your point.

newdle/client/src/components/answer/MultipleSlot.js
```diff
@@ -0,0 +1,22 @@
...
+        <Option {...option} key={option.slot} onClick={() => dispatch(option.action())} />
+      ))}
+    </div>
+  );
+}
+
+MultipleSlot.propTypes = {
+  height: PropTypes.number.isRequired,
+  pos: PropTypes.number.isRequired,
+  options: PropTypes.array.isRequired,
```
ThiefMaster: Since you are unpacking the objects from the array, I'd prefer an `arrayOf(PropTypes.shape({...}))` here

newdle/client/src/components/answer/AnswerHeader.module.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../styles/palette';
+
+.title-container {
+  margin: 2px;
+  margin: 2px;
```
mic4ael: Side effect of a merge? :P

meluru: Thanks for noticing :)

PR #62, opened 2019-09-25
Time slots improvements

### LINKED ISSUE #51: Style time slots a bit better
* Show clearly that two adjacent slots are not connected;
 * Show start/end times on hover.

mvidalgarcia:
closes #51

### DIFF
--- a/newdle/client/src/components/Slot.js
+++ b/newdle/client/src/components/Slot.js
@@ -1,16 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import {Popup} from 'semantic-ui-react';
 import styles from './Timeline.module.scss';
 
-export default function Slot({width, pos, moreStyles, onClick, children}) {
+export default function Slot({width, pos, moreStyles, onClick, children, tooltip}) {
   return (
-    <div
-      onClick={onClick}
-      className={`${styles['slot']} ${moreStyles}`}
-      style={{left: `${pos}%`, width: `${width}%`}}
-    >
-      {children}
-    </div>
+    <Popup
+      trigger={
+        <div
+          onClick={onClick}
+          className={`${styles['slot']} ${moreStyles}`}
+          style={{left: `${pos}%`, width: `${width}%`}}
+        >
+          {children}
+        </div>
+      }
+      content={tooltip}
+      disabled={!tooltip}
+    />
   );
 }
 
@@ -20,10 +27,12 @@ Slot.propTypes = {
   moreStyles: PropTypes.string,
   onClick: PropTypes.func,
   children: PropTypes.node,
+  tooltip: PropTypes.string,
 };
 
 Slot.defaultProps = {
   moreStyles: '',
   onClick: null,
   children: null,
+  tooltip: null,
 };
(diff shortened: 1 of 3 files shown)


newdle/client/src/components/Timeline.js
```diff
@@ -233,19 +234,56 @@ function TimelineContent({busySlots, minHour, maxHour}) {
...
+  const [minHour, setMinHour] = useState(defaultMinHour);
+  const [maxHour, setMaxHour] = useState(defaultMaxHour);
+  const candidates = useSelector(getTimeslotsForActiveDate);
+  const duration = useSelector(getDuration);
   const hourSeries = _.range(minHour, maxHour + hourStep, hourStep);
   const hourSpan = maxHour - minHour;
+  const defaultHourSpan = defaultMaxHour - defaultMinHour;
   const busySlots = calculateBusyPositions(availability, minHour, maxHour);
+
+  const computeHourSpan = useCallback(() => {
```
ThiefMaster: I don't think `useCallback` is needed here - you are not passing the function to anything so there is no benefit in preserving its identity between renders.

mvidalgarcia: ESLint complains then 
> The 'computeHourSpan' function makes the dependencies of useEffect Hook (at line 285) change on every render. Move it inside the useEffect callback. Alternatively, wrap the 'computeHourSpan' definition into its own useCallback() Hook.

ThiefMaster: ah it's a dep of useEffect. then you do need the useCallback

newdle/client/src/components/Slot.js
```diff
@@ -1,16 +1,23 @@
...
+export default function Slot({width, pos, moreStyles, onClick, children, tooltip}) {
   return (
-    <div
-      onClick={onClick}
-      className={`${styles['slot']} ${moreStyles}`}
-      style={{left: `${pos}%`, width: `${width}%`}}
-    >
-      {children}
-    </div>
+    <Popup
```
mic4ael: How about using `bottom center` as the `Popup`'s position?

newdle/client/src/components/Timeline.js
```diff
@@ -233,19 +234,52 @@ function TimelineContent({busySlots, minHour, maxHour}) {
...
+  const defaultHourSpan = defaultMaxHour - defaultMinHour;
   const busySlots = calculateBusyPositions(availability, minHour, maxHour);
+
+  useEffect(() => {
+    const resetHourSpan = () => {
+      setMinHour(defaultMinHour);
+      setMaxHour(defaultMaxHour);
+    };
+
+    const candidatesMoment = candidates.map(c => moment(c, 'HH:mm'));
```
mic4ael: Move it inside the `if` and check for `candidates.length` instead?

meluru: We have this `toMoment ` function in util/date. I think it would be more consistent to use it.

meluru: We have this `toMoment` function in util/date.js. I guess it would be more consistent to use it also here

mvidalgarcia: Didn't know, thanks!

PR #69, opened 2019-09-27
Fix slot rendering when end time > 24h

### LINKED ISSUE #67: Timeslots touching midnight look broken
image

This happens as soon as the timeslot ends at/after midnight.

pferreir:
image

Closes #67


ThiefMaster:
I think the "expanding" logic is still slightly broken: Try adding one slot at 00:00 and one at 23:00:

image

PR #70, opened 2019-09-27
New readme

mic4ael:


README.md
```diff
@@ -1,38 +1,81 @@
...
+
+**Newdle** is the new, shiny tool brought to you by the [Indico Team](https://getindico.io/about) @ [CERN](https://home.cern) 🎉
+Here at CERN we **hold a lot of meetings**. While [Indico](https://getindico.io) makes it super easy to manage those meetings, we still **lose a lot of time** 📈 trying to schedule them, which usually involves numerous emails and private messages. That is what newdle has been created for: to streamline the process of choosing **the perfect date and time** 🗓 for your next meeting/event.
+
+Newdle is part of the [MALT project](https://malt.web.cern.ch).
+
+## Why another tool?
+
+
+It is true that there is an abundance of similar tools like e.g. [Nuages](https://nuages.domainepublic.net) or [croodle](https://github.com/jelhan/croodle) which are just two of many others available out there. Despite that, we still believe that newdle will provide users with more features and will offer much better user experience. We are putting huge efforts in order to be sure that newdle will meet the following requirements:
```
pferreir: It's quite nice, but I would give more importance to the items below. It looks almost as if our differentiating factor is UX and then those things are just extras...
Plus, I don't think there's an "abundance" of similar tools. I would say something like "there are several commercial projects (including Doodle, which was our main inspiration) and a handful of Open Source projects with similar functionality and in different states of development (or abandonment). However none of them provide bla bla bla..."

README.md
```diff
@@ -1,38 +1,81 @@
...
+Before running the alembic migrations make sure you  have created a database called `newdle` (or adjust the config file). Having done that, run `flask db upgrade` to upgrade the schema.
+
+**:warning: Schema changes :warning:**
+
+At the initial stages of the development it might be pretty common to change the DB schema which also incurs modifications to the SQLAlchemy models. To facilitate the process of keeping actual schema and models in sync we provide `make newdb` command. The purpose of it is to update the initial alembic revision (according to the models declared) and recreate your local database. This is a destructive operation! This command will be removed as soon as there are more alembic revisions.
+
+
+## Running the development server
+
+To run the dev servers, use `make flask-server` and `make react-server` (in separate terminals). You can use the `FLASK_HOST`, `FLASK_PORT` and `REACT_PORT` environment variables to override where the dev servers will listen (make sure to set it for both dev servers, since the react server needs to know where the Flask app is running).
```
mvidalgarcia:
```suggestion
To run the dev servers, use `make flask-server` and `make react-server` (in separate terminals). You can use the `FLASK_HOST`, `FLASK_PORT` and `REACT_PORT` environment variables to override where the dev servers will listen (make sure to set it for both dev servers, since the React server needs to know where the Flask app is running).
```

README.md
```diff
@@ -1,38 +1,81 @@
...
+## Other available `make` targets
 
-The unit tests can be run using `make test`. Linting (with pycodestyle) can be done using `make lint`.
+We provide a couple of additional `make` targets that should streamline the development process:
 
-After updating the database structure, run `make newdb` to update the initial alembic revision
-and recreate your local database. This is a destructive operation! This command will be removed
-as soon as there are more alembic revisions.
+ - `make clean` - removes all generated files
+ - `make distclean` - runs `clean` target first and removes config files afterwards;
```
mvidalgarcia: I'd add semicolon on either all the bullet points or none.

ThiefMaster: +1 on "none"

mic4ael: Oh, good catch. I don't know how I ended up with only one semicolon :P I will remove it ofc.

PR #71, opened 2019-09-27
Make timezone selector dropdown lazy

ThiefMaster:
It takes quite long to render, and most people won't change its value, so now we just render some simple HTML and only turn it into a real dropdown when clicking it.

### DIFF
--- a/newdle/client/src/components/LazyDropdown.js
+++ b/newdle/client/src/components/LazyDropdown.js
@@ -4,44 +4,12 @@ import {Dropdown} from 'semantic-ui-react';
 export default function LazyDropdown(props) {
   const [activated, setActivated] = useState(false);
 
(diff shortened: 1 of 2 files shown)


newdle/client/src/components/LazyDropdown.js
```diff
@@ -0,0 +1,49 @@
...
+  const [activated, setActivated] = useState(false);
+
+  const handleActivate = () => {
+    setActivated(true);
+  };
+
+  if (!activated) {
+    const activeText = props.value ? props.options.find(x => x.value === props.value).text : '';
+    return (
+      /* eslint-disable jsx-a11y/role-has-required-aria-props */
```
ThiefMaster: (it's complaining about an aria prop that's not added by SUIR either)

PR #74, opened 2019-10-01
Summary page

### LINKED ISSUE #27: "Newdle summary" page (basic page only)
This page should serve both as the "final page" after a newdle gets created and something the creator can come back to to see the status of the newdle. It should display:
 * a summary of the participants and selected options as well as their choices;
 * the link to the newdle that can be shared with the participants;

image

image

panagiotappl:
Implements page where newdle organizer selects the final date of the meeting and summary page after the final date has been set (which is almost empty for now).

Closes #27


newdle/client/src/reducers/index.js
```diff
@@ -16,6 +16,8 @@ export default combineReducers({
     switch (action.type) {
       case NEWDLE_RECEIVED:
         return action.newdle;
+      case NEWDLE_UPDATED:
```
mic4ael:
```javascript
case NEWDLE_RECEIVED:
case NEWDLE_UPDATED:
    return action.newdle;
```

newdle/client/src/selectors.js
```diff
@@ -43,6 +43,8 @@ export const getFullTimeslots = state =>
 export const getCreatedNewdle = state => state.creation.createdNewdle;
 
 // newdle
+export const getNewdle = state => state.newdle;
+export const getNewdleTitle = state => (state.newdle && state.newdle.title) || null;
```
mic4ael: Maybe we could combine those functions that use `state.newdle` with `createSelector`? And handle cases in which `state.newdle === null` there

newdle/client/src/components/Summary.module.scss
```diff
@@ -0,0 +1,37 @@
...
+
+    &:hover {
+      background-color: teal;
+      color: $white;
+    }
+  }
+
+  .create-event-button {
+    background-color: #00394a;
+    color: $white;
```
mic4ael: New line after that attribute?

newdle/client/src/components/Summary.js
```diff
@@ -0,0 +1,80 @@
...
+    } catch (exc) {
+      setSubmitting(false);
+      setError(exc.toString());
+      return;
+    }
+    dispatch(updateNewdle(updatedNewdle));
+    setSubmitting(false);
+  }
+
+  return (
```
mic4ael: Shoulnd't you have a check before this return?

```javascript
if (!newdle) {
    return null;
}
```

newdle/client/src/components/Summary.js
```diff
@@ -0,0 +1,80 @@
...
+        </>
+      )}
+      {newdle && newdle.final_dt && (
+        <>
+          <div className={styles['summary-container']}>
+            <div className={styles['datetime']}>
+              <Icon name="calendar alternate outline" />
+              {serializeDate(newdle.final_dt, 'MMMM Do YYYY')}
+            </div>
+
```
mic4ael: Is it on purpose here?

newdle/client/src/components/Summary.js
```diff
@@ -0,0 +1,80 @@
...
+import ParticipantTable from './ParticipantTable';
+import {clearFinalDate} from '../actions';
+import {getNewdle, getFinalDate} from '../selectors';
+import {updateNewdle} from '../actions';
+import client from '../client';
+import styles from './Summary.module.scss';
+
+export default function Summary() {
+  const [submitting, setSubmitting] = useState(false);
+  const [error, setError] = useState('');
```
mic4ael: Shouldn't the initial value be `null`?

newdle/client/src/components/ParticipantTable.js
```diff
@@ -63,11 +69,12 @@ function AvailabilityRow({availability: {startDt, available}, setActiveDate, act
   const numberOfParticipants = useSelector(getNumberOfParticipants);
   const duration = useSelector(getNewdleDuration);
   const startTime = toMoment(startDt, 'YYYY-MM-DDTHH:mm');
+  const dispatch = useDispatch();
 
   return (
     <Table.Row
       className={styles['participant-row']}
-      onClick={() => setActiveDate(startDt)}
+      onClick={() => dispatch(setActiveDate(startDt))}
```
mic4ael: Is it something that should be here actually? I think that `ParticipantTable` shouldn't be using `dispatch`. You can just pass a callback here that uses `dispatch` internally.

newdle/api.py
```diff
@@ -124,6 +125,19 @@ def get_newdle(code):
     return schema_cls().jsonify(newdle)
 
 
+@api.route('/newdle/<code>', methods=('PATCH',))
+@use_args(UpdateNewdleSchema(), locations=('json',))
+def update_newdle(args, code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    restricted = not g.user or newdle.creator_uid != g.user['uid']
```
ThiefMaster: Since this doesn't have `@allow_anonymous`, you don't need to check for not having `g.user`, so you can just check `if newdle.creator_uid != g.user['uid']: ...` in the next line

newdle/client/src/components/Summary.js
```diff
@@ -0,0 +1,80 @@
...
+  const newdle = useSelector(getNewdle);
+  const finalDate = useSelector(getFinalDate);
```
ThiefMaster: I don't think we need this in redux at all. It's never used outside this component, so it could stay in local state

newdle/client/src/components/Summary.js
```diff
@@ -0,0 +1,80 @@
...
+      setError(exc.toString());
+      return;
+    }
+    dispatch(updateNewdle(updatedNewdle));
+    setSubmitting(false);
+  }
+
+  return (
+    <Container text>
+      {newdle && !newdle.final_dt && (
```
ThiefMaster: I'd bail out early with a Loader until `newdle` exists instead of checking it here and below.

newdle/client/src/client.js
```diff
@@ -43,6 +43,13 @@ class Client {
     return this._request(flask`api.get_newdle`({code}), {anonymous: !fullDetails});
   }
 
+  updateNewdle(code, newdle) {
```
ThiefMaster: What do you think about making this a more specific `setFinalDT(code, dt)` operation? The underlying call would not change but I think as far as the Client is concerned this is a separate operation from "just" updating a newdle (especially since that's something we don't support)

newdle/client/src/components/NewdleHeader.js
```diff
@@ -0,0 +1,28 @@
...
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(fetchNewdle(code, true));
+  }, [code, dispatch]);
+
+  return (
+    <Container text className={styles['summary']}>
+      {title && user && (
+        <div className={styles['header']}>
```
ThiefMaster: `styles.header` etc.

newdle/client/src/client.js
panagiotappl: I thought that we were going to support updating a newdle in future, so this operation could have more uses than just setting the final date.

newdle/client/src/components/Summary.js
```diff
@@ -0,0 +1,76 @@
...
+import styles from './Summary.module.scss';
+
+export default function Summary() {
+  const [submitting, setSubmitting] = useState(false);
+  const [error, setError] = useState('');
+  const [finalDate, setFinalDate] = useState(null);
+  const newdle = useSelector(getNewdle);
+  const dispatch = useDispatch();
+
+  async function update() {
```
ThiefMaster: usually we use arrow functions when nesting functions -> `const update = async () => {`

newdle/client/src/components/Summary.js
```diff
@@ -0,0 +1,76 @@
...
+    setError('');
+    setSubmitting(true);
+    let updatedNewdle;
+    try {
+      updatedNewdle = await client.setFinalDt(newdle.code, {final_dt: finalDate});
```
ThiefMaster: The second argument should just be `finalDate` now that the method is called `setFinalDt`

newdle/client/src/client.js
```diff
@@ -51,6 +51,13 @@ class Client {
     return this._request(flask`api.get_busy_times`({date, email}));
   }
 
+  setFinalDt(code, newdle) {
```
ThiefMaster: `setFinalDT` (`Dt` looks weird) or `setFinalDate`

ThiefMaster: the second argument should just be the date, not the whole newdle

newdle/client/src/components/App.js
```diff
@@ -23,6 +24,7 @@ export default function App() {
           <Route exact path="/new" component={CreateNewdle} />
           <Route exact path="/new/success" component={NewdleCreated} />
           <Route exact path="/mine" component={MyNewdles} />
+          <Route path="/newdle/:code/summary" component={Summary} />
```
ThiefMaster: `exact`

tests/models_test.py
```diff
@@ -14,12 +14,11 @@ def test_create_newdle(dummy_newdle):
 
 def test_set_newdle_final(dummy_newdle, db_session):
     newdle = Newdle.query.get(dummy_newdle.id)
-    local_tz = timezone('Europe/Zurich')
```
ThiefMaster: check for unused imports ;)

newdle/api.py
```diff
@@ -166,6 +167,18 @@ def get_newdle(code):
     return schema_cls().jsonify(newdle)
 
 
+@api.route('/newdle/<code>', methods=('PATCH',))
+@use_args(UpdateNewdleSchema(), locations=('json',))
+def update_newdle(args, code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    if newdle.creator_uid != g.user['uid']:
+        return jsonify(error='forbidden'), 403
```
ThiefMaster: you can just `raise Forbidden` here - this will already result in a jsonified error with the default message of the error (which is fine here, since it should never happen anyway)

newdle/client/src/components/NewdleHeader.js
```diff
@@ -0,0 +1,34 @@
...
+import {useDispatch, useSelector} from 'react-redux';
+import {Container, Label} from 'semantic-ui-react';
+import {getNewdleTitle, getUserInfo, getNewdleFinalDt} from '../selectors';
+import {fetchNewdle} from '../actions';
+import styles from './NewdleHeader.module.scss';
+
+export default function NewdleHeader({match}) {
+  const code = match.params.code;
+  const title = useSelector(getNewdleTitle);
+  const user = useSelector(getUserInfo);
```
ThiefMaster: that's not correct, it returns your own user info, not the newdle creator. we don't store the name of the newdle creator so far (just an identifier), so I'd either fully omit it or put a hardcoded dummy name below (and add a task to store the creator's name when creating a newdle)

newdle/client/src/components/Summary.js
ThiefMaster: it's the error message as a string, so empty string seems like a good default to me

newdle/client/src/components/NewdleHeader.js
panagiotappl: I used this because I thought that this summary page will only be seen by the currently logged in user (you can only see the summary of your personal newdles)

ThiefMaster: ok, good point!

but then the whole message is a bit useless IMHO

panagiotappl: It might be useless in the end. I was following the mockups and imo without this the title looks a bit empty.

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -0,0 +1,76 @@
...
+        </>
+      )}
+      {newdle.final_dt && (
+        <>
+          <div className={styles['summary-container']}>
+            <div className={styles['datetime']}>
+              <Icon name="calendar alternate outline" />
+              {serializeDate(newdle.final_dt, 'MMMM Do YYYY')}
+            </div>
+            <div className={styles['datetime']}>
```
meluru: `styles.datetime` ?

panagiotappl: I'm not sure whether it's a good idea to mix two different styles in the same file.

ThiefMaster: At least I prefer using the bracket syntax only if needed because of dashes etc.

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -0,0 +1,76 @@
...
+      setSubmitting(false);
+      setError(exc.toString());
+      return;
+    }
+    dispatch(updateNewdle(updatedNewdle));
+    setSubmitting(false);
+  };
+
+  if (!newdle) {
+    return <Loader />;
```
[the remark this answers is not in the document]
panagiotappl: Fixed it. I'm still not sure whether a single loader there is the best idea.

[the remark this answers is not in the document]
ThiefMaster: We can always make it prettier later ;) returning `null` would also be an option for now

PR #75, opened 2019-10-01
Update `README.md`

### LINKED ISSUE #63: Improve README
Some things which are missing:
 * app logo;
 * badges (only TravisCI for now?);
 * small intro saying what the app does;
 * CERN footer like the one we have on the Indico README;
  * ... any others?

mic4ael:
closes #63


README.md
```diff
@@ -1,38 +1,76 @@
...
...f our main inspirations when starting the development process. The biggest advantage of newdle over it is the fact that newdle is free of charge and Open Source - you just have to set it up on your server and you can start newdling! Apart from the commercial products on the market, there is a handful of Open Source projects e.g. [Nuages](https://nuages.domainepublic.net) or [croodle](https://github.com/jelhan/croodle) which differ a lot in the features they offer. We have noticed that the majority of those tools don't provide users with proper User Experience which makes using them harder than it should be. This is one of the major advantages of newdle. We made sure that our tool is easy to use and intuitive with really good-looking design. Since we are also the developers behind [Indico](https://getindico.io) we thought of integrating newdle with it. So, for example after a poll/newdle has finished it will be possible to create an event in [Indico](https://getindico.io). Also newdle will make it a lot easier to choose possible slots for your meeting since we will integrate it with enterprise calendars to fetch participant availability. While making sure we still protect your data.
```
pferreir:
```suggestion
It is true that there are already several commercial and Open Source solutions available that facilitate creating "polls". One of them is [Doodle](https://doodle.com) which was one of our main inspirations when starting the development process. The biggest advantage of newdle over it is the fact that newdle is free of charge and Open Source - you just have to set it up on your server and you can start newdling! Apart from the commercial products on the market, there is a handful of Open Source projects e.g. [Nuages](https://nuages.domainepublic.net) or [croodle](https://github.com/jelhan/croodle) which differ a lot in the features they offer. We have noticed that the majority of those tools don't provide users with proper User Experience which makes using them harder than it should be. This is one of the major advantages of newdle. We made sure that our tool is easy to use and intuitive with really good-looking design.

## Integration
**This is still ongoing work**
Since we are also the developers behind [Indico](https://getindico.io) we thought of integrating newdle with it. So, for example after a poll/newdle has finished it will be possible to create an event in [Indico](https://getindico.io). Also newdle will make it a lot easier to choose possible slots for your meeting since we will integrate it with enterprise calendars to fetch participant availability, while making sure we still protect your data.
```

pferreir: I would just split the last part into a different section. Otherwise it looks perfect to me, thanks!

README.md
```diff
@@ -1,38 +1,82 @@
...
+ * [ ] Integration with calendaring systems;
+ * [ ] Integration with Indico;
 
-This is something new and cool! 🎉
+If you would like to help with any of that, please contact us through the relevant GitHub issue.
 
-## Database
+# Newdle [![Travis Build Status](https://travis-ci.com/indico/newdle.svg?branch=master)](https://travis-ci.org/indico/newdle) [![License](https://img.shields.io/github/license/indico/newdle.svg)](https://github.com/indico/newdle/blob/master/LICENSE) [![Made at CERN!](https://img.shields.io/badge/CERN-Open%20Source-%232980b9.svg)](https://home.cern)
+
+<p align="center">
```
ThiefMaster: Is there no way to avoid putting HTML in there?!

README.md
```diff
@@ -1,38 +1,82 @@
-# Newdle
+## WARNING: newdle is not yet production ready
+### Currently missing
+ * [ ] E-mail sending;
```
ThiefMaster: can we skip those weird semicolons? :D

ThiefMaster: And I'm pretty sure people won't update this list so I'd rather just mention it's still under heavy development and not ready for real usage.

README.md
```diff
@@ -1,38 +1,82 @@
...
+ * [ ] Integration with Indico;
 
-This is something new and cool! 🎉
+If you would like to help with any of that, please contact us through the relevant GitHub issue.
 
-## Database
+# Newdle [![Travis Build Status](https://travis-ci.com/indico/newdle.svg?branch=master)](https://travis-ci.org/indico/newdle) [![License](https://img.shields.io/github/license/indico/newdle.svg)](https://github.com/indico/newdle/blob/master/LICENSE) [![Made at CERN!](https://img.shields.io/badge/CERN-Open%20Source-%232980b9.svg)](https://home.cern)
+
+<p align="center">
+  <img src="./newdle/client/src/images/logo_color.svg" width="300"><br>
```
ThiefMaster: relative paths here seem very risky

README.md
```diff
@@ -1,38 +1,82 @@
-# Newdle
+## WARNING: newdle is not yet production ready
```
ThiefMaster: Having this above the "Newdle" title looks weird IMO

mic4ael: Isn't that the point? It should be on top so that it is visible?

README.md
mic4ael: I don't think there is another way to do that

README.md
mic4ael: Hm, I could either move it to the root path of the repo or to the `assets` repo and reference it by here via URL.

PR #78, opened 2019-10-01
Make "back" buttons nicer

### LINKED ISSUE #64: Improve "change xxxx" options in last step
They currently look like this:

image

I think it's not very readable and we could probably come up with something better. One option would be two buttons, one with a one-arrow icon and the other one with a "fast" forward double-arrow icon.
I also think that making step indicators clickable would help a great deal with UX.

image

pferreir:
The "back" buttons now look nicer:
image

The `Navigator` was also improved. The steps are now clickable:
image

Closes #64

### DIFF
--- a/newdle/client/src/components/Navigator.js
+++ b/newdle/client/src/components/Navigator.js
@@ -1,9 +1,11 @@
+import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import {Container} from 'semantic-ui-react';
-import {useSelector} from 'react-redux';
-import {getStep} from '../selectors';
+import {useDispatch, useSelector} from 'react-redux';
+import {getStep, getFullTimeslots} from '../selectors';
 import styles from './Navigator.module.scss';
+import {setStep} from '../actions';
 
 export default function Navigator() {
   const steps = [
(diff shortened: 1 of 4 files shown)


newdle/client/src/components/Navigator.js
```diff
@@ -21,34 +22,54 @@ export default function Navigator() {
...
-          {index + 1}
-        </Step>
-      ))}
+      <ul className={styles['step-box']}>
+        {steps.map((_, index) => (
+          <Step
+            key={index}
+            index={index + 1}
+            active={index + 1 === active}
+            onClick={index + 1 < active ? () => dispatch(setStep(index + 1)) : null}
```
ThiefMaster: I'd allow navigating forward as well, and only disallow going to step 3 if there are no timeslots defined (could add a condition to the corresponding `steps` object that runs the selector, like we do to disable the next button on the step's page)

pferreir: Done.

newdle/client/src/components/Navigator.js
```diff
@@ -21,34 +23,56 @@ export default function Navigator() {
...
-        </Step>
-      ))}
+      <ul className={styles['step-box']}>
+        {steps.map((__, index) => (
+          <Step
+            key={index}
+            index={index + 1}
+            active={index + 1 === active}
+            // Step 3 (index === 2) should only be clickable if there are already slots defined
+            onClick={index !== 2 || !_.isEmpty(slots) ? () => dispatch(setStep(index + 1)) : null}
```
ThiefMaster: Passing `null` doesn't use the default prop, only `undefined` does.

pferreir: Does it really matter, though? The child component doesn't misbehave when `onClick=null`.

pferreir: But OK, it's prettier, since otherwise the default value would never be used...

ThiefMaster: no, i just noticed that you are specifying a default that's never used

pferreir: Yeah, I'm fixing it, thanks 😉

PR #73, opened 2019-10-01
Add backend for "My Newdles" page

ThiefMaster:


newdle/client/src/components/MyNewdles.js
```diff
@@ -1,18 +1,55 @@
...
 import styles from './MyNewdles.module.scss';
+import client from '../client';
 
-function MyNewdle({
-  newdle: {code, title, participantsCount, participantsTotal, duration, finalDt, timezone},
-}) {
+export default function MyNewdles() {
+  const [newdles, setNewdles] = useState(null);
+
+  useEffect(() => {
```
ThiefMaster: I did not put this in redux since the data is only needed on this page, so using redux here would just make things more complex.

PR #94, opened 2019-10-11
Add anonymous participants

### LINKED ISSUE #81: Answer page: add button to submit answer
This should obviously send a request to submit the answer.

### LINKED ISSUE #90: Answer page: anonymous participant
Right now we are required to have a "participant code". In the case of a user who is given the link to the newdle, we won't have neither their name nor availability. We should provide a page which asks for their name and shows no availability info.

pferreir:
Closes #81 and closes #90

image

image

image

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -14,6 +14,7 @@ from .core.util import DATE_FORMAT, format_dt
 from .core.webargs import abort, use_args, use_kwargs
 from .models import Newdle, Participant
 from .schemas import (
+    CreateAnonymousParticipantSchema,
     MyNewdleSchema,
     NewdleSchema,
     NewNewdleSchema,
@@ -214,5 +215,16 @@ def update_participant(args, code, participant_code):
             )
     for key, value in args.items():
         setattr(participant, key, value)
+    db.session.commit()
+    return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/participants', methods=('POST',))
+@allow_anonymous
+@use_args(CreateAnonymousParticipantSchema(), locations=('json',))
+def create_anonymous_participant(args, code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    participant = Participant(newdle=newdle, **args)
+    newdle.participants.add(participant)
     db.session.commit()
     return ParticipantSchema().jsonify(participant)
--- a/newdle/client/src/client.js
+++ b/newdle/client/src/client.js
@@ -1,6 +1,20 @@
 import flask from 'flask-urls.macro';
+import {useReducer} from 'react';
 import {getToken, isAcquiringToken} from './selectors';
 import {tokenExpired, tokenNeeded} from './actions';
+
+function backendReducer(state, action) {
+  switch (action.type) {
+    case 'submit':
+      return {...state, submitting: true, error: '', result: null};
+    case 'error':
+      return {...state, submitting: false, error: action.error};
+    case 'success':
+      return {...state, submitting: false, result: action.result};
+    default:
+      return state;
+  }
+}
 
 class ClientError extends Error {
   constructor(url, code, message, data = null) {
@@ -21,6 +35,46 @@ class Client {
       throw new Error('Tried to use client that is not connected to a store');
     }
     return getToken(this.store.getState());
+  }
+
+  /*
+   * This method returns a React hook which handles the possible states regarding backend
+   * communication. It's possible to execute several backend API calls sequentially.
+   * The results of the first call will be fed as parameters for the second, and so on...
+   *
+   * @param {Function} funcs - the client functions which will be invoked, in order. The
+   * first one will receive the `call(...)` parameters, while the others will be fed the
+   * results of the previous one.
+   */
+  useBackend(...funcs) {
+    const [state, dispatch] = useReducer(backendReducer, {
+      submitting: false,
+      error: '',
+      result: null,
+    });
+
+    const call = async (...params) => {
+      const end = async result => {
+        dispatch({type: 'success', result});
+        return result;
+      };
+
+      dispatch({type: 'submit'});
+      const f = funcs.reverse().reduce((prevPromise, func) => {
+        return async (...params) => {
+          try {
+            const result = await func.bind(this)(...params);
+            return await prevPromise(result);
+          } catch (exc) {
+            dispatch({type: 'error', error: exc.toString()});
+          }
+        };
+      }, end);
+
+      return await f(...params);
+    };
+
+    return [call, state.submitting, state.error, state.result];
   }
 
   getMe() {
@@ -63,16 +117,16 @@ class Client {
     return this._request(flask`api.get_participant`(params), {anonymous: true});
   }
 
-  saveParticipantAvailability(newdleCode, participantCode, availability) {
+  answerNewdleParticipant = (newdleCode, participantCode, availability) => {
     const params = {code: newdleCode, participant_code: participantCode};
-    return this._request(flask`api.get_participant`(params), {
+    return this._request(flask`api.update_participant`(params), {
       anonymous: true,
       method: 'PATCH',
       body: JSON.stringify({
         answers: availability,
       }),
     });
-  }
+  };
 
   async _request(url, options = {}, withStatus = false, isRetry = false) {
     const headers = {Accept: 'application/json'};
--- a/newdle/client/src/client.js
+++ b/newdle/client/src/client.js
@@ -117,13 +117,25 @@ class Client {
     return this._request(flask`api.get_participant`(params), {anonymous: true});
   }
 
-  answerNewdleParticipant = (newdleCode, participantCode, availability) => {
+  answerNewdleParticipant = (newdleCode, participantCode, answers) => {
     const params = {code: newdleCode, participant_code: participantCode};
     return this._request(flask`api.update_participant`(params), {
       anonymous: true,
       method: 'PATCH',
       body: JSON.stringify({
-        answers: availability,
+        answers,
+      }),
+    });
+  };
+
+  answerNewdleAnonymous = (newdleCode, participantName, answers) => {
+    const params = {code: newdleCode};
+    return this._request(flask`api.create_anonymous_participant`(params), {
+      anonymous: true,
+      method: 'POST',
+      body: JSON.stringify({
+        answers,
+        name: participantName,
       }),
     });
   };
--- a/newdle/client/src/client.js
+++ b/newdle/client/src/client.js
@@ -117,7 +117,7 @@ class Client {
     return this._request(flask`api.get_participant`(params), {anonymous: true});
   }
 
-  answerNewdleParticipant = (newdleCode, participantCode, answers) => {
+  updateParticipantAnswers(newdleCode, participantCode, answers) {
     const params = {code: newdleCode, participant_code: participantCode};
     return this._request(flask`api.update_participant`(params), {
       anonymous: true,
@@ -126,19 +126,18 @@ class Client {
         answers,
       }),
     });
-  };
+  }
 
-  answerNewdleAnonymous = (newdleCode, participantName, answers) => {
+  createAnonymousParticipant(newdleCode, participantName) {
     const params = {code: newdleCode};
     return this._request(flask`api.create_anonymous_participant`(params), {
       anonymous: true,
       method: 'POST',
       body: JSON.stringify({
-        answers,
         name: participantName,
       }),
     });
-  };
+  }
 
   async _request(url, options = {}, withStatus = false, isRetry = false) {
     const headers = {Accept: 'application/json'};
--- a/newdle/client/src/components/App.js
+++ b/newdle/client/src/components/App.js
@@ -25,7 +25,7 @@ export default function App() {
           <Route exact path="/new/success" component={CreationSuccessPage} />
           <Route exact path="/mine" component={MyNewdles} />
           <Route path="/newdle/:code/summary" component={SummaryPage} />
-          <Route exact path="/newdle/:code/:partcode" component={AnswerPage} />
+          <Route exact path="/newdle/:code/:partcode?" component={AnswerPage} />
           <Route render={() => 'This page does not exist'} />
         </Switch>
         <LoginPrompt />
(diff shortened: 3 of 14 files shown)


newdle/api.py
```diff
@@ -216,3 +217,25 @@ def update_participant(args, code, participant_code):
...
     db.session.commit()
     return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/participants', methods=('POST',))
+@allow_anonymous
+@use_args(CreateAnonymousParticipantSchema(), locations=('json',))
+def create_anonymous_participant(args, code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    # Once again, we can't validate this in webargs
```
ThiefMaster: You could use this API just to create the participant (ie not accepting answers) and then submit their answers in a a separate request. That way you don't need tu duplicate this rather big piece of code.

But I guess with the hook that's a bit harder to implement...

newdle/client/src/client.js
```diff
@@ -31,13 +31,13 @@ class Client {
     return this._request(flask`api.users`({q}));
   }
 
-  createNewdle(title, duration, timezone, timeslots, participants) {
+  createNewdle = (title, duration, timezone, timeslots, participants) => {
```
ThiefMaster: That's why I avoided passing `client.something` in the hook's API I suggested on MM - I don't think having to rely on bound functions is really nice

newdle/client/src/components/App.js
```diff
@@ -25,7 +25,7 @@ export default function App() {
           <Route exact path="/new/success" component={CreationSuccessPage} />
           <Route exact path="/mine" component={MyNewdles} />
           <Route path="/newdle/:code/summary" component={SummaryPage} />
-          <Route exact path="/newdle/:code/:partcode" component={AnswerPage} />
+          <Route exact path="/newdle/:code/:partcode?" component={AnswerPage} />
```
ThiefMaster: I'm positively surprised that this works without a trailing slash when there's no partcode!

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -1,15 +1,41 @@
...
+  const participantCode = params.partcode;
   const newdle = useSelector(getNewdle);
   const numberOfTimeslots = useSelector(getNumberOfTimeslots);
   const numberOfAvailable = useSelector(getNumberOfAvailableAnswers);
+  const availabilityData = useSelector(getAnswers);
+  const [name, setName] = useState('');
+  const history = useHistory();
+  const [saved, setSaved] = useState(false);
+
+  const [_answerNewdle, submitting, error] = useClient(
```
ThiefMaster: `submitAnswer` instead of `_answerNewdle`?

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -1,15 +1,41 @@
...
+    participantCode ? client.answerNewdleParticipant : client.answerNewdleAnonymous
+  );
+
+  const canSubmit = (participantCode || name.length >= 2) && !submitting && numberOfAvailable > 0;
```
ThiefMaster: Why can't you submit with zero available? I think it makes perfect sense to explicitly state that you are not available at any time.

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -1,15 +1,41 @@
...
+
+  async function answerNewdle() {
+    const {code} = await _answerNewdle(newdle.code, participantCode || name, availabilityData);
+    history.push(`/newdle/${newdle.code}/${code}`);
```
ThiefMaster: that should only happen if you don't have a `match.partcode` yet.

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -18,19 +44,85 @@ export default function AnswerPage() {
...
+          </span>
+          <Button
+            size="large"
+            color="violet"
+            content="Send you answer"
+            disabled={saved || submitting || !canSubmit}
+            loading={submitting}
+            icon="send"
+            onClick={() => {
+              if (canSubmit) {
```
ThiefMaster: does onClick even run if you click a disabled button?

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -18,19 +44,85 @@ export default function AnswerPage() {
...
+              }
+            }}
+          />
+        </Grid.Row>
       </Grid>
     </div>
   );
 }
+
+AnswerPage.propTypes = {
```
ThiefMaster: If you use react-router's `useParams()` hooks to access the route params you have a good excuse to not bother with proptypes here :D

newdle/client/src/util/hooks.js
```diff
@@ -0,0 +1,28 @@
+import {useState} from 'react';
+
+/**
+ * This hook handles the possible states regarding backend communication.
+ *
+ * @param {Function} func - the client function which will be invoked
+ */
+export function useClient(func) {
+  const [submitting, setSubmitting] = useState(false);
+  const [error, setError] = useState('');
```
ThiefMaster: I think you could use a [`useReducer`](https://reactjs.org/docs/hooks-reference.html#usereducer) hook here to avoid having to do two separate updates

newdle/client/src/util/hooks.js
```diff
@@ -0,0 +1,28 @@
+import {useState} from 'react';
+
+/**
+ * This hook handles the possible states regarding backend communication.
+ *
+ * @param {Function} func - the client function which will be invoked
+ */
+export function useClient(func) {
```
ThiefMaster: what about putting this as an extra export into `client.js`?

newdle/client/src/components/answer/AnswerPage.js
pferreir: Indeed.

newdle/client/src/util/hooks.js
pferreir: Thought of it too, I don't have a strong opinion about it.

newdle/api.py
pferreir: Good idea.

newdle/client/src/client.js
pferreir: It's true that it's not super pretty, but in any case it's only here and it's for a good reason.

pferreir: I mean, another option could be to pass the client object as well, since it's the structure which contains the token. TBH I don't like much the way `client.js` works ☹️ It's not explicit what you are getting once you import a function, and that is (IMO) confusing.

ThiefMaster: Actually I think when using separate requests you'll need to pass a promise directly to the hook's function instead of just the args so you can do something like this:

```
_answerNewdle(async () => {
  const {partCode} = await client.createParticipant();
  return await client.saveParticipantAnswers(partCode, ...);
});
```

That way the hook can switch to submitting state, then await on the promise from the async function, and set the result returned by it in the end.

I think errors also also properly propagate to the try..catch inside the hook.

pferreir: We could as well store the tokens in redux.

ThiefMaster: I'd try to avoid putting data from the URL into redux ;) that gets very quickly very messy

ThiefMaster: > I mean, another option could be to pass the client object as well,

The disadvantage there is that now you don't have a direct reference to the function anymore, so editors don't know that the (string?) name of the function passed to the hook is that particular function...

pferreir: Wait... the token is already in Redux. Why do we even need this Client object to be stateful?

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -18,19 +44,85 @@ export default function AnswerPage() {
...
+                {numberOfAvailable} out of {numberOfTimeslots} options chosen
+              </>
+            ) : (
+              <em>No options chosen</em>
+            )}
+          </span>
+          <Button
+            size="large"
+            color="violet"
+            content="Send you answer"
```
meluru: `Send your answer`?

newdle/api.py
pferreir: Done

newdle/api.py
```diff
@@ -216,3 +217,14 @@ def update_participant(args, code, participant_code):
...
     return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/participants', methods=('POST',))
+@allow_anonymous
+@use_args(CreateAnonymousParticipantSchema(), locations=('json',))
+def create_anonymous_participant(args, code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    participant = Participant(name=args['name'], newdle=newdle)
+    db.session.add(participant)
```
ThiefMaster: I think instead of specifying `newdle` above it'd be cleaner to replace this (unnecessary) line with `newdle.participants.add(participant)`

newdle/api.py
```diff
@@ -216,3 +217,14 @@ def update_participant(args, code, participant_code):
...
     db.session.commit()
     return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/participants', methods=('POST',))
+@allow_anonymous
+@use_args(CreateAnonymousParticipantSchema(), locations=('json',))
+def create_anonymous_participant(args, code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    participant = Participant(name=args['name'], newdle=newdle)
```
ThiefMaster: `**args`?

newdle/client/src/client.js
```diff
@@ -23,6 +24,60 @@ class Client {
...
+   * communication. It's possible to execute several backend API calls sequentially.
+   * The results of the first call will be fed as parameters for the second, and so on...
+   *
+   * @param {Function} funcs - the client functions which will be invoked, in order. The
+   * first one will receive the `call(...)` parameters, while the others will be fed the
+   * results of the previous one.
+   */
+  useBackend(...funcs) {
+    const [state, dispatch] = useReducer(
+      (state, action) => {
```
ThiefMaster: AFAIK this should usually be defined outside the hook itself

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -12,14 +15,34 @@ import {
...
+    ? client.useBackend(client.updateParticipantAnswers)
+    : client.useBackend(client.createAnonymousParticipant, ({code}) =>
+        client.updateParticipantAnswers(newdle.code, code, availabilityData)
+      );
+
+  const canSubmit = (participantCode || name.length >= 2) && !submitting;
+
+  async function answerNewdle() {
+    const {code} = await submitAnswer(newdle.code, participantCode || name, availabilityData);
+    history.push(`/newdle/${newdle.code}/${code}`);
```
ThiefMaster: You can end up in an inconsistent state if the answer request fails but participant creation worked. Should we push the history change right after participant instead?

Also, this should only run if a new participant was created, or you get an extra history entry you don't need.

newdle/client/src/client.js
```diff
@@ -23,6 +24,60 @@ class Client {
     return getToken(this.store.getState());
   }
 
+  /*
+   * This method returns a React hook handles the possible states regarding backend
```
mic4ael: `returns a React hook (which|that)`?

newdle/client/src/components/answer/AnswerPage.js
pferreir: Yeah, about that second part, I had added it but then lost the change during one of the rebases.

newdle/client/src/styles/forms.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../styles/palette';
+
+%important-input {
+  width: 100%;
+  border-bottom: 1px solid $purple !important;
+  padding-bottom: 10px !important;
```
mic4ael: I don't want to nitpick but I think it would be good to have a blank line between a CSS property and a selector.

newdle/client/src/components/answer/answer.module.scss
```diff
@@ -106,3 +114,19 @@
...
+  align-items: baseline;
+
+  .options-msg {
+    transition: color 0.25s ease-out;
+    &:global(.none) {
+      color: $grey;
+    }
+
+    color: $don-juan;
+    margin-right: 1em;
```
mic4ael: Maybe keep the properties on top and other selectors after them?

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -28,24 +62,80 @@ export default function AnswerPage() {
   return (
     <div>
       <Grid container>
+        {error && (
+          <Grid.Row centered>
+            <Message error>
+              <p>Something when wrong while sending your answer:</p>
```
mic4ael: `Something went wrong`?

pferreir: Yes, something went wrong with my brain, indeed.

newdle/client/src/styles/forms.scss
pferreir: Sounds reasonale to me 😉

newdle/client/src/client.js
```diff
@@ -1,7 +1,21 @@
 import flask from 'flask-urls.macro';
+import {useReducer} from 'react';
 import {getToken, isAcquiringToken} from './selectors';
 import {tokenExpired, tokenNeeded} from './actions';
 
+function backendReducer(state, action) {
+  switch (action.type) {
+    case 'submit':
+      return {...state, submitting: true};
```
ThiefMaster: shouldn't this clear error and result?

ThiefMaster:
image

Having this text wrapped on two lines looks a bit strange IMO

ThiefMaster:
image

I like it, but I think on :hankey:-y screens or projectors the placeholder might be barely readable.

ThiefMaster:
When submitting for an invalid participant ID, you get a somewhat ugly JS error. But I guess this can be fixed later when actually fetching the participant information in case a code is provided (and keeping the UI / submit button disabled until the data has been fetched)

pferreir:
> When submitting for an invalid participant ID, you get a somewhat ugly JS error. But I guess this can be fixed later when actually fetching the participant information in case a code is provided (and keeping the UI / submit button disabled until the data has been fetched)

Yeah, I think this can be fixed later.

ThiefMaster:
The same error happens when the request to save the answers fails (even if you're submitting as an existing participant)

PR #95, opened 2019-10-11
Make 'accept all available slots' switch functional

### LINKED ISSUE #82: Answer page: implement "accept all options..."
This should turn all options to green in a non-destructive way (e.g. switching it off should come back to the previous selection state).

image

ThiefMaster:
Currently using dummy availability data.

closes #82


pferreir:
> Currently using dummy availability data.

You mean "free/busy" data 😉 We should really make sure we don't mix up the two, since the code is already becoming confusing because of that.

PR #101, opened 2019-10-15
Fetch and display busy times on answer page

### LINKED ISSUE #83: Answer page: fetch and display availability slots for current user.
Those should be displayed as per the mock-ups.

image

meluru:
Closes #83

### DIFF
--- a/newdle/client/src/actions.js
+++ b/newdle/client/src/actions.js
@@ -33,6 +33,7 @@ export const SET_ANSWER_ACTIVE_DATE = 'Change answer selected date';
 export const CHOOSE_ALL_AVAILABLE = 'Choose all slots where user is available';
 export const CHOOSE_MANUALLY = 'Manually select available slots';
 export const SET_ANSWER_BUSY_TIMES = 'Set answer busy times';
+export const PARTICIPANT_RECEIVED = 'Received participant data';
 
 export function loginWindowOpened(id) {
   return {type: LOGIN_WINDOW_OPENED, id};
@@ -172,11 +173,18 @@ export function updateNewdle(newdle) {
   return {type: NEWDLE_UPDATED, newdle};
 }
 
-export function fetchBusyTimesForAnswer(participant, dates) {
+export function fetchBusyTimesForAnswer(participantEmail, dates) {
   return async dispatch => {
     dates.forEach(async date => {
-      const times = await client.getBusyTimes(date, participant.email);
+      const times = await client.getBusyTimes(date, participantEmail);
       dispatch({type: SET_ANSWER_BUSY_TIMES, date, times});
     });
+  };
+}
+
+export function fetchParticipant(newdleCode, participantCode) {
+  return async dispatch => {
+    const participant = await client.getParticipant(newdleCode, participantCode);
+    dispatch({type: PARTICIPANT_RECEIVED, participant});
   };
 }
--- a/newdle/client/src/actions.js
+++ b/newdle/client/src/actions.js
@@ -32,6 +32,7 @@ export const REPLACE_ANSWERS = 'Replace newdle answers';
 export const SET_ANSWER_ACTIVE_DATE = 'Change answer selected date';
 export const CHOOSE_ALL_AVAILABLE = 'Choose all slots where user is available';
 export const CHOOSE_MANUALLY = 'Manually select available slots';
+export const SET_ANSWER_BUSY_TIMES = 'Set answer busy times';
 
 export function loginWindowOpened(id) {
   return {type: LOGIN_WINDOW_OPENED, id};
@@ -169,4 +170,13 @@ export function setAnswerActiveDate(date) {
 
 export function updateNewdle(newdle) {
   return {type: NEWDLE_UPDATED, newdle};
+}
+
+export function fetchBusyTimesForAnswer(participant, dates) {
+  return async dispatch => {
+    dates.forEach(async date => {
+      const times = await client.getBusyTimes(date, participant.email);
+      dispatch({type: SET_ANSWER_BUSY_TIMES, date, times});
+    });
+  };
 }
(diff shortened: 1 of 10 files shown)


newdle/client/src/actions.js
```diff
@@ -170,3 +172,19 @@ export function setAnswerActiveDate(date) {
 export function updateNewdle(newdle) {
   return {type: NEWDLE_UPDATED, newdle};
 }
+
+export function fetchBusyTimesForAnswer(participantEmail, dates) {
+  return async dispatch => {
+    dates.forEach(async date => {
```
mic4ael: Can't we do it in one go (one request)?

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -55,10 +59,21 @@ export default function AnswerPage() {
     setSaved(true);
   }
 
+  useEffect(() => {
```
pferreir: Great, that's gonna conflict with my PR 😆

newdle/client/src/actions.js
```diff
@@ -170,3 +172,19 @@ export function setAnswerActiveDate(date) {
...
+      const times = await client.getBusyTimes(date, participantEmail);
+      dispatch({type: SET_ANSWER_BUSY_TIMES, date, times});
+    });
+  };
+}
+
+export function fetchParticipant(newdleCode, participantCode) {
```
mic4ael: I think @pferreir already added similar function in his PR?

newdle/client/src/components/answer/Calendar.js
```diff
@@ -115,6 +116,26 @@ function calculateOptionsPositions(options, duration, minHour, maxHour, answers)
   });
 }
 
+function getBusySlotProps(slot, minHour, maxHour) {
+  const startTime = slot[0];
```
mic4ael: `const [startTime, endTime] = slot`

newdle/client/src/components/answer/Calendar.js
```diff
@@ -115,6 +116,26 @@ function calculateOptionsPositions(options, duration, minHour, maxHour, answers)
...
+    endTime,
+    height: calculateHeight(start, end, minHour, maxHour),
+    pos: calculatePosition(start, minHour, maxHour),
+    key: `${startTime}-${endTime}`,
+  };
+}
+
+function calculateBusyPositions(busyTimes, minHour, maxHour) {
+  return Object.entries(busyTimes).map(([date, times]) => {
+    return {date: date, times: times.map(slot => getBusySlotProps(slot, minHour, maxHour))};
```
mic4ael: `{date, ...`

newdle/client/src/actions.js
ThiefMaster: i thought about this as well - since this is for a single participant i think fetching it at once would be fine, unless the api has bad performance. OTOH, for a big newdle with many slots, performance with separate requests may be better.

anyway, this is currently using an API not meant for it just to have something to test with. and that particular API is day-based since the creation page only shows one day at a time.

pferreir: I thought about it too but it doesn't seem that heavy to me. We're talking of max. a few days at a time, which may indeed cause problems if the API is slow (as @ThiefMaster suggests), but otherwise pretty normal for a modern web app.

meluru: Yeah, I'm doing it this way for performance reasons

meluru: Okay, It seems like everyone is more into one request, so I'll fix that :)

ThiefMaster: I would wait with that until we add an API to get your own availability. It probably won't be a big change here anyway.

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -55,6 +59,16 @@ export default function AnswerPage() {
...
+    if (newdle && participantCode) {
+      dispatch(fetchParticipant(newdle.code, participantCode));
+    }
+  }, [newdle, participantCode, dispatch]);
+
+  useEffect(() => {
+    dispatch(fetchBusyTimesForAnswer(participantEmail, dates));
```
ThiefMaster: this breaks if you do not have a participant, doesn't it?

newdle/client/src/components/answer/Calendar.js
```diff
@@ -32,8 +33,7 @@ function calculateHeight(start, end, minHour, maxHour, duration) {
     endMins = maxHour * 60;
   }
   const height = ((endMins - startMins) / ((maxHour - minHour) * 60)) * 100;
```
mic4ael: return the expression here directly instead of assigning it to a variable

newdle/client/src/components/answer/AnswerPage.js
meluru: It didn't break for some reason but fixed :)

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -26,6 +28,8 @@ export default function AnswerPage() {
   const availabilityData = useSelector(getAnswers);
   const allAvailableSelected = useSelector(isAllAvailableSelected);
   const allAvailableDisabled = useSelector(isAllAvailableSelectedImplicitly);
+  const dates = useSelector(getCalendarDates);
```
ThiefMaster: This contains duplicates if a day has multiple timeslots. Not related to your PR but maybe a good thing to change since it causes duplicate requests.

newdle/client/src/answerSelectors.js
```diff
@@ -100,5 +98,7 @@ export const getNumberOfAvailableAnswers = createSelector(
   getAnswers,
   answers => Object.values(answers).filter(answer => answer === 'available').length
 );
+export const getParticipantEmail = state =>
+  state.answer.participant && state.answer.participant.email;
```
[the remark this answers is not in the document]
ThiefMaster: OK nevermind on this. it's a bug that we create participants without an email, I just fixed it in master (442ec69)

PR #102, opened 2019-10-15
Add participant name when answering

### LINKED ISSUE #96: Show participant name when opening answer URL
When I open the URL of an answer form I should get my name somewhere, so that it's clear that this is my answer page.

pferreir:
image

Closes #96

### DIFF
--- a/newdle/client/src/answerSelectors.js
+++ b/newdle/client/src/answerSelectors.js
@@ -5,6 +5,7 @@ import {overlaps, serializeDate, toMoment} from './util/date';
 
 export const getNewdle = state => state.answer.newdle;
 const getHandpickedAnswers = state => state.answer.answers;
+export const getParticipant = state => state.answer.participant;
 export const getNewdleDuration = state => state.answer.newdle && state.answer.newdle.duration;
 export const getNewdleTimeslots = state =>
   (state.answer.newdle && state.answer.newdle.timeslots) || [];
--- a/newdle/client/src/answerSelectors.js
+++ b/newdle/client/src/answerSelectors.js
@@ -22,8 +22,10 @@ export const getCalendarDates = createSelector(
 export const getActiveDate = state =>
   state.answer.calendarActiveDate || getCalendarDates(state)[0] || serializeDate(moment());
 
+/** All "busy" times, indexed by day */
 export const getBusyTimes = state => state.answer.busyTimes || {};
 
+/** All "busy" times, flat version */
 const getFlatBusyTimes = createSelector(
   getBusyTimes,
   busyTimes =>
@@ -48,7 +50,11 @@ const getFlatBusyTimes = createSelector(
       })
     )
 );
+
+/** Whether the "accept all option where I'm available" option is checked in the UI */
 export const isAllAvailableSelectedExplicitly = state => state.answer.allAvailable;
+
+/** All time slots during which the person is free */
 const getFreeTimeslots = createSelector(
   getFlatBusyTimes,
   getNewdleTimeslots,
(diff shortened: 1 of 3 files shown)


newdle/client/src/actions.js
```diff
@@ -99,6 +100,13 @@ export function removeParticipant(participant) {
   return {type: REMOVE_PARTICIPANT, participant};
 }
 
+export function fetchParticipant(newdleCode, participantCode, action = PARTICIPANT_RECEIVED) {
```
ThiefMaster: I don't think we need the action as an argument here (so far)

newdle/client/src/answerSelectors.js
```diff
@@ -19,11 +20,17 @@ export const getCalendarDates = createSelector(
...
   state.answer.calendarActiveDate || getCalendarDates(state)[0] || serializeDate(moment());
+/**
+ * All "busy" times, indexed by day
+ */
 export const getBusyTimes = () => ({
   '2019-10-03': [['12:00', '13:00'], ['13:30', '14:30']],
   '2019-10-11': [['09:45', '12:00'], ['16:00', '18:00']],
   '2019-10-12': [['22:00', '00:00']],
 });
+/**
```
ThiefMaster: Is `/** blah blah */` valid for jsdoc-style comments that fit in a single line?

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -10,13 +10,59 @@ import {
...
+            if (e.key === 'Enter') {
+              onSubmit();
+            }
+          }}
+        />
+      </div>
+    );
+  } else {
+    return (
+      participant && (
```
mic4ael: Can it be `null` or `undefined` if it is for sure not `anonymous`?

pferreir: Yes, while it's being loaded.

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -10,13 +10,59 @@ import {
...
+          onChange={(_, data) => setName(data.value)}
+          onKeyDown={e => {
+            if (e.key === 'Enter') {
+              onSubmit();
+            }
+          }}
+        />
+      </div>
+    );
+  } else {
```
mic4ael:
```javascript

} else if (participant) {
  return ...
} else {
  return null;
}
```

newdle/client/src/components/answer/AnswerPage.js
mic4ael: Ah, right.

newdle/client/src/actions.js
mic4ael: Are there any cases in which we will pass a custom `action` to this function? I don't see any now.

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -10,13 +10,59 @@ import {
...
+function ParticipantName({anonymous, setName, onSubmit, disabled}) {
+  const dispatch = useDispatch();
+  const {partcode: participantCode, code: newdleCode} = useParams();
+  const participant = useSelector(getParticipant);
+
+  useEffect(() => {
+    if (!anonymous) {
+      dispatch(fetchParticipant(newdleCode, participantCode));
+      return () => {
+        dispatch(abortAnswering());
```
ThiefMaster: I don't think we need this here again; we already do it in `AnswerHeader`.

However, I think fetching from the header isn't super clear (I just did it there since we do the same on the summary page).
What about having some kind of `AnswerController` component that doesn't render anything but takes care of fetching/aborting? This component would then be mounted through yet another `<Route>` matching th answer URL.

pferreir: Agreed. Not something I'd do in this PR, though.

newdle/client/src/answerSelectors.js
pferreir: Yeah, it's just my editor which adds multi-line by default. I'll fix it.

PR #109, opened 2019-10-16
Send summary e-mails

### LINKED ISSUE #104: Send e-mail from "final" page
We should add a button next to "create event", which would send a simple e-mail to all known participants whose e-mail address we know about. The mail should just announce what the final date is.

meluru:
Closes #104

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -331,4 +331,4 @@ def send_result_emails(code):
             'timezone': newdle.timezone,
         },
     )
-    return jsonify({})
+    return '', 204
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -312,6 +312,8 @@ def create_participant(code):
 @api.route('/newdle/<code>/send-result-emails', methods=('POST',))
 def send_result_emails(code):
     newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    if newdle.creator_uid != g.user['uid']:
+        raise Forbidden
     date = newdle.final_dt.strftime('%-d %B %Y')
     start_time = newdle.final_dt.strftime('%H:%M')
     end_time = (newdle.final_dt + newdle.duration).strftime('%H:%M')
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -311,4 +311,22 @@ def create_participant(code):
 
 @api.route('/newdle/<code>/send-result-emails', methods=('POST',))
 def send_result_emails(code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    date = newdle.final_dt.strftime('%-d %B %Y')
+    start_time = newdle.final_dt.strftime('%H:%M')
+    end_time = (newdle.final_dt + newdle.duration).strftime('%H:%M')
+    notify_newdle_participants(
+        newdle,
+        f'Result: {newdle.title}',
+        'result_email.txt',
+        'result_email.html',
+        lambda p: {
+            'creator': newdle.creator_name,
+            'title': newdle.title,
+            'date': date,
+            'start_time': start_time,
+            'end_time': end_time,
+            'timezone': newdle.timezone,
+        },
+    )
     return jsonify({})
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -307,3 +307,8 @@ def create_participant(code):
         newdle.participants.add(participant)
         db.session.commit()
     return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/send_summary_emails', methods=('POST',))
+def send_summary_emails(code):
+    return jsonify({})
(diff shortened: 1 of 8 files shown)


newdle/api.py
```diff
@@ -237,3 +237,8 @@ def create_anonymous_participant(args, code):
     newdle.participants.add(participant)
     db.session.commit()
     return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/send_summary_emails', methods=('POST',))
```
ThiefMaster: dashes instead of underscores please ;)

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -30,14 +32,39 @@ export default function SummaryPage() {
     setSubmitting(false);
   };
 
+  const sendSummaryEmails = async () => {
```
ThiefMaster: I think you can use pedro's new `client.useBackend` hook here

newdle/api.py
```diff
@@ -237,3 +237,8 @@ def create_anonymous_participant(args, code):
...
+def send_summary_emails(code):
```
ThiefMaster: not sure about the term "summary email".. maybe "result email" instead?

newdle/client/src/components/summary/SummaryPage.js
pferreir: It's not Pedro's, it's newdle's 😉

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -11,6 +11,8 @@ import styles from './summary.module.scss';
 export default function SummaryPage() {
   const [submitting, setSubmitting] = useState(false);
   const [error, setError] = useState('');
+  const [mailSending, setMailSending] = useState(false);
+  const [mailSent, setMailSent] = useState(false);
```
pferreir: Actually, now that I see, we could probably merge these two states into one: `mailState = null | sending | sent`

meluru: I don't use the submitting state now that I went for useBackend :)

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -9,35 +9,46 @@ import client from '../../client';
...
-      return;
+    const updatedNewdle = await _setFinalDate(newdle.code, finalDate);
+    if (updatedNewdle) {
+      dispatch(updateNewdle(updatedNewdle));
     }
-    dispatch(updateNewdle(updatedNewdle));
-    setSubmitting(false);
+  };
+
+  const sendResultEmails = async () => {
```
ThiefMaster: you can just call `_sendResultEmails(newdle.code);` directly' no need to async/await if you don't actually do something with the result (or once the call finished)

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -9,35 +9,46 @@ import client from '../../client';
...
         <>
+          {mailError && (
+            <Message error>
+              <p>Something when wrong when sending result mails:</p>
+              <code>{mailError}</code>
+            </Message>
+          )}
+          {mailSent && (
+            <Message success>
+              <p>Result mails have been sent to participants</p>
```
ThiefMaster: "The participants have been notified of the final date" or something like that?

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -51,18 +62,30 @@ export default function SummaryPage() {
             </div>
           </div>
           <div className={styles['button-row']}>
+            <Button
+              icon
+              color="blue"
+              labelPosition="left"
+              disabled={mailSending || mailSent}
+              onClick={() => {
```
ThiefMaster: `onClick={sendResultEmails}`

newdle/api.py
```diff
@@ -237,3 +237,8 @@ def create_anonymous_participant(args, code):
     newdle.participants.add(participant)
     db.session.commit()
     return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/send-result-emails', methods=('POST',))
+def send_result_emails(code):
+    return jsonify({})
```
ThiefMaster: `return '', 204` ("no content") is probably fine here for now.

At some point it could be nice to return more detailed information, e.g. if sending an email to someone failed, but for the prototype that's too much. Failing the request is a bit ugly since you won't really know that the newdle has actually been created, and most likely some people got emails, but it's a prototype..

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -51,18 +60,28 @@ export default function SummaryPage() {
             </div>
           </div>
           <div className={styles['button-row']}>
+            <Button
+              icon
+              color="blue"
+              labelPosition="left"
+              disabled={mailSending || mailSent}
```
ThiefMaster: What about adding `loading={mailSending}` to have the nice progress animation on the button while the request is running? Especially since sending emails can take a big depending on how many participants you have.

PR #113, opened 2019-10-16
Answer Page UI

### LINKED ISSUE #97: Better answer editing UI
It's already possible to edit an answer by knowing its URL. However, the UI is not very clear.
I suspect, however, that we will need some way of figuring out whether it is the first time the user is answering that particular newdle, so that we can show the button as "Save answer" or "Submit answer", for instance. Rather than adding yet another column to the `Participant` model, we could just check whether `answers` is empty. Opinions?

panagiotappl:
closes #97 

TODO:

- [x] Render existing participant answers in timeline
- [x] Use different button whether user has answered once or not
- [x] Create participant code from logged in user (and not only from anonymous)
- [x] Fetch busy times for logged in participant before they acquire a code
- [x] Disable update button if no change has been made
- [x] Check if the users match, and if not show a message ("answering on behalf of" or something like that)

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -230,7 +230,6 @@ def update_newdle(args, code):
 
 
 @api.route('/newdle/<code>/participants/me')
-@allow_anonymous
 def get_participant_me(code):
     participant = Participant.query.filter(
         Participant.newdle.has(Newdle.code == code),
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -127,9 +127,18 @@ def get_busy_times(date, uid):
 
 
 @api.route('/newdle/<code>/participants/<participant_code>/busy')
+@api.route('/newdle/<code>/participants/me/busy')
 @allow_anonymous
 @use_kwargs({'date': fields.Date(format=DATE_FORMAT, required=True)})
-def get_participant_busy_times(code, participant_code, date):
+def get_participant_busy_times(date, code, participant_code=None):
+    if participant_code is None:
+        # we don't need to check anything in this case, since it's data
+        # for the currently logged-in user
+        if not g.user:
+            return jsonify(error='token_missing'), 401
+        return _get_busy_times(date, g.user['uid'])
+    # if a participant is specified, only allow getting busy times for a valid
+    # timeslots of the newdle to avoid leaking data to anonymous people
     participant = Participant.query.filter(
         Participant.newdle.has(Newdle.code == code),
         Participant.code == participant_code,
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -218,6 +218,19 @@ def update_newdle(args, code):
     return NewdleSchema().jsonify(newdle)
 
 
+@api.route('/newdle/<code>/participants/me')
+@allow_anonymous
+def get_participant_me(code):
+    participant = Participant.query.filter(
+        Participant.newdle.has(Newdle.code == code),
+        Participant.auth_uid == g.user['uid'],
+    ).first()
+    if participant:
+        return ParticipantSchema().jsonify(participant)
+    else:
+        return jsonify(None)
+
+
 @api.route('/newdle/<code>/participants/<participant_code>')
 @allow_anonymous
 def get_participant(code, participant_code):
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -268,3 +268,19 @@ def create_anonymous_participant(args, code):
     newdle.participants.add(participant)
     db.session.commit()
     return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/participants/me', methods=('PUT',))
+def create_participant(code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    name = f'{g.user["first_name"]} {g.user["last_name"]}'
+    participant = Participant.query.filter_by(
+        newdle=newdle, auth_uid=g.user['uid']
+    ).first()
+    if not participant:
+        participant = Participant(
+            name=name, email=g.user['email'], auth_uid=g.user['uid']
+        )
+        newdle.participants.add(participant)
+        db.session.commit()
+    return ParticipantSchema().jsonify(participant)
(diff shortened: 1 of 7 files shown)


newdle/api.py
```diff
@@ -191,6 +191,19 @@ def update_newdle(args, code):
     return NewdleSchema().jsonify(newdle)
 
 
+@api.route('/newdle/<code>/participants/me')
+@allow_anonymous
+def get_participant_me(code):
+    auth_uid = g.user['uid']
+    participant = Participant.query.filter(
+        Participant.newdle.has(Newdle.code == code), Participant.auth_uid == auth_uid
```
ThiefMaster: i'd just use `g.user['uid']` here directly

newdle/api.py
```diff
@@ -237,3 +250,16 @@ def create_anonymous_participant(args, code):
     newdle.participants.add(participant)
     db.session.commit()
     return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/participants/me', methods=('PUT',))
+def create_participant(code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    user = g.user
```
ThiefMaster: just use `g.user` directly

newdle/api.py
```diff
@@ -237,3 +250,16 @@ def create_anonymous_participant(args, code):
...
+    name = user['first_name'] + ' ' + user['last_name']
+    participant = Participant.query.filter_by(auth_uid=user['uid']).first()
+    if not participant:
+        participant = Participant(name=name, email=user['email'], auth_uid=user['uid'])
+    newdle.participants.add(participant)
+    db.session.commit()
```
ThiefMaster: these two lines should be inside the if body as well

newdle/client/src/client.js
```diff
@@ -112,9 +112,13 @@ class Client {
     });
   }
 
-  getParticipant(newdleCode, participantCode) {
-    const params = {code: newdleCode, participant_code: participantCode};
-    return this._request(flask`api.get_participant`(params), {anonymous: true});
+  getParticipant(newdleCode, participantCode, me) {
+    if (me) {
```
ThiefMaster: I'd use `participantCode === null` for that, but nothing wrong with being explicit.

newdle/client/src/client.js
```diff
@@ -112,9 +112,13 @@ class Client {
     });
   }
 
-  getParticipant(newdleCode, participantCode) {
-    const params = {code: newdleCode, participant_code: participantCode};
-    return this._request(flask`api.get_participant`(params), {anonymous: true});
+  getParticipant(newdleCode, participantCode, me) {
```
ThiefMaster: `me = false`?

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -15,13 +15,22 @@ import {
...
+import {getUserInfo} from '../../selectors';
 import {chooseAllAvailable, fetchBusyTimesForAnswer, fetchParticipant} from '../../actions';
 import styles from './answer.module.scss';
 import client from '../../client';
 
 function ParticipantName({anonymous, setName, onSubmit, disabled}) {
   const participant = useSelector(getParticipant);
+  const user = useSelector(getUserInfo);
+  let p = null;
+  //if (participant) {
```
ThiefMaster: :see_no_evil:

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -86,19 +101,25 @@ export default function AnswerPage() {
...
     setSaved(true);
   }
 
   useEffect(() => {
-    if (newdle && participantCode) {
-      dispatch(fetchParticipant(newdle.code, participantCode));
+    if (newdle && (participantCode || user)) {
+      dispatch(fetchParticipant(newdle.code, participantCode, !!user));
     }
   }, [newdle, participantCode, dispatch]);
```
ThiefMaster: pretty sure you need to include `user` here as well

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -86,19 +101,25 @@ export default function AnswerPage() {
...
 
+  useEffect(() => {
+    if (user && !participantCode && participant) {
+      history.push(`/newdle/${newdle.code}/${participant.code}`);
```
ThiefMaster: i'd probably use `.replace()` here. pushing something where using the back button will immediately redirect you again is usually not great because it pretty much breaks the back button

newdle/api.py
```diff
@@ -237,3 +250,16 @@ def create_anonymous_participant(args, code):
     newdle.participants.add(participant)
     db.session.commit()
     return ParticipantSchema().jsonify(participant)
+
+
+@api.route('/newdle/<code>/participants/me', methods=('PUT',))
+def create_participant(code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
+    user = g.user
+    name = user['first_name'] + ' ' + user['last_name']
```
mic4ael: `f'{user["first_name"]} {user["last_name"]}'`

newdle/client/src/client.js
panagiotappl: I specified that inside the `fetchParticipant` action. Should I specify it here too?

ThiefMaster: I think it makes sense to keep defaults in all "public" places (and the client is one for sure)

newdle/api.py
panagiotappl: Aren't they going to be executed anyway if they are directly outside the if body? (There's no return inside if)

ThiefMaster: but you don't need to execute them if you didn't create a new participant..

panagiotappl: Actually, the problem here is that I didn't query participants by newdle as well.

newdle/api.py
```diff
@@ -191,6 +191,19 @@ def update_newdle(args, code):
...
+@allow_anonymous
+def get_participant_me(code):
+    participant = Participant.query.filter(
+        Participant.newdle.has(Newdle.code == code),
+        Participant.auth_uid == g.user['uid'],
+    ).first()
+    if participant:
+        return ParticipantSchema().jsonify(participant)
+    else:
+        return jsonify(None)
```
panagiotappl: @ThiefMaster Returning None in this case breaks the answer reducer (as this is returned from `PARITICPANT_RECIEVED` and it's null)

panagiotappl: I don't think so. I'll add a check  for this.

PR #117, opened 2019-10-17
Block answering when newdle is finished

### LINKED ISSUE #111: Block newdle once it's in "finished" state
It's still possible to answer it.

meluru:
Closes #111 

Answer page:

image

Changed on summary page as well:

image

### DIFF
--- a/newdle/client/src/components/answer/AnswerPage.js
+++ b/newdle/client/src/components/answer/AnswerPage.js
@@ -68,9 +68,8 @@ export default function AnswerPage() {
   const participantUID = useSelector(getParticipantUID);
   const [name, setName] = useState('');
   const history = useHistory();
-  const [saved, setSaved] = useState(false);
 
-  const [submitAnswer, submitting, error] = participantCode
+  const [submitAnswer, submitting, error, submitResult] = participantCode
     ? client.useBackend(client.updateParticipantAnswers)
     : client.useBackend(
         async (...params) => {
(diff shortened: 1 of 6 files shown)


newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -109,78 +110,96 @@ export default function AnswerPage() {
...
-          <Grid.Row centered>
-            <Message success>
-              <p>Your answer has been saved!</p>
-            </Message>
+            <Grid.Column width={12}>
+              <Message
+                info
+                icon="info circle"
+                header="This newdle has already finished"
+                content="It is not possible to answer this newdle anymore. The author has already chosen a final date."
```
ThiefMaster: Isn't "The author has already chosen a final date." clear from what we show below?

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -109,78 +110,96 @@ export default function AnswerPage() {
     return null;
   }
 
+  const {final_dt: finalDate} = newdle;
+
   return (
     <div>
       <Grid container>
-        {error && (
+        {finalDate ? (
```
ThiefMaster: What about moving this outside this `return`? ie `if (finalDate) { return ...; }`?

that avoids this massive indentation change and i think it's also clearer because it's completely separate from the rest here

PR #118, opened 2019-10-17
Error handling

### LINKED ISSUE #103: Handle errors - invalid newdle/participant codes
Currently the client-side app just blows up if one of those codes are invalid. We should rather show a proper error message / 404 page.

mic4ael:
closes #103

### DIFF
--- a/newdle/client/src/actions.js
+++ b/newdle/client/src/actions.js
@@ -34,7 +34,8 @@ export const CHOOSE_ALL_AVAILABLE = 'Choose all slots where user is available';
 export const CHOOSE_MANUALLY = 'Manually select available slots';
 export const SET_ANSWER_BUSY_TIMES = 'Set answer busy times';
 export const PARTICIPANT_RECEIVED = 'Received participant data';
-export const SET_ERROR = 'Error happened';
+export const ADD_ERROR = 'Error occurred';
+export const REMOVE_ERROR = 'Remove error';
 export const CLEAR_ERROR = 'Clear the error';
 
 export function loginWindowOpened(id) {
@@ -65,14 +66,11 @@ export function userLogout() {
 
 export function loadUser() {
   return async dispatch => {
-    try {
-      const user = await client.getMe();
+    const user = await client.getMe();
+    if (user) {
       dispatch({type: USER_RECEIVED, user});
-      return user;
-    } catch (err) {
-      dispatch(setError(err.message));
-      return null;
     }
+    return user;
   };
 }
 
@@ -112,12 +110,8 @@ export function fetchParticipantBusyTimes(participants, date) {
   return async dispatch => {
     participants.forEach(async participant => {
       dispatch({type: SET_PARTICIPANT_BUSY_TIMES, id: participant.uid, date, times: null});
-      try {
-        const times = await client.getBusyTimes(date, participant.uid);
-        dispatch({type: SET_PARTICIPANT_BUSY_TIMES, id: participant.uid, date, times});
-      } catch (err) {
-        dispatch(setError(err.message));
-      }
+      const times = await client.getBusyTimes(date, participant.uid);
+      dispatch({type: SET_PARTICIPANT_BUSY_TIMES, id: participant.uid, date, times});
     });
   };
 }
@@ -148,7 +142,6 @@ export function fetchNewdle(code, fullDetails = false, action = NEWDLE_RECEIVED)
       const newdle = await client.getNewdle(code, fullDetails);
       dispatch({type: action, newdle});
     } catch (err) {
-      dispatch(setError(err.message));
       dispatch({type: action, newdle: null});
     }
   };
@@ -192,31 +185,27 @@ export function updateNewdle(newdle) {
 export function fetchBusyTimesForAnswer(newdleCode, participantCode, dates) {
   return async dispatch => {
     dates.forEach(async date => {
-      try {
-        const times = await client.getBusyTimes(date, null, newdleCode, participantCode);
-        dispatch({type: SET_ANSWER_BUSY_TIMES, date, times});
-      } catch (err) {
-        dispatch(setError(err.message));
-      }
+      const times = await client.getBusyTimes(date, null, newdleCode, participantCode);
+      dispatch({type: SET_ANSWER_BUSY_TIMES, date, times});
     });
   };
 }
 
 export function fetchParticipant(newdleCode, participantCode) {
   return async dispatch => {
-    try {
-      const participant = await client.getParticipant(newdleCode, participantCode);
-      if (participant) {
-        dispatch({type: PARTICIPANT_RECEIVED, participant});
-      }
-    } catch (err) {
-      dispatch(setError(err.message));
+    const participant = await client.getParticipant(newdleCode, participantCode);
+    if (participant) {
+      dispatch({type: PARTICIPANT_RECEIVED, participant});
     }
   };
 }
 
-export function setError(error) {
-  return {type: SET_ERROR, error};
+export function addError(error) {
+  return {type: ADD_ERROR, error};
+}
+
+export function removeError(errorId) {
+  return {type: REMOVE_ERROR, id: errorId};
 }
 
 export function clearError() {
--- a/newdle/client/src/actions.js
+++ b/newdle/client/src/actions.js
@@ -36,7 +36,7 @@ export const SET_ANSWER_BUSY_TIMES = 'Set answer busy times';
 export const PARTICIPANT_RECEIVED = 'Received participant data';
 export const ADD_ERROR = 'Error occurred';
 export const REMOVE_ERROR = 'Remove error';
-export const CLEAR_ERROR = 'Clear the error';
+export const CLEAR_ERRORS = 'Clears all the errors';
 
 export function loginWindowOpened(id) {
   return {type: LOGIN_WINDOW_OPENED, id};
@@ -66,8 +66,8 @@ export function userLogout() {
 
 export function loadUser() {
   return async dispatch => {
-    const user = await client.getMe();
-    if (user) {
+    const user = await client.catchErrors(client.getMe());
+    if (user !== undefined) {
       dispatch({type: USER_RECEIVED, user});
     }
     return user;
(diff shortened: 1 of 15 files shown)


pferreir:
Not sure if you noticed, but the tests are broken.

newdle/client/src/components/App.js
```diff
@@ -3,33 +3,42 @@ import {useSelector} from 'react-redux';
...
-          <Route exact path="/new" component={CreationPage} />
-          <Route exact path="/new/success" component={CreationSuccessPage} />
-          <Route exact path="/mine" component={MyNewdles} />
-          <Route path="/newdle/:code/summary" component={SummaryPage} />
-          <Route exact path="/newdle/:code/:partcode?" component={AnswerPage} />
-          <Route render={() => 'This page does not exist'} />
-        </Switch>
-        <LoginPrompt />
-        {loggingIn && <LoggingIn />}
+        {!error ? (
```
pferreir: Any reason in particular to use the negative formulation in here? I would do `error ? ...` to reduce confusion.

newdle/client/src/components/ErrorMessage.js
```diff
@@ -0,0 +1,31 @@
...
+
+  const unregister = history.listen(() => {
+    dispatch(clearError());
+  });
+
+  useEffect(() => {
+    return () => {
+      unregister();
+    };
+  }, [unregister]);
```
pferreir: I'm not sure I understand what this is supposed to do. Maybe we could add some comments?

newdle/client/src/components/App.js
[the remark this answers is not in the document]
ThiefMaster: even worse for a failed search query.. this shouldn't break everything and make you lose anything you entered

mic4ael:
> Not sure if you noticed, but the tests are broken.

I didn't since I left after creating this PR ;)

newdle/client/src/components/ErrorMessage.js
```diff
@@ -0,0 +1,31 @@
...
+import PropTypes from 'prop-types';
+import {useHistory} from 'react-router';
+import {Container, Message} from 'semantic-ui-react';
+import {clearError} from '../actions';
+
+export default function ErrorMessage({error}) {
+  const history = useHistory();
+  const dispatch = useDispatch();
+
+  const unregister = history.listen(() => {
```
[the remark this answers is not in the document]
mic4ael: Will check tomorrow ;)

newdle/client/src/components/ErrorMessage.js
mic4ael: This is responsible for removing the callback registered with `listen` once the component gets unmounted.

newdle/client/src/components/App.js
mic4ael: Hm, no strong opinion here. Honestly I don't find it that confusing but true that it is not necessary to negate it here ;)

newdle/client/src/components/creation/userSearch/UserSearch.js
```diff
@@ -19,8 +19,12 @@ async function searchUsers(data, setResults) {
   if (!q.trim()) {
     return;
   }
-  const results = await client.searchUsers(q);
-  setResults(results);
+  try {
+    const results = await client.searchUsers(q);
+    setResults(results);
+  } catch (err) {
+    handleError(err);
```
pferreir: Wouldn't it make sense to include the whole error handling thing directly into the `client` instead?

ThiefMaster: Yeah, and return more structured error data from the server, indicating whether it's something expected or not - check what I'm doing in `require_token` for example.

So whenever you have an error code you can determine how to handle it; and if there's no error code I'd say it's fine to just show the generic error message similar to how you already do.

newdle/client/src/client.js
```diff
@@ -1,14 +1,12 @@
...
 import {useReducer} from 'react';
 import {getToken, isAcquiringToken} from './selectors';
-import {tokenExpired, tokenNeeded} from './actions';
+import {tokenExpired, tokenNeeded, addError} from './actions';
 
 function backendReducer(state, action) {
   switch (action.type) {
     case 'submit':
       return {...state, submitting: true, error: '', result: null};
-    case 'error':
```
ThiefMaster: I think we should keep this; a component may want to know that a call it did failed

newdle/client/src/client.js
```diff
@@ -181,7 +179,7 @@ class Client {
     try {
       resp = await fetch(url, requestOptions);
     } catch (err) {
-      throw new ClientError(url, 0, err);
+      this.store.dispatch(addError(err.message));
```
ThiefMaster: Now you aren't aborting execution of the code below anymore...

ThiefMaster: also, during development we now have much less information on what went wrong. I think we should at least `console.error()` the actual ClientError...

newdle/client/src/components/MyNewdles.js
```diff
@@ -7,25 +8,21 @@ import client from '../client';
...
   useEffect(() => {
     let aborted = false;
     (async () => {
-      let newdles;
-      try {
-        newdles = await client.getMyNewdles();
-      } catch (exc) {
-        return;
-      }
+      const newdles = await client.getMyNewdles();
```
ThiefMaster: This code should probably be updated to use the client hook

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -85,7 +85,7 @@ export default function AnswerPage() {
   const participantAnswersChanged = useSelector(haveParticipantAnswersChanged);
   const busyTimesLoaded = useSelector(hasBusyTimes);
 
-  const [submitAnswer, submitting, error, submitResult] = participantCode
+  const [submitAnswer, submitting, , submitResult] = participantCode
```
ThiefMaster: yuck, that's actually valid?

mic4ael: Yup.

newdle/client/src/components/ErrorMessage.js
```diff
@@ -0,0 +1,42 @@
...
+import {clearError, removeError} from '../actions';
+
+import styles from './ErrorMessage.module.scss';
+
+export default function ErrorMessage({id, error}) {
+  const history = useHistory();
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    // Remove the callback defined with `listen` once the component gtes unmounted
```
ThiefMaster: *gets

newdle/client/src/components/MyNewdles.js
```diff
@@ -1,31 +1,12 @@
...
-      if (!aborted) {
-        setNewdles(newdles);
-      }
-    })();
-
-    return () => {
-      aborted = true;
-    };
-  }, []);
+  const [newdles] = client.useBackend(() => client.getMyNewdles(), []);
```
ThiefMaster: @pferreir maybe you have a good idea if this can be made nicer without ending up with react hook dependency hell / having to use useCallback here ;)

newdle/client/src/reducers/index.js
```diff
@@ -25,4 +32,18 @@ export default combineReducers({
...
   },
+  error: (state = [], action) => {
+    switch (action.type) {
+      case ADD_ERROR: {
+        const id = state.length !== 0 ? state[state.length - 1].id + 1 : 0;
+        return [...state, {id, error: action.error}];
+      }
+      case REMOVE_ERROR:
+        return state.filter(error => error.id !== action.id);
+      case CLEAR_ERROR:
```
ThiefMaster: `CLEAR_ERRORS`?

newdle/client/src/client.js
```diff
@@ -65,17 +65,53 @@ class Client {
...
       return await f(...params);
     };
-
     return [call, state.submitting, state.error, state.result];
   }
+  /*
+   * It is similar to `useBackend` however the function passed as func is
+   * called immediatelly. This can be useful for example when you need to fetch
+   * some data while the component is being mounted.
+   */
```
pferreir: Is the JSDoc OK?

pferreir: I meant the content. But actually, I believe valid JSDoc is something like:
```js
/**
 * ...
 */
```

newdle/client/src/components/MyNewdles.js
pferreir: Instead of passing a closure to `useBackend(...)` you could just pass `client.getMyNewdles` and bind `func` to `this` inside `useBackend`. Unless you think we will ever need to pass a function that shouldn't be bound to `client`. But in that case you could always use a closure like you did here.

ThiefMaster: the problem with that is when you want to pass arguments to the function call.. it'd have to be a separate param of the hook, but then you'd also need to have this in the dependency of the useEffect inside the hook...

newdle/client/src/actions.js
```diff
@@ -34,6 +34,9 @@ export const CHOOSE_ALL_AVAILABLE = 'Choose all slots where user is available';
 export const CHOOSE_MANUALLY = 'Manually select available slots';
 export const SET_ANSWER_BUSY_TIMES = 'Set answer busy times';
 export const PARTICIPANT_RECEIVED = 'Received participant data';
+export const ADD_ERROR = 'Error occurred';
+export const REMOVE_ERROR = 'Remove error';
+export const CLEAR_ERRORS = 'Clears all the errors';
```
pferreir: Nitpicking, but this should be in the imperative mood ("Clear all the errors"), right?

newdle/client/src/components/MyNewdles.js
pferreir: But you can always just pass a closure if you need to pass arguments too, right?

newdle/client/src/actions.js
mic4ael: Right, thanks!

newdle/client/src/client.js
```diff
@@ -65,18 +65,57 @@ class Client {
...
       }, end);
 
       return await f(...params);
     };
-
     return [call, state.submitting, state.error, state.result];
   }
 
+  /**
+   * `useBackendLazy` works fine for methods that are invoked manually (e.g. clicking a button) but may cause problems
```
ThiefMaster: this docstring is on the wrong function..

mic4ael: I think it is on the right function :P but maybe it was confusing that I started the docstring with `useBackendLazy`

PR #123, opened 2019-10-18
Create reusable NewdleTitle component

### LINKED ISSUE #122: Create generic reusable Newdle Title
Right now the newdle title appears differently on different pages (eg. AnswerPage and Summary Page)

panagiotappl:
Same component can be now used for Summary and Answer pages
closes #122 
Examples: 
indicomeeting
indicometting2


newdle/client/src/components/summary/SummaryHeader.js
```diff
@@ -16,19 +15,10 @@ export default function SummaryHeader({match}) {
...
-            <Label color={finalDt ? 'blue' : 'green'} size="tiny" className={styles.label}>
-              {finalDt ? 'finished' : 'ongoing'}
-            </Label>
-          </div>
-          <div className={styles.subtitle}>by {user.name}</div>
-        </>
-      )}
-    </Container>
-  );
+  const label = finalDt ? 'finished' : 'ongoing';
```
ThiefMaster: i'd move that right before the return or even inside the `label={...}`

newdle/client/src/components/NewdleTitle.js
```diff
@@ -0,0 +1,34 @@
...
+import styles from './NewdleTitle.module.scss';
+
+export default function NewdleTitle({title, author, label}) {
+  return (
+    <Container text className={styles.box}>
+      <div className={styles.title}>
+        <h1 className={styles.header}>{title}</h1>
+        {label && (
+          <Label
+            color={label === 'finished' ? 'blue' : 'green'}
```
ThiefMaster: please don't check human-targeted text for something like this - it'll break in the future with i18n. better to use props to specify the state of the newdle, and then based on this show the label (if it's not null) and determine its color.

if you want to keep it for now, change the prop type from string to `.oneOf(['finished', 'ongoing'])` so it's absolutely clear that this is not just a text that will be displayed

PR #124, opened 2019-10-18
Add user data signatures

### LINKED ISSUE #115: Sign UID + name and check signature before creating a participant.

pferreir:
Also took the occasion to rename "anonymous participant" -> "unknown participant", since those participants do have name anyway.

Closes #115

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -17,9 +17,9 @@ from .models import Newdle, Participant
 from .notifications import notify_newdle_participants
 from .schemas import (
     MyNewdleSchema,
-    NewAnonymousParticipantSchema,
     NewdleSchema,
     NewNewdleSchema,
+    NewUnknownParticipantSchema,
     ParticipantSchema,
     RestrictedNewdleSchema,
     UpdateNewdleSchema,
@@ -150,7 +150,7 @@ def get_participant_busy_times(date, code, participant_code=None):
         Participant.code == participant_code,
     ).first_or_404('Invalid code')
     if participant.auth_uid is None:
-        abort(422, messages={'participant_code': ['Participant is anonymous']})
+        abort(422, messages={'participant_code': ['Participant is an unknown user']})
     if not any(date == ts.date() for ts in participant.newdle.timeslots):
         abort(422, messages={'date': ['Date has no timeslots']})
     return _get_busy_times(date, participant.auth_uid)
@@ -288,8 +288,8 @@ def update_participant(args, code, participant_code):
 
 @api.route('/newdle/<code>/participants', methods=('POST',))
 @allow_anonymous
-@use_args(NewAnonymousParticipantSchema(), locations=('json',))
-def create_anonymous_participant(args, code):
+@use_args(NewUnknownParticipantSchema(), locations=('json',))
+def create_unknown_participant(args, code):
     newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
     if newdle.final_dt:
         raise Forbidden('This newdle has finished')
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -11,13 +11,13 @@ from werkzeug.exceptions import Forbidden, UnprocessableEntity
 from .cern_integration import search_cern_users
 from .core.auth import user_info_from_app_token
 from .core.db import db
-from .core.util import DATE_FORMAT, format_dt, range_union
+from .core.util import DATE_FORMAT, format_dt, range_union, sign_user
 from .core.webargs import abort, use_args, use_kwargs
 from .models import Newdle, Participant
 from .notifications import notify_newdle_participants
 from .schemas import (
-    CreateAnonymousParticipantSchema,
     MyNewdleSchema,
+    NewAnonymousParticipantSchema,
     NewdleSchema,
     NewNewdleSchema,
     ParticipantSchema,
@@ -112,7 +112,13 @@ def users(q):
         total, data = len(res), res[:10]
     else:
         total, data = search_cern_users(q, 10)
-    return {'total': total, 'users': UserSearchResultSchema(many=True).dump(data)}
+    return {
+        'total': total,
+        'users': [
+            sign_user(u, fields={'email', 'name', 'uid'})
+            for u in UserSearchResultSchema(many=True).dump(data)
+        ],
+    }
 
 
 @api.route('/users/busy')
@@ -282,7 +288,7 @@ def update_participant(args, code, participant_code):
 
 @api.route('/newdle/<code>/participants', methods=('POST',))
 @allow_anonymous
-@use_args(CreateAnonymousParticipantSchema(), locations=('json',))
+@use_args(NewAnonymousParticipantSchema(), locations=('json',))
 def create_anonymous_participant(args, code):
     newdle = Newdle.query.filter_by(code=code).first_or_404('Invalid code')
     if newdle.final_dt:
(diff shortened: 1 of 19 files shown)


newdle/api.py
```diff
@@ -181,14 +190,20 @@ def get_my_newdles():
...
+
     newdle = Newdle(
         title=title,
         creator_uid=g.user['uid'],
         creator_name=f'{g.user["first_name"]} {g.user["last_name"]}',
         duration=duration,
         timezone=timezone,
         timeslots=timeslots,
-        participants={Participant(**p) for p in participants},
+        participants=set(p[1] for p in participants),
```
ThiefMaster: `{p[1] for p in participants}` should work as well

newdle/core/util.py
```diff
@@ -33,3 +36,24 @@ def range_union(ranges):
...
+
+
+def sign_user(user):
+    """Sign user data (UID + e-mail)."""
+    signer = Signer(current_app.config['SECRET_KEY'], salt='newdle-users')
+    email = user['email']
+    uid = user['uid']
+    return dict(
+        user,
+        signature=signer.get_signature(f'{uid}-{email}'.encode('ascii')).decode(
```
ThiefMaster: shouldn't we include the name as well? (maybe by using `frozenset(user.items())` as input)

tests/api_test.py
```diff
@@ -114,6 +130,55 @@ def test_create_newdle_duplicate_timeslot(flask_client, dummy_uid):
     }
 
 
+@pytest.mark.usefixtures('db_session')
+def test_create_newdle_participant_email_sending(flask_client, dummy_uid, mocker):
+    notif_func = mocker.patch('newdle.api.notify_newdle_participants')
```
ThiefMaster: you can also use the `locmem` email backend; that way all emails are written to a list instead of being actually sent

pferreir: Oh cool!

newdle/api.py
pferreir: Indeed.

newdle/core/util.py
pferreir: `frozenset(...)`? Why? We need to ensure consistent sorting between attributes. I was thinking more of something like:

```python
'-'.join(v for k, v in sorted(user.items()))
```

tests/api_test.py
pferreir: OK, I still don't know how to get the backend, since it seems like it's not persisted anywhere. Anyway, I am already using it to avoid methods that send e-mails sending. In this method, however, I'd like to count how many mails are sent.

ThiefMaster: `from newdle.vendor import django_mail`

`django_mail.outbox`

newdle/api.py
```diff
@@ -181,14 +190,22 @@ def get_my_newdles():
 @api.route('/newdle/', methods=('POST',))
 @use_kwargs(NewNewdleSchema(), locations=('json',))
 def create_newdle(title, duration, timezone, timeslots, participants):
+    participants = [(p.pop('signature'), Participant(**p)) for p in participants]
+
+    for signature, p in participants:
```
ThiefMaster: no idea if it looks better, but `if not all(check_user_signature(...) for sig, p in participants)`

newdle/core/util.py
```diff
@@ -33,3 +36,20 @@ def range_union(ranges):
...
+    return dict(
+        user_data, signature=signer.get_signature(text.encode('utf-8')).decode('ascii')
+    )
+
+
+def check_user_signature(user_data, signature):
+    """Check that user data (UID + e-mail) matches the signature."""
+    signer = Signer(current_app.config['SECRET_KEY'], salt='newdle-users')
+    text = '-'.join(v for k, v in sorted(user_data.items()))
+    sig = signer.verify_signature(text.encode('utf-8'), signature.encode('ascii'))
```
ThiefMaster: `return ...`

ThiefMaster: AFAIK utf-8 is the default encoding in py3 so it can be omitted in `.encode()`

newdle/schemas.py
```diff
@@ -26,13 +33,20 @@ def sort_users(self, data, many, **kwargs):
...
     name = fields.String(required=True)
-    email = fields.String()
-    auth_uid = fields.String()
 
 
-class ParticipantSchema(NewParticipantSchema):
+class NewKnownParticipantSchema(NewUnknownParticipantSchema):
+    email = fields.String(required=True)
+    auth_uid = fields.String(required=True)
+    signature = fields.String(required=True)
```
ThiefMaster: I think the signature check could be done in a [`validates_schema`](https://marshmallow.readthedocs.io/en/stable/api_reference.html#marshmallow.decorators.validates_schema)-decorated function here

newdle/core/util.py
ThiefMaster: I was hoping the signer would automatically take care of converting non-string data to a suitable format ;) but yeah, if it has to be a string then sorting manually is needed of course..

newdle/api.py
pferreir: It ends up being harder to read, TBH...

newdle/schemas.py
pferreir: Great idea, much cleaner!

tests/api_test.py
pferreir: Thanks!

newdle/core/util.py
```diff
@@ -33,3 +36,28 @@ def range_union(ranges):
...
+
+
+def _get_signature_source_bytes(data, fields=None):
+    if fields:
+        data = {k: v for k, v in data.items() if k in fields}
+    return '-'.join(v for k, v in sorted(data.items())).encode()
+
+
+def sign_user(user_data, fields=None):
+    """Sign user data (UID + e-mail)."""
```
ThiefMaster: i think the `(...)` part of this docstring is outdated

newdle/core/util.py
```diff
@@ -33,3 +36,28 @@ def range_union(ranges):
...
+    return dict(
+        user_data,
+        signature=signer.get_signature(
+            _get_signature_source_bytes(user_data, fields)
+        ).decode('ascii'),
+    )
+
+
+def check_user_signature(user_data, signature, fields=None):
+    """Check that user data (UID + e-mail) matches the signature."""
```
ThiefMaster: same here

newdle/schemas.py
```diff
@@ -26,13 +34,34 @@ def sort_users(self, data, many, **kwargs):
...
+        data = dict(data)
+        signature = data.pop('signature')
+        auth_uid = data.pop('auth_uid')
+        if not check_user_signature(dict(data, uid=auth_uid), signature):
+            raise ValidationError("Participant's user signature is invalid!")
+
+    @post_load
+    def remove_signature(self, data, **kwargs):
+        """Remove signature, which is not needed after validation."""
+        data.pop('signature')
```
ThiefMaster: `del data['signature']`

tests/api_test.py
```diff
@@ -114,6 +124,55 @@ def test_create_newdle_duplicate_timeslot(flask_client, dummy_uid):
     }
 
 
+@pytest.mark.usefixtures('db_session')
+def test_create_newdle_participant_email_sending(flask_client, dummy_uid):
+    django_mail.outbox = []
```
ThiefMaster: this could be in a `autouse=True` fixture ;)

pferreir: Even better, a named fixture you can use wherever you wish and which returns you the actual inbox list:
```python
@pytest.fixture
def mail_queue():
    from newdle.vendor import django_mail
    django_mail.outbox = []
    return django_mail.outbox
```

PR #125, opened 2019-10-21
Add new icons to homepage

meluru:
obraz


newdle/client/src/components/home/HomePage.js
```diff
@@ -20,15 +20,15 @@ export default function HomePage() {
...
             Choose your participants
           </li>
           <li>
-            <TimetableIcon className={styles.icon} />
+            <img src={calendarIcon} alt={''} className={styles.icon} />
             Set the time slots based on their availability
           </li>
           <li>
-            <DiscussIcon className={styles.icon} />
+            <img src={answersIcon} alt={''} className={styles.icon} />
```
pferreir: is the `alt={''}` really needed, and in that case shouldn't we provide a more meaningful description?
BTW, I think `alt=""` would be equivalent?

newdle/client/src/components/home/HomePage.js
```diff
@@ -20,15 +20,15 @@ export default function HomePage() {
         <h2>How does it work?</h2>
         <ol className={styles.box}>
           <li>
-            <TeamIcon className={styles.icon} />
+            <img src={participantsIcon} alt={''} className={styles.icon} />
```
ThiefMaster: `alt=""`

newdle/client/src/components/home/HomePage.js
```diff
@@ -61,26 +61,6 @@ export default function HomePage() {
             </a>
             , the place where the web was born.
           </p>
-          <p className={styles.credits}>
```
ThiefMaster: I guess the CSS rule can be removed as well

newdle/client/src/components/home/HomePage.js
ThiefMaster: I think `alt` makes no sense for those icons since they provide no information that's not part of the text. It's a bit like putting a non-empty alt on an avatar next to the username: it'd just add noise for people who rely on alt tags

pferreir: Yeah, let's just keep `alt=""` then.

PR #140, opened 2019-10-23
Add copy to clipboard btn in creation success page

### LINKED ISSUE #132: Add "copy to clipboard" button in newdle creation success page
image

skbro:
Closes #132 
Copy to Clipboard Button added using clipboard API.
Screenshot from 2019-10-24 00-01-40

### DIFF
--- a/newdle/client/src/components/creation/CreationSuccessPage.js
+++ b/newdle/client/src/components/creation/CreationSuccessPage.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import {Button, Container, Header} from 'semantic-ui-react';
+import {Button, Container, Header, Input, Popup} from 'semantic-ui-react';
 import {useSelector} from 'react-redux';
 import {useHistory} from 'react-router';
 import {Redirect} from 'react-router-dom';
@@ -38,7 +38,28 @@ export default function CreationSuccessPage() {
             invite:
           </p>
         )}
-        <div className={styles['newdle-link']}>{newdle.url}</div>
+        <Input
+          className={styles['newdle-link']}
+          readOnly
+          value={newdle.url}
+          action={
+            navigator.clipboard && (
+              <Popup
+                content="Copied!"
+                on="click"
+                position="top center"
+                inverted
+                trigger={
+                  <Button
+                    icon="copy"
+                    title="Copy to clipboard"
+                    onClick={() => navigator.clipboard.writeText(newdle.url)}
+                  />
+                }
+              />
+            )
+          }
+        />
       </div>
       <div className={styles['summary-button']}>
         <Button color="teal" onClick={handleSummaryClick}>
--- a/newdle/client/src/components/creation/CreationSuccessPage.js
+++ b/newdle/client/src/components/creation/CreationSuccessPage.js
@@ -43,6 +43,9 @@ export default function CreationSuccessPage() {
           fluid
           readOnly
           value={newdle.url}
+          onFocus={evt => {
+            evt.target.select();
+          }}
           action={
             navigator.clipboard && (
               <Popup
--- a/newdle/client/src/components/creation/CreationSuccessPage.js
+++ b/newdle/client/src/components/creation/CreationSuccessPage.js
@@ -40,6 +40,7 @@ export default function CreationSuccessPage() {
         )}
         <Input
           className={styles['newdle-link']}
+          fluid
           readOnly
           value={newdle.url}
           action={
--- a/newdle/client/src/components/creation/CreationSuccessPage.module.scss
+++ b/newdle/client/src/components/creation/CreationSuccessPage.module.scss
@@ -10,8 +10,6 @@
 .newdle-link {
   display: flex;
   justify-content: center;
-  width: 100%;
-  font-size: 1.5em !important;
 
   input {
     text-align: center !important;
--- a/newdle/client/src/components/creation/CreationSuccessPage.module.scss
+++ b/newdle/client/src/components/creation/CreationSuccessPage.module.scss
@@ -10,9 +10,12 @@
 .newdle-link {
   display: flex;
   justify-content: center;
-  padding: 0.67857143em 1em;
-  border: 1px solid rgba(34, 36, 38, 0.15);
-  border-radius: 0.28571429rem;
+  width: 100%;
+  font-size: 1.5em !important;
+
+  input {
+    text-align: center !important;
+  }
 }
 
 .success-message {


ThiefMaster:
Thanks! I think it would be nicer if the button was either attached to the field or inside it.

What about using a readonly Input with an attached button? Check https://react.semantic-ui.com/elements/input/ for some examples.

Just please don't go for the huge copy button in one of their examples (without the text it might be nice though).

skbro:
Okay, sure.

skbro:
Closes Issue #132 
Changed Grid Component to Disabled Input 
Screenshot from 2019-10-24 00-49-21

newdle/client/src/components/creation/CreationSuccessPage.js
```diff
@@ -38,7 +38,17 @@ export default function CreationSuccessPage() {
             invite:
           </p>
         )}
-        <div className={styles['newdle-link']}>{newdle.url}</div>
+        <Input
+          className={styles['newdle-link']}
+          disabled
```
ThiefMaster: I think `readOnly` would be more appropriate here, but I'll have to try it out and see it in action. Could even select the contents of the field when focusing it..

newdle/client/src/components/creation/CreationSuccessPage.js
```diff
@@ -38,7 +38,17 @@ export default function CreationSuccessPage() {
...
+          action={{
+            icon: 'copy',
+            onClick: async () => {
+              await navigator.clipboard.writeText(newdle.url);
```
ThiefMaster: You don't need to `await` here (and thus no `async` needed in the line above) since there's nothing you do after copying the link anyway.

Could be nice to show some kind of "copied" message though - but not sure if it's really needed (feel free to give it a try though), or use a CSS animation to style the button a bit differently to make it clear that something happened when you clicked it.

newdle/client/src/components/creation/CreationSuccessPage.js
skbro: Yeah, the disabled field is having some problems to select the contents. Let me try with readOnly and I will push accordingly.

skbro:
Added a little CSS animation to show something happened when the copy button was clicked. I tried adding the popup component to show a nice copied message, but I think state is required for that.

newdle/client/src/components/creation/CreationSuccessPage.js
```diff
@@ -38,7 +38,19 @@ export default function CreationSuccessPage() {
...
+          readOnly={true}
+          value={newdle.url}
+          action={{icon: 'copy'}}
+        >
+          <input />
+          <Button
+            title="Copy to clipboard"
+            onClick={() => navigator.clipboard.writeText(newdle.url)}
+            icon="copy"
+          ></Button>
```
ThiefMaster: `<Button ... />`

newdle/client/src/components/creation/CreationSuccessPage.js
```diff
@@ -38,7 +38,19 @@ export default function CreationSuccessPage() {
...
           </p>
         )}
-        <div className={styles['newdle-link']}>{newdle.url}</div>
+        <Input
+          className={styles['newdle-link']}
+          readOnly={true}
+          value={newdle.url}
+          action={{icon: 'copy'}}
+        >
+          <input />
```
ThiefMaster: that looks wrong, do we need that?!

newdle/client/src/components/creation/CreationSuccessPage.js
```diff
@@ -38,7 +38,19 @@ export default function CreationSuccessPage() {
             invite:
           </p>
         )}
-        <div className={styles['newdle-link']}>{newdle.url}</div>
+        <Input
+          className={styles['newdle-link']}
+          readOnly={true}
```
ThiefMaster: just `readOnly`

newdle/client/src/components/creation/CreationSuccessPage.js
```diff
@@ -38,7 +38,19 @@ export default function CreationSuccessPage() {
...
+          value={newdle.url}
+          action={{icon: 'copy'}}
+        >
+          <input />
+          <Button
+            title="Copy to clipboard"
+            onClick={() => navigator.clipboard.writeText(newdle.url)}
```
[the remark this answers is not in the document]
ThiefMaster: awaiting is only needed when you actually do something after it finished

newdle/client/src/components/creation/CreationSuccessPage.module.scss
```diff
@@ -10,9 +10,17 @@
...
-  border: 1px solid rgba(34, 36, 38, 0.15);
-  border-radius: 0.28571429rem;
+  width: 100%;
+  font-size: 1.5em !important;
+
+  input {
+    color: black !important;
+  }
+
+  button:active {
```
ThiefMaster: not a big fan of this "animation" style... maybe using state to show a Popup is nicer after all? maybe we can actually go back to only having the copy button in this PR and then, if you want, you send another one that adds the popup?

skbro: Let me try adding popup again and remove the animation.

newdle/client/src/components/creation/CreationSuccessPage.js
skbro: Sure

newdle/client/src/components/creation/CreationSuccessPage.js
skbro: Yeah, I think that is not required. Going to remove it

skbro:
Screenshot from 2019-10-25 09-59-32
Added Popup, removed all custom animations and other reviewed problems.

newdle/client/src/components/creation/CreationSuccessPage.js
```diff
@@ -38,7 +38,26 @@ export default function CreationSuccessPage() {
...
+        <Input
+          className={styles['newdle-link']}
+          readOnly
+          value={newdle.url}
+          action={
+            <Popup
+              content="Copied!"
+              on="click"
+              position="top center"
+              inverted
```
pferreir: I understand that it doesn't make much difference because the copying itself is an instantaneous operation. However, the "right" way to do this would be to only display the `<Popup />` once `writeText(...)` succeeds. I'm not sure under which conditions `writeText(...)` might fail (the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) talk of not having permission to access the clipboard), but maybe it would be worth making the `<Popup />` controlled and adding an additional state to take care of that.
@ThiefMaster opinions?

newdle/client/src/components/creation/CreationSuccessPage.js
```diff
@@ -38,7 +38,26 @@ export default function CreationSuccessPage() {
...
+              trigger={
+                <Button
```
pferreir: The asynchronous Clipboard API is not supported on Edge, so I would probably hide this button in that case. We can just check the truthiness of `navigator.clipboard`, I believe. Anyway, it would be nice if we test it with Edge before we merge.

newdle/client/src/components/creation/CreationSuccessPage.js
ThiefMaster: Agreed, for showing the popup I think using state that's only set after copying would be nicer.. (and then maybe using setTimeout to hide it after a second or two).

We should probably also try..catch around the call to avoid an error when the promise does get rejected. Probably no need to actually do anything in there for now, since it seems unlikely that it happens..

newdle/client/src/components/creation/CreationSuccessPage.js
ThiefMaster: That's weird, generally localhost should be exempt from "https-only" restrictions. If not I'd consider it a bug in the browser...

skbro: [Found on stackoverflow](https://stackoverflow.com/a/51823007)

ThiefMaster: oops, I guess I misread your last comment as "doesn't work on localhost".

ThiefMaster: `action={navigator.clipboard && <...>}` should be fine, we did the same in the indico PR linked in a commend of the original issue

newdle/client/src/components/creation/CreationSuccessPage.js
ThiefMaster: I tested in firefox with `dom.events.asyncClipboard` disabled and the button gets hidden as expected. I think for now it's fine, if we find out that there are cases where `navigator.clipboard` is defined but copying still fails we can still add error handling there

ThiefMaster:
Thanks a lot, it looks good now. I added some small improvements and will now merge it.

Just a tip for the future: Create a branch for your PRs instaed of working directly on your master branch. That way you keep master clean.

skbro:
Yeah, I will make sure of that.
Thank you for the awesome learning experience.

pferreir:
Thanks a lot for your help, @skbro ! It's really appreciated!

PR #144, opened 2019-10-24
Fix elements responsiveness in home page

nop33:

### DIFF
--- a/newdle/client/src/components/TopHeader.js
+++ b/newdle/client/src/components/TopHeader.js
@@ -12,7 +12,7 @@ import styles from './TopHeader.module.scss';
 export default function TopHeader() {
   return (
     <header className={styles.header}>
-      <Grid columns={3}>
+      <Grid columns={2}>
         <Grid.Row verticalAlign="middle">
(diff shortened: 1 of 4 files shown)


nop33:
I am not 100% happy with the solution regarding the logo. For a better solution, I need to understand better the reason the middle (empty) column of the header exists:

image

ThiefMaster:
It looks like it's only there for alignment: https://github.com/indico/newdle/blob/master/newdle/client/src/components/TopHeader.js#L24

maybe @panagiotappl remembers

nop33:
If that's the case, I'd rather remove the middle column to give more space for the logo.

nop33:
I included a couple more small and quick fixes. Here's a preview:

image

ThiefMaster:
I just noticed some small issue with inconsistent icon sizes on phone-sized screens only after merging it... any chance you want to do another for that?

image

nop33:
will do!

nop33:
probably just need to add `flex-shrink: 0` to the icon, but I'll test later

Issue #134, opened 2019-10-23
Improve responsiveness of newdle summary page


pferreir:
![image](https://user-images.githubusercontent.com/179599/90772076-715c6680-e2f4-11ea-903b-1ad1cd0bf8d2.png)

hugonxc:
Can I work on this one?

ThiefMaster:
sure!

hugonxc:
Hey there! I have a few questions.
First one: I got and runtime error when I click on login button. I follow all steps on README, but I still got:
``
RuntimeError: The session is unavailable because no secret key was set.  Set the secret_key on the application to something unique and secret.
``
And: I have some default user to logging on devel server?
Thanks

ThiefMaster:
You need to add this to your `newdle.cfg`:

    SECRET_KEY = 'some random garbage here'

Also, you want `SKIP_LOGIN = True` so you are logged in as a dummy user when you click the login button.

hugonxc:
Awesome!

ThiefMaster:
I just added a commit to master that puts useful developer defaults in the config when creating it (the makefile won't be used for non-development setups anyway).

PR #146, opened 2019-10-25
Improve summary page responsiveness

hugonxc:
This is my collaboration related to issue #134, there you can see the previous behavior of the summary page. Now, these are the behaviors:
On the desktop view:
newdle_after
And the mobile view:
newdle_mobile_after
As you can see, I change the logo when the site goes to a responsive view. I made this image, which is in PNG format, by my self(I'm not a professional designer) so it's recommended to change it later by your official design pattern.

All tests run successfully.

I hope this PR can collaborate on this project.


newdle/migrations/versions/20191024_1629_000000000000_create_initial_tables.py
```diff
@@ -1,7 +1,7 @@
 """Create initial tables
 
 Revision ID: 000000000000
-Create Date: 2019-09-12 17:18:07.401658
+Create Date: 2019-10-24 16:29:39.913975
```
mic4ael: This shouldn't end up in the commit

newdle/client/src/components/TopHeader.module.scss
```diff
@@ -8,3 +8,11 @@
     height: 2.5em;
   }
 }
+
+.logo {
+  content: url('../images/logo.svg');
+
+  @media (max-width: 768px) {
+    content: url('../images/logo_mobile.png');
```
ThiefMaster: Why do we need a png for this? Shouldn't it be possible to just scale the SVG properly?

pferreir: Shouldn't this be solved by #144 anyway? Maybe this can just be removed.

ThiefMaster: Indeed, please rebase to the latest master to have all the responsive fixes related to the front page and global elements.

hugonxc:
If I use unstackable this is the result on mobile and desktop:
Desktop:
desk_unstack
Mobile
unstackable

Do you prefer this way?

ThiefMaster:
Please use rebases instead of merges to update your branch; otherwise we need to squash the commits when merging to avoid the mess merge commits in PR branches bring with them.

ThiefMaster:
About the UI, it looks a bit "squeezed" to me in the mobile view. BTW, make sure to submit some availability for the newdle to see how it looks like with participants being listed there.

PR #153, opened 2019-10-26
Use nicer availability ring in summary

### LINKED ISSUE #151: Replace react-circular-progressbar with something nicer that can show all three states
We are currently using `react-circular-progressbar` to show how many people are available on the summary page (in the [ParticipantTable](https://github.com/indico/newdle/blob/89815bc8adb9f52e260e31ec81109adf752c686f/newdle/client/src/components/ParticipantTable.js#L80-L91) component):

image

The library's rendering is not amazing - you can see some red on the outer edges of the green bar. Also, we have a third state in addition to available/unavailable, which is "if need be". The library does not support multiple states.

Instead of what we have now, it should:

- look clean (no weird colored "border")
- support all 3 states (red/yellow/green)

You can either find a library (nothing huge like D3, which is 250kb minified!) that does this, or add a new react component for it that generates the SVG itself.

rpanczer:
Replaced `react-circular-progressbar` with AvailabilityRing.js, a component that returns an svg and a counter to show meeting participant status.

AvailabilityRing.js support 3 states (Red, Yellow, Green) and displays a counter (below the ring) for the number of `accepted invitations / total invitations`.

Props include: 
-  available: number of people available
- ifNeeded: number of people available if needed
- totalParticipants: number of people invited
- strokeWidth: width of stroke (optional)
- radius: radius of ring (optional)

Example States:
image

image

image

image

Definitely open to feedback :)

closes #151

### DIFF
--- a/newdle/client/src/components/AvailabilityRing.js
+++ b/newdle/client/src/components/AvailabilityRing.js
@@ -1,0 +1,76 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function AvailabilityRing({
+  available,
+  ifNeeded,
+  totalParticipants,
+  strokeWidth,
+  radius,
+}) {
+  const size = radius * 1.1;
+  const normalizedRadius = radius - 14;
+  const circumference = normalizedRadius * 2 * Math.PI;
+
+  return (
+    <div>
+      <svg height={size * 2} width={size * 2}>
+        <circle
+          stroke="#00ac46"
+          fill="transparent"
+          strokeWidth={strokeWidth}
+          strokeDasharray={
+            circumference * (available / totalParticipants) +
+            ' ' +
+            circumference * ((totalParticipants - available) / totalParticipants)
+          }
+          transform={`rotate(-90 ${size} ${size})`}
+          r={normalizedRadius}
+          cx={size}
+          cy={size}
+        />
+        <circle
+          stroke="#fdc500"
+          fill="transparent"
+          strokeWidth={strokeWidth}
+          strokeDasharray={circumference * (ifNeeded / totalParticipants) + ' ' + circumference}
+          transform={`rotate(${-90 + (available / totalParticipants) * 360} ${size} ${size})`}
+          r={normalizedRadius}
+          cx={size}
+          cy={size}
+        />
+        <circle
+          stroke="#dc0000"
+          fill="transparent"
+          strokeWidth={strokeWidth}
+          strokeDasharray={
+            circumference * ((totalParticipants - (available + ifNeeded)) / totalParticipants) +
+            ' ' +
+            circumference
+          }
+          transform={`rotate(${-90 +
+            ((available + ifNeeded) / totalParticipants) * 360} ${size} ${size})`}
+          r={normalizedRadius}
+          cx={size}
+          cy={size}
+        />
+        <text x="50%" y="50%" textAnchor="middle" stroke="#000" strokeWidth=".3px" dy=".3em">
+          {available + '/' + totalParticipants}
+        </text>
+      </svg>
+    </div>
+  );
+}
+
+AvailabilityRing.propTypes = {
+  available: PropTypes.number.isRequired,
+  ifNeeded: PropTypes.number.isRequired,
+  totalParticipants: PropTypes.number.isRequired,
+  strokeWidth: PropTypes.number,
+  radius: PropTypes.number,
+};
+
+AvailabilityRing.defaultProps = {
+  strokeWidth: 10,
+  radius: 50,
+};
--- a/newdle/client/src/components/ParticipantTable.js
+++ b/newdle/client/src/components/ParticipantTable.js
@@ -58,11 +58,7 @@ function ParticipantNames({participants}) {
   }
 }
 
-function AvailabilityRow({
-  availability: {startDt, available, participationCount},
-  setActiveDate,
-  active,
-}) {
+function AvailabilityRow({availability: {startDt, available}, setActiveDate, active}) {
   const numberOfParticipants = useSelector(getNumberOfParticipants);
   const duration = useSelector(getNewdleDuration);
   const startTime = toMoment(startDt, 'YYYY-MM-DDTHH:mm');
@@ -81,8 +77,8 @@ function AvailabilityRow({
         <div className={styles['wrapper']}>
           <div className={styles['availability-indicator']}>
             <AvailabilityRing
-              available={participationCount.available}
-              ifNeeded={participationCount.ifneedbe}
+              available={available.filter(p => p.fullyAvailable).length}
+              ifNeeded={available.filter(p => !p.fullyAvailable).length}
               totalParticipants={numberOfParticipants}
             />
           </div>
--- a/newdle/client/src/components/ParticipantTable.js
+++ b/newdle/client/src/components/ParticipantTable.js
@@ -2,10 +2,9 @@ import React, {useState} from 'react';
 import moment from 'moment';
 import {useSelector} from 'react-redux';
 import {Radio, Icon, Label, Table} from 'semantic-ui-react';
-import {CircularProgressbar, buildStyles} from 'react-circular-progressbar';
+import AvailabilityRing from './AvailabilityRing';
 import {serializeDate, toMoment} from '../util/date';
 import {getNewdleDuration, getNumberOfParticipants, getParticipantAvailability} from '../selectors';
-import 'react-circular-progressbar/dist/styles.css';
 import styles from './ParticipantTable.module.scss';
 
 const MAX_PARTICIPANTS_SHOWN = 4;
@@ -59,7 +58,11 @@ function ParticipantNames({participants}) {
   }
 }
 
-function AvailabilityRow({availability: {startDt, available}, setActiveDate, active}) {
+function AvailabilityRow({
+  availability: {startDt, available, participationCount},
+  setActiveDate,
+  active,
+}) {
   const numberOfParticipants = useSelector(getNumberOfParticipants);
   const duration = useSelector(getNewdleDuration);
   const startTime = toMoment(startDt, 'YYYY-MM-DDTHH:mm');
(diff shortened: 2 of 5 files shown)


newdle/client/src/components/AvailabilityRing.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+
+AvailabilityRing.defaultProps = {
```
ThiefMaster: I'd move this after the function (while it works thanks to hoisting, it's a bit weird to access something before its definition)

newdle/client/src/components/ParticipantTable.js
```diff
@@ -77,17 +76,10 @@ function AvailabilityRow({availability: {startDt, available}, setActiveDate, act
...
-              styles={buildStyles({
-                textSize: '18px',
-                strokeLinecap: 'butt',
-                pathColor: numberOfParticipants === 0 ? 'gray' : 'lightgreen',
-                trailColor: 'red',
-                textColor: 'black',
-              })}
+            <ProgressRing
+              available={available.length}
+              ifNeeded={0}
```
ThiefMaster: Would you mind adding another commit to use actual data here?

I think the best option for that would be changing the `getParticipantAvailability` selector to also include the number of `ifneedbe` answers.

newdle/client/src/components/ParticipantTable.js
```diff
@@ -2,10 +2,9 @@ import React, {useState} from 'react';
 import moment from 'moment';
 import {useSelector} from 'react-redux';
 import {Radio, Icon, Label, Table} from 'semantic-ui-react';
-import {CircularProgressbar, buildStyles} from 'react-circular-progressbar';
```
ThiefMaster: Could you `npm remove` it (while inside `newdle/client`) so we don't have it as a dependency in `package{,-lock}.json` anymore?

newdle/client/src/components/ParticipantTable.js
```diff
@@ -2,10 +2,9 @@ import React, {useState} from 'react';
...
+import ProgressRing from './AvailabilityRing';
```
ThiefMaster: Let's import it as `AvailabilityRing` (or rename the file and function to `ProgressRing`)

newdle/client/src/components/AvailabilityRing.js
```diff
@@ -0,0 +1,76 @@
...
+
+  const greenOffset = `rotate(-90 ${size} ${size})`;
+  const yellowOffset = `rotate(${-90 + (available / totalParticipants) * 360} ${size} ${size})`;
+  const redOffset = `rotate(${-90 +
+    ((available + ifNeeded) / totalParticipants) * 360} ${size} ${size})`;
+
+  let style = {
+    'width': '25%',
+    'color': 'black',
+    'text-align': 'center',
```
ThiefMaster: This should be `textAlign` (since the style dict in React uses the JS names for style attributes, not the CSS ones). You can also get rid of the quotes around the keys now since everything is a valid JS identifier.

newdle/client/src/components/AvailabilityRing.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+export default function AvailabilityRing({
+  available,
+  ifNeeded,
+  totalParticipants,
+  strokeWidth,
+  radius,
+}) {
+  let size = radius * 1.1;
```
ThiefMaster: Please use `const` for anything that's not reassigned

newdle/client/src/components/AvailabilityRing.js
```diff
@@ -0,0 +1,76 @@
...
+    'width': '25%',
+    'color': 'black',
+    'text-align': 'center',
+    'margin': 'auto',
+  };
+  return (
+    <>
+      <svg height={size * 2} width={size * 2}>
+        <circle
+          stroke="#00ac46" // green - grün
```
ThiefMaster: heh, how did those half-German comments end up here? 🙈

newdle/client/src/components/AvailabilityRing.js
```diff
@@ -0,0 +1,76 @@
...
+            ' ' +
+            circumference
+          }
+          transform={redOffset}
+          r={normalizedRadius}
+          cx={size}
+          cy={size}
+        />
+      </svg>
+      <div style={style}>{available + '/' + totalParticipants}</div>
```
ThiefMaster: I think having this inside would look nicer - but only if we can properly center it. If not maybe it's better to keep it below.

newdle/client/src/components/AvailabilityRing.js
```diff
@@ -0,0 +1,76 @@
...
+  let stroke = 7;
+  let normalizedRadius = radius - stroke * 2;
+  let circumference = normalizedRadius * 2 * Math.PI;
+
+  const greenOffset = `rotate(-90 ${size} ${size})`;
+  const yellowOffset = `rotate(${-90 + (available / totalParticipants) * 360} ${size} ${size})`;
+  const redOffset = `rotate(${-90 +
+    ((available + ifNeeded) / totalParticipants) * 360} ${size} ${size})`;
+
+  let style = {
```
ThiefMaster: I'd move this to a CSS module; no need to have it in the JS code

ThiefMaster:
This already looks great! 🎉

Just a few things to improve/fix (see the review comments above).

Amending the commit message to use imperative form (something like "Use nicer availability ring in summary") would be nice as well (otherwise I need to do a manual rebase or squash with a new message when merging).

newdle/client/src/components/AvailabilityRing.js
rpanczer: I can move this - I originally thought to keep this in the .js because it is less a style and more a functional part of the component, even though it is css.

ThiefMaster: If it was for the circle itself I'd agree, but it's only for the text below so it I'd say it's more styling than a functional part.

rpanczer: Oh I misunderstood! I thought you were referring to redOffset, etc.

newdle/client/src/components/ParticipantTable.js
rpanczer: I made the change (and all the others) but I am not sure how to best test this one, since I was having trouble with logins (and can't accept the invite as other people).

ThiefMaster: With `SKIP_LOGINS` you only have one account, but if you access the newdle URL without `/summary` in an incognito tab, you can submit anonymously (ie creating a new participant whenever you submit).

You can also change the email backend in the config from `...smtp...` to `...console...` and then copy the personal link and use that one to answer as an invited participant.

rpanczer:
image

newdle/client/src/components/AvailabilityRing.js
```diff
@@ -0,0 +1,67 @@
...
+        />
+        <text x="50%" y="50%" textAnchor="middle" stroke="#000" strokeWidth=".3px" dy=".3em">
+          {available + '/' + totalParticipants}
+        </text>
+      </svg>
+    </div>
+  );
+}
+
+AvailabilityRing.defaultProps = {
```
ThiefMaster: Please add propTypes as well

newdle/client/src/selectors.js
```diff
@@ -104,10 +104,14 @@ export const getParticipantAvailability = createSelector(
   getNewdleParticipants,
   (timeslots, participants) => {
     return timeslots.map(timeslot => {
+      console.log('time', timeslot);
```
ThiefMaster: 👀 

(just to not forget about removing this later)

newdle/client/src/selectors.js
```diff
@@ -104,10 +104,14 @@ export const getParticipantAvailability = createSelector(
...
       const available = participants
-        .filter(part => ['available', 'ifneedbe'].includes(part.answers[timeslot]))
+        .filter(part => ['available'].includes(part.answers[timeslot]))
```
ThiefMaster: `part.answers[timeslot] === 'available'`

newdle/client/src/selectors.js
```diff
@@ -104,10 +104,14 @@ export const getParticipantAvailability = createSelector(
...
         .map(part => ({...part, fullyAvailable: part.answers[timeslot] === 'available'}));
-      return {startDt: timeslot, available};
+      const ifneedbe = participants
+        .filter(part => ['ifneedbe'].includes(part.answers[timeslot]))
```
ThiefMaster: `part.answers[timeslot] === 'ifneedbe'`

newdle/client/src/selectors.js
```diff
@@ -104,10 +104,14 @@ export const getParticipantAvailability = createSelector(
   getNewdleParticipants,
   (timeslots, participants) => {
     return timeslots.map(timeslot => {
+      console.log('time', timeslot);
       const available = participants
-        .filter(part => ['available', 'ifneedbe'].includes(part.answers[timeslot]))
+        .filter(part => ['available'].includes(part.answers[timeslot]))
         .map(part => ({...part, fullyAvailable: part.answers[timeslot] === 'available'}));
-      return {startDt: timeslot, available};
+      const ifneedbe = participants
```
ThiefMaster: This is pretty much repeating the statement above. I wonder if it wouldn't be nicer to define an object before (with available and ifneedbe set to 0), and then just using `.forEach()` and incrementing the respective counter in that object.

rpanczer: I think I followed you. I tried to implement `.forEach() in the most recent push`

ThiefMaster: Thanks, I'll have a closer look at the updated selector tomorrow... need to double-check what data we have there before commenting ;)

rpanczer: No problem - thanks for bearing with me on this commit.

pferreir:
Thanks, @rpanczer!

newdle/client/src/components/AvailabilityRing.js
```diff
@@ -0,0 +1,76 @@
...
+    <div>
+      <svg height={size * 2} width={size * 2}>
+        <circle
+          stroke="#00ac46"
+          fill="transparent"
+          strokeWidth={strokeWidth}
+          strokeDasharray={
+            circumference * (available / totalParticipants) +
+            ' ' +
+            circumference * ((totalParticipants - available) / totalParticipants)
```
ThiefMaster: @rpanczer i'm curious, what's the need for this calculation here? as far as I understood how this value works is that just using `circumference` here should be enough, since that way you never get more than one dash for that stroke

PR #156, opened 2019-10-29
Display create first newdle screen on my newdle screen if no newdles

### LINKED ISSUE #145: Show something useful on "My Newdles" page if there aren't any
Right now we show a rather ugly empty page if you go to "My Newdles" and don't have any.

It'd be nicer to show a message there that you don't have any newdles, and maybe even a button to create your first newdle.

rpanczer:
Display create first newdle screen on my newdle screen if no newdles - closes #145

image

### DIFF
--- a/newdle/client/src/components/MyNewdles.js
+++ b/newdle/client/src/components/MyNewdles.js
@@ -30,7 +30,8 @@ export default function MyNewdles() {
                 history.push('/new');
               }}
             >
-              Create your first Newdle! <span>🍜</span>
+              {/* eslint-disable-next-line jsx-a11y/accessible-emoji */}
+              Create your first newdle! 🍜
             </Button>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


newdle/client/src/components/MyNewdles.js
```diff
@@ -1,24 +1,45 @@
...
+          <div>
+            <Button
+              color="violet"
+              type="submit"
+              onClick={() => {
+                history.push('/new');
+              }}
+            >
+              Create your First Newdle!{' '}
+              <span role="img" aria-label="Newdle">
```
ThiefMaster: I don't think we need aria roles here since this icon is just to make the button prettier - if someone cannot see icons (e.g. in a screenreader), they do not miss any important information by not having it.

And it would probably be confusing if it was read as "Create your first newdle. Newdle." or something similar...

rpanczer:
@mic4ael I agree, that's more readable.

PR #161, opened 2019-10-29
adding contributors to readme

jrstrayhorn:
I know this wasn't asked for but I used https://gh-hof.netlify.com/ to generate the contributors for your project and added them to the readme.

Hopefully this adds value to your repo.

FYI - I didn't add myself because I wasn't sure if this would get included.. if it does or if you guys like this I can add mine in 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,3 +86,17 @@ We provide a couple of additional `make` targets that should streamline the deve
 > In applying the MIT license, CERN does not waive the privileges and immunities
 > granted to it by virtue of its status as an Intergovernmental Organization
 > or submit itself to any jurisdiction.
+
+## Contributors
+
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Hey, thanks for the PR (and sorry for the long delay in answering).

We still need to discuss among the team whether we want to include this in our readme or not, but in case we decide that we do want this we would certainly not want to have to maintain this manually but rather use GitHub actions to automatically update it.

jrstrayhorn:
ah! that totally makes sense.. I appreciate the reply!!

ThiefMaster:
Closing this for now as we did not really find a consensus on whether we want to have this or not - and if we wanted to add it, this would need to be automated anyway and not rely on manual updates.

PR #163, opened 2019-11-04
Fix the handling of timezones in free/busy views

pferreir:
* This PR makes it so that the free/busy views take into account the current timezone the user is in.
 * This means as well that, during the creation of the newdle, updates in the timezone field will result in the free/busy information being fetched again.
 * I have also fixed the handling of free/busy dates that fall outside the visible hour range.
 * In what concerns the Exchange part, I have made sure we operate using the account's timezone and then properly calculate the overlaps based on the newdle's timezone;

### DIFF
--- a/newdle/providers/free_busy/exchange.py
+++ b/newdle/providers/free_busy/exchange.py
@@ -1,28 +1,55 @@
 from datetime import datetime, time, timedelta
 
+import pytz
 from exchangelib import NTLM, Account, Configuration, Credentials
 from exchangelib.errors import (
     ErrorMailRecipientNotFound,
     ErrorProxyRequestProcessingFailed,
 )
 from flask import current_app
-from pytz import utc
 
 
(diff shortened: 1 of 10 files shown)


newdle/providers/free_busy/exchange.py
```diff
@@ -46,8 +73,13 @@ def fetch_free_busy(date, uid):
         ),
     ]
 
-    tz = uid_account.default_timezone
-    start = utc.localize(datetime.combine(date, time.min)).astimezone(tz)
+    if tz in NON_STANDARD_TZS:
```
ThiefMaster: tz = NON_STANDARD_TZS.get(tz, tz)

pferreir: Yeah, thought of it but it looked dirtier that way... "explicit better than implicit" etc...

meluru: +1 to more explicit code here

PR #171, opened 2019-12-10
Add "newdles I'm in" page to user profile

### LINKED ISSUE #164: Add "newdles I'm in" page to user profile
This should be similar to "My newdles".

image

If anyone external wants to work on this, starting with a copy of `MyNewdles.js` is probably a good idea. Of course you will still need to add backend and a client method for it, but that should be pretty straightforward. The new backend function should come with a unit test.

plourenco:
Closes #164 

Some notes:
- Should "expired" newdles, i.e., with final date set be excluded?
- This PR doesn't change any of the page styles, should we add something to distinguish the two pages? A title such as "The newdles you are part of" and "The newdles you created", or merging the two together in the same page.
- A suggestion for an interesting feature would be, replacing the placeholder "x out of X participants answered" with something like "You chose the '2019-09-11T13:00' slot" or "You chose 3 slots", for the newdles-im-in page, since you can't see the participants.

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -1,5 +1,4 @@
 import uuid
-from datetime import date
 from importlib import import_module
 
 from faker import Faker
@@ -216,12 +215,7 @@ def get_newdles_participating():
     newdle = (
         Participant.query.filter_by(auth_uid=g.user['uid'])
         .join(Participant.newdle)
-        .order_by(
-            Participant.answers != '{}',
-            Newdle.final_dt >= date.today(),
-            Newdle.final_dt,
-            Newdle.id.desc(),
-        )
+        .order_by(Newdle.id.desc())
     )
     return NewdleParticipantSchema(many=True).jsonify(newdle)
 
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -1,4 +1,5 @@
 import uuid
+from datetime import date
 from importlib import import_module
 
 from faker import Faker
@@ -7,7 +8,7 @@ from itsdangerous import BadData, SignatureExpired
 from marshmallow import fields
 from marshmallow.validate import OneOf
 from pytz import common_timezones_set
-from sqlalchemy.orm import noload, selectinload
+from sqlalchemy.orm import selectinload
 from werkzeug.exceptions import Forbidden, UnprocessableEntity
 
 from .cern_integration import search_cern_users
@@ -19,6 +20,7 @@ from .models import Newdle, Participant
 from .notifications import notify_newdle_participants
 from .schemas import (
     MyNewdleSchema,
+    NewdleParticipantSchema,
     NewdleSchema,
     NewNewdleSchema,
     NewUnknownParticipantSchema,
@@ -209,15 +211,19 @@ def get_my_newdles():
     return MyNewdleSchema(many=True).jsonify(newdle)
 
 
-@api.route('/newdles/in')
-def get_newdles_im_in():
+@api.route('/newdles/participating')
+def get_newdles_participating():
     newdle = (
-        Newdle.query.options(noload('participants'))
-        .filter(Newdle.participants.any(auth_uid=g.user['uid']))
-        .order_by(Newdle.final_dt.isnot(None), Newdle.final_dt.desc(), Newdle.id.desc())
-        .all()
+        Participant.query.filter_by(auth_uid=g.user['uid'])
+        .join(Participant.newdle)
+        .order_by(
+            Participant.answers != '{}',
+            Newdle.final_dt >= date.today(),
+            Newdle.final_dt,
+            Newdle.id.desc(),
+        )
     )
-    return RestrictedNewdleSchema(many=True).jsonify(newdle)
+    return NewdleParticipantSchema(many=True).jsonify(newdle)
 
 
 @api.route('/newdle/', methods=('POST',))
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -7,7 +7,7 @@ from itsdangerous import BadData, SignatureExpired
 from marshmallow import fields
 from marshmallow.validate import OneOf
 from pytz import common_timezones_set
-from sqlalchemy.orm import selectinload
+from sqlalchemy.orm import noload, selectinload
 from werkzeug.exceptions import Forbidden, UnprocessableEntity
 
 from .cern_integration import search_cern_users
@@ -207,6 +207,17 @@ def get_my_newdles():
         .all()
     )
     return MyNewdleSchema(many=True).jsonify(newdle)
+
+
+@api.route('/newdles/in')
+def get_newdles_im_in():
+    newdle = (
+        Newdle.query.options(noload('participants'))
+        .filter(Newdle.participants.any(auth_uid=g.user['uid']))
+        .order_by(Newdle.final_dt.isnot(None), Newdle.final_dt.desc(), Newdle.id.desc())
+        .all()
+    )
+    return RestrictedNewdleSchema(many=True).jsonify(newdle)
 
 
 @api.route('/newdle/', methods=('POST',))
--- a/newdle/client/src/components/App.js
+++ b/newdle/client/src/components/App.js
@@ -10,7 +10,7 @@ import TopHeader from './TopHeader';
 import LoginPrompt from './login/LoginPrompt';
 import LoggingIn from './login/LoggingIn';
 import MyNewdles from './MyNewdles';
-import NewdlesImIn from './NewdlesImIn';
+import NewdlesParticipating from './NewdlesParticipating';
 import ErrorMessage from './ErrorMessage';
 
 import './App.module.scss';
@@ -31,7 +31,7 @@ export default function App() {
           <Route exact path="/new" component={CreationPage} />
           <Route exact path="/new/success" component={CreationSuccessPage} />
           <Route exact path="/mine" component={MyNewdles} />
-          <Route exact path="/in" component={NewdlesImIn} />
+          <Route exact path="/participating" component={NewdlesParticipating} />
           <Route path="/newdle/:code/summary" component={SummaryPage} />
           <Route exact path="/newdle/:code/:partcode?" component={AnswerPage} />
           <Route render={() => <div>This page does not exist</div>} />
--- a/newdle/client/src/components/App.js
+++ b/newdle/client/src/components/App.js
@@ -10,6 +10,7 @@ import TopHeader from './TopHeader';
 import LoginPrompt from './login/LoginPrompt';
 import LoggingIn from './login/LoggingIn';
 import MyNewdles from './MyNewdles';
+import NewdlesImIn from './NewdlesImIn';
 import ErrorMessage from './ErrorMessage';
 
 import './App.module.scss';
@@ -30,6 +31,7 @@ export default function App() {
           <Route exact path="/new" component={CreationPage} />
           <Route exact path="/new/success" component={CreationSuccessPage} />
           <Route exact path="/mine" component={MyNewdles} />
+          <Route exact path="/in" component={NewdlesImIn} />
           <Route path="/newdle/:code/summary" component={SummaryPage} />
           <Route exact path="/newdle/:code/:partcode?" component={AnswerPage} />
           <Route render={() => <div>This page does not exist</div>} />
--- a/newdle/client/src/components/NewdlesImIn.js
+++ b/newdle/client/src/components/NewdlesImIn.js
@@ -1,0 +1,74 @@
+import React from 'react';
+import {Container, Icon, Label, Placeholder} from 'semantic-ui-react';
+import {useHistory} from 'react-router';
+import {serializeDate, toMoment} from '../util/date';
+import client from '../client';
+import {usePageTitle} from '../util/hooks';
+import styles from './MyNewdles.module.scss';
+
+export default function NewdlesImIn() {
+  const [newdles, loading] = client.useBackend(() => client.getNewdlesImIn(), []);
+  usePageTitle("Newdles I'm in");
+
+  let content;
+  if (loading || newdles === null) {
+    content = (
+      <>
+        <Placeholder className={styles.newdle} />
+        <Placeholder className={styles.newdle} />
+        <Placeholder className={styles.newdle} />
+      </>
+    );
+  } else if (newdles.length === 0) {
+    content = (
+      <div className={styles['no-newdle-container']}>
+        {/* eslint-disable-next-line jsx-a11y/accessible-emoji */}
+        <h2>You are not part of any newdles yet... 🍜</h2>
+      </div>
+    );
+  } else {
+    content = newdles.map(newdle => <NewdleAsParticipant key={newdle.id} newdle={newdle} />);
+  }
+
+  return (
+    <Container text className={styles.newdles}>
+      {content}
+    </Container>
+  );
+}
+
+function NewdleAsParticipant({newdle: {code, title, duration, final_dt: finalDT, timezone}}) {
+  const history = useHistory();
+  const startTime = finalDT ? serializeDate(finalDT, 'HH:mm') : undefined;
+  const endTime = finalDT
+    ? serializeDate(toMoment(finalDT).add(duration, 'm'), 'HH:mm')
+    : undefined;
+  const url = `/newdle/${code}`;
+
+  return (
+    <div className={styles.newdle} onClick={() => history.push(url)}>
+      <h3 className={styles.title}>
+        <a href={url} onClick={evt => evt.preventDefault()}>
+          {title}
+        </a>
+        <Label color={finalDT ? 'blue' : 'green'} size="tiny" className={styles.state}>
+          {finalDT ? 'finished' : 'ongoing'}
+        </Label>
+      </h3>
+      {finalDT && (
+        <div className={styles.info}>
+          <span>
+            <Icon name="calendar alternate outline" />
+            <label>{serializeDate(finalDT, 'dddd, D MMMM')}</label>
+          </span>
+          <span>
+            <Icon name="clock outline" />
+            <label>
+              {`${startTime} - ${endTime}`} ({timezone})
+            </label>
+          </span>
+        </div>
+      )}
+    </div>
+  );
+}
--- a/newdle/client/src/components/NewdlesParticipating.js
+++ b/newdle/client/src/components/NewdlesParticipating.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import flask from 'flask-urls.macro';
 import {Container, Icon, Label, Placeholder} from 'semantic-ui-react';
 import {useHistory} from 'react-router';
 import {serializeDate, toMoment} from '../util/date';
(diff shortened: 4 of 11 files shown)


ThiefMaster:
(i haven't had a look at the code yet, so this comment is just based on your notes)

> Should "expired" newdles, i.e., with final date set be excluded?

No, I don't think so. These may actually be the most interesting ones (at least while the date is in the future) if you didn't notice the email (or the organizer didn't send one)

> This PR doesn't change any of the page styles, should we add something to distinguish the two pages?

I haven't realized that we don't show a title on "My Newdles". We should indeed make it somehow clear on which page you are. I think this could be a separate task/PR since this one is just about adding the new page (unless you want to give it a try of course). One option would be showing it similar to the "Welcome to newdle!" message on the front page.

> replacing the placeholder "x out of X participants answered"

Very good point, especially since  - as you mentioned - this information is currently not available to anyone but the creator of a newdle. Showing information about the slots you chose seems like a great idea here!

ThiefMaster:
I'm not sure why it didn't happen automatically for you (through the `husky` dependency that should have automatically installed a `pre-commit` hook during the initial `make`) when doing `git commit`, but could you run `make format`, amend and force-push?

(due to #172 CI didn't fail even though it should have.)

pedroo21:
Sorry, that was my problem. I had a previous issue while building the wheels for regex (while installing black) and psycopg2, using MacOS 10.15, which caused the git hooks (the next step) not to be setup.
I've updated the commit. Thanks for your suggestions, I will add those as separate commits!

ThiefMaster:
Thanks! There's still some weirdness with the autoformatting; as if some of the Python dependencies aren't seen as installed thirdparty libs by isort. Since this is completely out of the scope of this task, here's the patch that will switch to the correct import formatting:

```patch
diff --git a/conftest.py b/conftest.py
index cfa62f4..19a717d 100644
--- a/conftest.py
+++ b/conftest.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta

 import flask_migrate
 import pytest
+
 from newdle.core.app import create_app
 from newdle.core.db import db
 from newdle.models import Newdle, Participant
diff --git a/newdle/api.py b/newdle/api.py
index ae463d7..973ec4b 100644
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -1,15 +1,14 @@
 import uuid
 from importlib import import_module

+from faker import Faker
 from flask import Blueprint, current_app, g, jsonify, request, url_for
 from itsdangerous import BadData, SignatureExpired
-from werkzeug.exceptions import Forbidden, UnprocessableEntity
-
-from faker import Faker
 from marshmallow import fields
 from marshmallow.validate import OneOf
 from pytz import common_timezones_set
 from sqlalchemy.orm import noload, selectinload
+from werkzeug.exceptions import Forbidden, UnprocessableEntity

 from .cern_integration import search_cern_users
 from .core.auth import user_info_from_app_token
diff --git a/tests/api_test.py b/tests/api_test.py
index adaa67c..80d0dd9 100644
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -1,9 +1,9 @@
 from datetime import datetime, timedelta
 from operator import itemgetter

+import pytest
 from flask import url_for

-import pytest
 from newdle.core.auth import app_token_from_id_token
 from newdle.models import Newdle, Participant

```

You could also try to `make clean` and `make` to reset and recreate the virtualenv etc.

newdle/client/src/components/NewdlesImIn.js
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    <Container text className={styles.newdles}>
+      {content}
+    </Container>
+  );
+}
+
+function NewdleAsParticipant({newdle: {code, title, duration, final_dt: finalDT, timezone}}) {
+  const history = useHistory();
+  const startTime = finalDT ? serializeDate(finalDT, 'HH:mm') : undefined;
```
pferreir:
```js
const startTime = finalDT && serializeDate(finalDT, 'HH:mm');
```
is less explicitly but more idiomatic (and does the same thing in this case). Nothing wrong about the way you're doing it though.

pferreir:
Thanks for the PR! I just left a couple of very minor things, since @ThiefMaster has already covered pretty much everything (and maybe a few things more 😉 ).

newdle/client/src/components/NewdlesImIn.js
pferreir: OK, @ThiefMaster has just pointed out to me that this is probably copy-paste from the other page, so it's something which was already there (and somehow managed to evade the code review). So, never mind, we'll fix both later.

tests/api_test.py
```diff
@@ -212,6 +212,41 @@ def test_get_newdle(flask_client, dummy_newdle):
     }
 
 
+@pytest.mark.usefixtures('db_session')
+def test_newdles_im_not_in(flask_client, dummy_uid):
+    resp = flask_client.get(
+        url_for('api.get_newdles_im_in'), **make_test_auth(dummy_uid)
+    )
+    assert resp.status_code == 200
+    assert resp.json == []
```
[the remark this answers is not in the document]
pedroo21: Yes, I agree. Even though I would probably also support that == [] is more explicit about the check type.

newdle/client/src/components/NewdlesImIn.js
pedroo21: I agree, I prefer your suggestion, I've been writing most code in the same way. I didn't cover this case exactly by the same reason as the second comment, keep both in the same format. But I can start by adding this change here.

newdle/client/src/components/App.js
```diff
@@ -10,6 +10,7 @@ import TopHeader from './TopHeader';
 import LoginPrompt from './login/LoginPrompt';
 import LoggingIn from './login/LoggingIn';
 import MyNewdles from './MyNewdles';
+import NewdlesImIn from './NewdlesImIn';
```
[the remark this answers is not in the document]
pedroo21: Yes, I agree. The naming conventions I used need some work.
 I would suggest `NewdlesImParticipating`, applied here as well as `get_newdles_im_participating()` for the API call.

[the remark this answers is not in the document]
ThiefMaster: I'd like to avoid the (somewhat ugly) `Im` / `_im_` there

newdle/api.py
```diff
@@ -200,6 +200,17 @@ def get_my_newdles():
...
 
+@api.route('/newdles/in')
+def get_newdles_im_in():
+    newdle = (
+        Newdle.query.options(noload('participants'))
+        .filter(Newdle.participants.any(auth_uid=g.user['uid']))
+        .order_by(Newdle.final_dt.isnot(None), Newdle.final_dt.desc(), Newdle.id.desc())
+        .all()
+    )
+    return RestrictedNewdleSchema(many=True).jsonify(newdle)
```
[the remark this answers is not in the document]
pedroo21: Yes, I think using the Participant would be ideal. Initially, I didn't choose it because it felt more logical to query for Newdles and return just a list of them (omitting any participants), just like "MyNewdles". But I'll test the same approach using the Participants.

newdle/api.py
```diff
@@ -200,6 +200,17 @@ def get_my_newdles():
     return MyNewdleSchema(many=True).jsonify(newdle)
 
 
+@api.route('/newdles/in')
+def get_newdles_im_in():
+    newdle = (
+        Newdle.query.options(noload('participants'))
+        .filter(Newdle.participants.any(auth_uid=g.user['uid']))
+        .order_by(Newdle.final_dt.isnot(None), Newdle.final_dt.desc(), Newdle.id.desc())
```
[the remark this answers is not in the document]
pedroo21: I'm assuming a newdle someone has been invited to, but didn't reply, should reflect a participant entry without answers (an empty JSON object {}).

The most naive approach would be to add a sql filter when the "answers" column is '{}' (which also includes converting the field to public). The alternative is to do a post filter.

Do you have a preference or any other suggestions?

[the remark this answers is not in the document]
ThiefMaster: Checking for an empty object should work, and it's indeed the way how to check if no answers have been submitted yet.

The reason why it's `_answers` (assuming that's what you mean with it not being public right now) is that we needed the property to convert to a nicer format in the Python side.

If you make it an `@hybrid_property` (from sqlalchemy) instead of `@property`, you can add an `@answers.expression` method (classmethod-ish, i.e. `cls` as the first arg) which just returns `cls._answers`. That way you can use `Participant.answers` when constructing the query.

newdle/api.py
```diff
@@ -200,15 +203,19 @@ def get_my_newdles():
...
+def get_newdles_participating():
     newdle = (
-        Newdle.query.options(noload('participants'))
-        .filter(Newdle.participants.any(auth_uid=g.user['uid']))
-        .order_by(Newdle.final_dt.isnot(None), Newdle.final_dt.desc(), Newdle.id.desc())
-        .all()
+        Participant.query.filter_by(auth_uid=g.user['uid'])
+        .join(Participant.newdle)
+        .order_by(
+            desc(Participant.answers == '{}'),
```
ThiefMaster: Probably a matter of style, but you could just use `!=` instead of `==` here and then get rid of the `desc()` wrapper. Same in the next line by negating the check.

FWIW, I think both versions are somewhat ugly in terms of readability, but AFAIK there's no nicer way of sorting by the result of a comparison...

newdle/client/src/components/NewdlesParticipating.js
```diff
@@ -69,6 +82,18 @@ function NewdleAsParticipant({newdle: {code, title, duration, final_dt: finalDT,
...
         </div>
       )}
+      <div className={styles.participants}>
+        <Icon name="clock outline" />
+        <label>
+          <em>
+            {finalDT
+              ? 'This newdle has finished'
+              : !slotsChosen.length
+              ? 'Pending your reply'
```
ThiefMaster: "Awaiting" instead of "Pending"?

newdle/api.py
```diff
@@ -7,7 +8,8 @@
 from marshmallow import fields
 from marshmallow.validate import OneOf
 from pytz import common_timezones_set
-from sqlalchemy.orm import selectinload
+from sqlalchemy import asc, desc, func
```
ThiefMaster: some of the imports (e.g. `asc`) are unused (see the CI failures)

pedroo21: Sorry, I accidentally missed that. I'm just about to force push.

newdle/api.py
pedroo21: Agree, I tried both approaches, though the previous was more explicit. But using a consistent format looks nicer.

newdle/api.py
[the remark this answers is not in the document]
pedroo21: Yes, I figured we probably wanted to keep the raw `_answers` private.
Thanks, I never knew about sqlalchemy's @hybrid_property annotation. I will add that.

pedroo21:
> I think having another unit test for the sorting would be useful!

Yes, agree, that feature requires a test. I will add it as a separate commit.

newdle/client/src/components/NewdlesParticipating.js
```diff
@@ -0,0 +1,99 @@
...
+          </span>
+        </div>
+      )}
+      <div className={styles.participants}>
+        <Icon name="clock outline" />
+        <label>
+          <em>
+            {finalDT
+              ? 'This newdle has finished'
+              : !slotsChosen.length
```
ThiefMaster: This check isn't correct; if you do not choose any slots because you are unavailable, it does not mean it's still awaiting a reply.

newdle/api.py
```diff
@@ -200,6 +202,21 @@ def get_my_newdles():
     return MyNewdleSchema(many=True).jsonify(newdle)
 
 
+@api.route('/newdles/participating')
+def get_newdles_participating():
+    newdle = (
+        Participant.query.filter_by(auth_uid=g.user['uid'])
+        .join(Participant.newdle)
+        .order_by(
+            Participant.answers != '{}',
```
ThiefMaster: There's one case we forgot: A newdle you didn't answer, but that has been finalized in the meantime. Adding a `Newdle.final_dt.isnot(None)` sort criterion at the very beginning should fix this.

newdle/client/src/components/NewdlesParticipating.js
pedroo21: Nice catch. I skipped that part because I had the idea of treating unavailable responses with a different message.
Should I just show "You replied with X timeslot(s)" in any case, independently of the type of response?

ThiefMaster: "You are not available" or something similar could work as well

newdle/api.py
pedroo21: Thanks. I wasn't counting for expired newdles.
The criterion I'm now using is:
- show newdles without an answer and final date first (require attention)
- then newdles with a final date in the future (sorted by recent ones)
- then newdles without a final date (on-going)
- then past ones
> (unless we want to hide them by default, maybe behind a "show past" button - but that'd be a separate task). @ThiefMaster 

I ended up updating the sorting method to the following:
```
.order_by(
    or_(Participant.answers != '{}', Newdle.final_dt.isnot(None)),
    or_(Newdle.final_dt <= date.today(), Newdle.final_dt.is_(None)),
    Newdle.final_dt.isnot(None),
    Newdle.final_dt,
    Newdle.id.desc(),
)
```

Starting from the bottom, the idea is:
- Set up the base order to use the id descending, as previously
- Order the newdles by their final date ascending, in order to keep upcoming newdle dates sorted by their start
- Push newdles without a final date to the top (the on-going, to be determined ones)
- Push the upcoming, non-expired, newdles to the top (the relevant ones). However, ignore newdles without any final date. Because this comparison sets me the Nulls last.
- Show the newdles that require your immediate attention, without any final answers, first.

I was trying to avoid using the somewhat 'ugly' or_ or and_, but date comparisons were affecting the nulls. Nevertheless, I'm open to any suggestions for an alternative.
I will commit soon with a test that validates this approach.

ThiefMaster: For `or_`/`and_ `vs `|`/`&` I usually decide like this:

- does it fit in one line? `|` and `&` it is (with parentheses around each of the operands)
- does it need to be wrapped? `and_` and `or_` are much more readable!

I think in this case using the operators wouldn't be so bad to use them..

ThiefMaster: Technicalities aside, should we consider doing some grouping in the UI? Right now everything is in the same list, so it doesn't look super amazing if there's more than a few entries.

And if we sent the different categories in separate lists (ie an object as the top-level response item), we could simplify some of the sorting logic.

cc @indico/development-team

pedroo21: I agree with the grouping. The sorting logic is becoming a bit too complex. Especially since the solution I provided sacrifices sorting past newdles correctly. While their priority is low, separate lists keep the view clean and easier to maintain.

newdle/client/src/components/NewdlesParticipating.js
```diff
@@ -0,0 +1,108 @@
...
+      <>
+        <Placeholder className={styles.newdle} />
+        <Placeholder className={styles.newdle} />
+        <Placeholder className={styles.newdle} />
+      </>
+    );
+  } else if (participations.length === 0) {
+    content = (
+      <div className={styles['no-newdle-container']}>
+        {/* eslint-disable-next-line jsx-a11y/accessible-emoji */}
```
ThiefMaster: not needed, there aren't any emoji in the next line

newdle/client/src/components/NewdlesParticipating.js
```diff
@@ -0,0 +1,108 @@
...
+        <h2>You are not part of any newdles yet.</h2>
+      </div>
+    );
+  } else {
+    participations
+      .sort((a, b) => new Date(a.newdle.final_dt) - new Date(b.newdle.final_dt))
```
ThiefMaster: I think some comments on each of the sorting lines could be helpful, since it'd be easier to read english instead of javascript to understand the sorting logic here ;)

newdle/client/src/components/NewdlesParticipating.js
```diff
@@ -0,0 +1,108 @@
...
+      {finalDT && (
+        <div className={styles.info}>
+          <span>
+            <Icon name="calendar alternate outline" />
+            <label>{serializeDate(finalDT, 'dddd, D MMMM')}</label>
+          </span>
+          <span>
+            <Icon name="clock outline" />
+            <label>
+              {`${startTime} - ${endTime}`} ({timezone})
```
ThiefMaster: `{startTime} - {endtime} ({timezone)`?

newdle/api.py
```diff
@@ -1,12 +1,13 @@
 import uuid
 from importlib import import_module
 
+from pytz import common_timezones_set
```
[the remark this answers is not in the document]
panagiotappl: Maybe the environment was not activated and this was automatically sorted like this? (I recall I had the same problem or similar)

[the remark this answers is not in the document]
ThiefMaster: I think using `isort` from outside the env could cause the same issue. And I just realized again that we aren't running isort during CI (see #172). Since that's a small one maybe it'd be nice to fix that issue first and then rebase this PR once it's merged to have the CI checks run on the changes here?

newdle/client/src/components/NewdlesParticipating.js
pedroo21: Not sure how that got through. Thanks!

newdle/client/src/components/NewdlesParticipating.js
pedroo21: Agree.

newdle/client/src/components/NewdlesParticipating.js
pedroo21: This was copied from `MyNewdles.js`. Should I keep it, or apply the same style in both?

ThiefMaster: i'd change it in both places unless it results in some ugly way of wrapping after running prettier

newdle/api.py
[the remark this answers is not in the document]
pedroo21: It was most likely due to the environment not being activated. As running it again restored the original formatting.

newdle/client/src/components/NewdlesParticipating.js
pedroo21: Before resolving this conversion, I added unavailable responses as part of "You've replied with X timeslot(s)", which should simplify the logic.

ThiefMaster: That sounds like a good solution!

newdle/client/src/components/NewdlesParticipating.js
```diff
@@ -0,0 +1,110 @@
...
+import client from '../client';
+import {usePageTitle} from '../util/hooks';
+import styles from './MyNewdles.module.scss';
+
+export default function NewdlesParticipating() {
+  const [participations, loading] = client.useBackend(() => client.getNewdlesParticipating(), []);
+  usePageTitle("Newdles I'm participating in");
+
+  let content;
+  if (loading || participations === null) {
```
mic4ael: I am wondering whether `participations === null` should be another `else if` clause. Because e.g. if `loading === false` and for some reason `participations` will evaluate to `null` we will end up with placeholders which will not properly show the current state of the application. what about adding `else if (participations === null) return null`

pedroo21: I agree, `participations === null` shouldn't show a loading state, this part was cloned from the previous implemented `MyNewdles.js`. However, I've been experimenting with both approaches and noticed the `loading` state isn't being enabled in this view because the backend reducer is expecting a `'submit'` action type and none is triggered for this fetch request. Just a `'success'` at the end of the call.

```
function backendReducer(state, action) {
  switch (action.type) {
    case 'submit':
      return {...state, submitting: true, error: '', result: null};
    case 'error':
      return {...state, submitting: false, error: action.error};
    case 'success':
      return {...state, submitting: false, result: action.result};
    default:
      return state;
  }
}
```

In practice, participations should always return empty, except when an error happens. So I'm not sure what is worth changing in this case, either:
- Refactor the loading state to always dispatch a 'submit' action before the API call and  'success'.
- Use the `participations === null` both as a loading and error state.

PR #176, opened 2020-02-14
Add free-busy backend for CERN Kopano

ThiefMaster:
No need to merge this until until the API is available in production, just to make sure everything actually works fine then (it does work fine against a local test instance though).


ThiefMaster:
This is no longer needed.. 🗑️

PR #181, opened 2020-03-05
Add the missing lint scripts

### LINKED ISSUE #172: `make lint` should also check that prettier, isort and black would not apply any changes
Right now CI doesn't fail if you have code not meeting the rules of these tools.

pedroo21:
Update the lint step to check the code style without applying any changes.
Also, lock the prettier version to 1.18 to prevent 1.19 new function literal heuristic (refer to prettier's #6921).

Closes #172

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,11 @@
   "private": true,
   "scripts": {
     "prettier": "prettier --write 'newdle/client/src/**/*.{js,jsx,json,scss,css}'",
+    "prettier:check": "prettier --check 'newdle/client/src/**/*.{js,jsx,json,scss,css}'",
     "isort": "isort --recursive setup.py conftest.py newdle/ tests/",
+    "isort:check": "npm run isort -- --diff --check-only",
     "black": "black setup.py conftest.py newdle/ tests/",
+    "black:check": "npm run black -- --diff --check",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,8 +2,11 @@
   "name": "root",
   "private": true,
   "scripts": {
+    "prettier:check": "npm run prettier -- --check",
```
ThiefMaster: that's quite cool, i totally forgot that you can pass args to another `npm run` command! I'd just move the `:check` versions after the normal commands

package.json
```diff
@@ -2,8 +2,11 @@
...
     "prettier": "prettier --write 'newdle/client/src/**/*.{js,jsx,json,scss,css}'",
+    "isort:check": "npm run isort -- --check-only",
```
ThiefMaster: I'd include `--diff` so it prints the diff of the failure to the CI output

(same for the other commands; it's nicer if you can make them print diffs in case they'd change something)

pedroo21: Agree!

package.json
pedroo21: Yes, the previous solution was getting too much verbose. I made an amend to move the checks.

Issue #160, opened 2019-10-29
Add (python) tests for API functions that aren't covered yet


ThiefMaster:
`api_test.py` has already a bunch of tests, but not all API functions are covered yet.

Missing ones:

- [x] `get_busy_times` - mock `_get_busy_times`; just ensure it requires authentication
- [ ] `get_participant_busy_times` - same here, but cover the logic that checks whether it will return busy times or fail with an error
- [x] `get_my_newdles` - make sure that it does not return participant codes in the response
- [x] `update_newdle`
- [x] `get_participant_me`
- [x] `create_unknown_participant`
- [x] `create_participant`
- [ ] `send_result_emails` - use the `mail_queue` fixture to check that emails are sent

To skip for now:

- `users` - makes more sense once we have pluggable user directories

---

Note: Especially if you are an external contributor, feel free to add just some of the tests listed above. You do not need to send a big PR adding tests for everything (but you are of course welcome to do so if you'd like)

pablostellet:
Hi @ThiefMaster,
Can I help on this task? :)

ThiefMaster:
Sure! If you want to work on any specific test please mention it in here in case someone else wants to work on it as well to avoid two people doing the same.

pablostellet:
Alright, I will run the project first then pick one and say here 😄

ThiefMaster:
@pablostellet Are you still going to work on this?

While there's nothing wrong with deciding to not contribute, it would be nice to at least leave a comment in such a case!

tomasr8:
I'd like to give this one a try ;)

pjseweryn:
Hi, I would like to help with this task as well, in particular with send_result_emails case as agreed with tomasr8.

ThiefMaster:
Hey, as long as you don't both work on the exact same test individually (ie duplicate work for no benefit) or actually work together I'm perfectly fine with that!

Biboba:
I guess that one can be closed ?
- ```get_participant_busy_times``` was solved in https://github.com/indico/newdle/pull/255
- ```send_result_emails``` was solved in https://github.com/indico/newdle/pull/259

PR #182, opened 2020-03-28
Add python tests for api.py

Biboba:
Add some of the missing tests in 'api_test.py' (#160):

- add test for create_participant
- create_unknown_participant
- get_participant_me
- update_newdle
- get_my_newdles

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -273,18 +273,17 @@ def test_update_newdle_unauthorized(flask_client, dummy_newdle):
         json={'title': 'foo'},
         **make_test_auth('someone'),
     )
-    assert resp.status_code == Forbidden.code
+    assert resp.status_code == 403
     assert resp.json == {'error': Forbidden.description}
 
 
 @pytest.mark.usefixtures('dummy_newdle')
 def test_update_newdle(flask_client, dummy_newdle, dummy_uid):
     final_dt = '2019-09-12T13:30'
-    dummy_newdle_json = {
+    expected_json = {
         'code': 'dummy',
         'creator_name': 'Dummy',
         'duration': 60,
-        'final_dt': None,
         'id': dummy_newdle.id,
         'timeslots': [
             '2019-09-11T13:00',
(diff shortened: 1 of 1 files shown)


tests/api_test.py
```diff
@@ -219,6 +255,74 @@ def test_get_newdle(flask_client, dummy_newdle):
...
+        'timezone': 'Europe/Paris',
+        'title': 'Test event1',
+        'url': 'http://flask.test/newdle/dummy1',
+    }
+    resp = flask_client.patch(
+        url_for('api.update_newdle', code='dummy'),
+        **make_test_auth(dummy_uid),
+        json=json,
+    )
+    dummy_newdle_json['final_dt'] = final_dt
```
ThiefMaster: Why not just set it directly when defining the expected answer above? You never use `dummy_newdle_json` before in any way where you would need access to the original data.

tests/api_test.py
```diff
@@ -219,6 +255,74 @@ def test_get_newdle(flask_client, dummy_newdle):
...
+            '2019-08-12T13:30',
+        ],
+        'timezone': 'Europe/Paris',
+        'title': 'Test event1',
+        'url': 'http://flask.test/newdle/dummy1',
+    }
+    resp = flask_client.patch(
+        url_for('api.update_newdle', code='dummy'),
+        **make_test_auth(dummy_uid),
+        json=json,
```
ThiefMaster: I'm undecided on whether sending a lot of stuff that *cannot* be changed is a good idea or not... 

- on one hand, it ensures that we do ignore anything that cannot be changed
- on the other hand, the fact that sending more stuff than what can actually be updated is possible indicates that we should maybe rather configure the schema to reject that. unfortunately this doesn't seem to be easily possible: marshmallow-code/webargs#87

So I guess it's best to stick with what you have since that will make the test fail if more gets updated than what should be updateable..

tests/api_test.py
```diff
@@ -349,3 +468,98 @@ def test_update_participant_answers_valid_slots(flask_client):
...
+        url_for('api.create_unknown_participant', code='dummy'), json={'name': name},
+    )
+    assert resp.status_code == 403
+    assert resp.json == {'error': 'This newdle has finished'}
+
+
+@pytest.mark.usefixtures('dummy_newdle')
+def test_create_unknown_participant(flask_client):
+    name = 'Unknown participant'
+    nb_participant = Participant.query.count()
```
ThiefMaster: I'd call this `num_participants`. Matter of taste though, I just find "nb" for "number" a bit strange

tests/api_test.py
```diff
@@ -219,6 +255,74 @@ def test_get_newdle(flask_client, dummy_newdle):
...
+
+
+@pytest.mark.usefixtures('db_session')
+def test_update_newdle_unauthorized(flask_client, dummy_newdle):
+    resp = flask_client.patch(
+        url_for('api.update_newdle', code='dummy'),
+        json={'title': 'foo'},
+        **make_test_auth('someone'),
+    )
+    assert resp.status_code == Forbidden.code
```
ThiefMaster: I'd just hardcode `403` here since it's common knowledge that 403 is the http status code for Forbidden

tests/api_test.py
```diff
@@ -219,6 +255,74 @@ def test_get_newdle(flask_client, dummy_newdle):
...
+    assert resp.json == {'error': Forbidden.description}
+
+
+@pytest.mark.usefixtures('dummy_newdle')
+def test_update_newdle(flask_client, dummy_newdle, dummy_uid):
+    final_dt = '2019-09-12T13:30'
+    dummy_newdle_json = {
```
ThiefMaster: Maybe `expected_json` would be a better name for this?

tests/api_test.py
Biboba: I thought it would make it clearer that we only expect 'final_dt' property to be updated even if all the properties were updated when submitted. Happy to change it if necessary !

ThiefMaster: I'd change it ;)

Another option would be to get the newdle json before the change and after, pop the `final_dt` from both, and then assert that they are identical (and that the popped `final_dt` matches what you PATCHed)

ThiefMaster:
I had kind of forgotten about this PR (maybe GitHub didn't notify me about you updating it, not sure). Thanks for updating it based on my comments :)

PR #198, opened 2020-08-11
Make default time for timeslots smarter

### LINKED ISSUE #188: Time definition could be streamlined
Hi,

In addition to https://github.com/indico/newdle/issues/183 (esp the "bulk time slot" option) I'd like to suggest the following:

When entering the digits for a meeting time, the default value (for me) is `10:00 AM`.
- If I enter "15" into the hour field, it gets "autocorrected" to `03:00 AM`. It would be even better if it got autocorrected to `03:00 PM` :-)
- Once I've entered a slot at 15:00 - sorry, "03:00 PM" ;-) - it again shows `10:00 AM` as the default value for the next slot. I think it would be a big usability improvement if the next default would be end-of-the-previous-slot if that exists. I.e. for the last slot being 15:00-15:30, suggest 15:30 as default value for a new slot to be added.

Cheers, Axel.

Leats:
Default time for timeslots will now be either the end of the previous one or the first unused one of the previous date, depending on whether there are timeslots on the current day already.
Closes #188.

### DIFF
--- a/newdle/client/src/components/creation/timeslots/Timeline.js
+++ b/newdle/client/src/components/creation/timeslots/Timeline.js
@@ -8,6 +8,7 @@ import {
   getCreationCalendarActiveDate,
   getDuration,
   getTimeslotsForActiveDate,
+  getNewTimeslotStartTime,
 } from '../../../selectors';
 import CandidateSlot from './CandidateSlot';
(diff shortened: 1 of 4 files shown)


Leats:
Bildschirmfoto 2020-08-11 um 14 40 03
Added a screenshot of the new button!

PR #210, opened 2020-08-19
Make answer page responsive

### LINKED ISSUE #133: Make newdle answer page responsive
Probably very related to #131 

image

panagiotappl:
closes #133

### DIFF
--- a/newdle/client/src/components/answer/Calendar.js
+++ b/newdle/client/src/components/answer/Calendar.js
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 import {Grid} from 'semantic-ui-react';
 import {useDispatch, useSelector} from 'react-redux';
 import {hourRange, serializeDate, toMoment, getHourSpan} from '../../util/date';
+import {useIsSmallScreen} from '../../util/hooks';
 import DayTimeline from './DayTimeline';
 import {
   getActiveDate,
@@ -182,7 +183,9 @@ export default function Calendar() {
   const busyTimes = useSelector(getBusyTimes);
   const activeDate = toMoment(useSelector(getActiveDate), HTML5_FMT.DATE);
   const dispatch = useDispatch();
+
   const defaultHourSpan = MAX_HOUR - MIN_HOUR;
+  const isTabletOrMobile = useIsSmallScreen();
 
   if (timeSlots.length === 0) {
     return 'No data';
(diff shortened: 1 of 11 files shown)


newdle/client/src/components/answer/Calendar.js
```diff
@@ -182,7 +183,9 @@ export default function Calendar() {
   const busyTimes = useSelector(getBusyTimes);
   const activeDate = toMoment(useSelector(getActiveDate), HTML5_FMT.DATE);
   const dispatch = useDispatch();
+
   const defaultHourSpan = MAX_HOUR - MIN_HOUR;
+  const isTabletOrMobile = useMediaQuery({query: '(max-width: 1224px)'});
```
ThiefMaster, line 188: You use this in multiple places, so maybe wrap it in a custom `useIsSmallScreen` hook so we don't have the media query string in multiple places.

newdle/client/src/components/answer/Calendar.js
```diff
@@ -203,19 +206,21 @@ export default function Calendar() {
   const activeDateIndex = optionsByDay.findIndex(({date: timeSlotDate}) =>
     toMoment(timeSlotDate, HTML5_FMT.DATE).isSame(activeDate, 'day')
   );
-
+  const numberOfVisible = isTabletOrMobile ? 1 : 3;
```
ThiefMaster, line 209: that's a strange name :D

What about `numDaysVisible` and `numColumns`?

package.json
```diff
@@ -29,5 +29,9 @@
     "husky": "^4.2.5",
     "lint-staged": "^10.2.11",
     "prettier": "~1.18.2"
+  },
+  "dependencies": {
+    "rc-time-picker": "^3.7.3",
+    "react-responsive": "^8.1.0"
```
ThiefMaster, line 35: This shouldn't be here

newdle/client/src/components/answer/Calendar.js
ThiefMaster: and for the failing lint check you could just call the hook again there instead of passing it as a prop

newdle/client/src/components/answer/Calendar.js
panagiotappl: I used that because it's also used as prop in `DayCarousel` with that name

PR #216, opened 2020-08-25
Make ParticipantTable responsive

### LINKED ISSUE #134: Improve responsiveness of newdle summary page

panagiotappl:
closes #134

image

### DIFF
--- a/newdle/client/src/components/ParticipantTable.js
+++ b/newdle/client/src/components/ParticipantTable.js
@@ -5,6 +5,7 @@ import {useSelector} from 'react-redux';
 import {Radio, Icon, Label, Table} from 'semantic-ui-react';
 import AvailabilityRing from './AvailabilityRing';
 import {serializeDate, toMoment} from '../util/date';
+import {useIsMobile} from 'src/util/hooks';
 import {getNewdleDuration, getNumberOfParticipants, getParticipantAvailability} from '../selectors';
 import styles from './ParticipantTable.module.scss';
 
@@ -20,10 +21,6 @@ function formatMeetingTime(startTime, duration) {
 function ParticipantNames({participants}) {
   const [renderAll, setRenderAll] = useState(false);
   const statusColors = {available: 'green', ifneedbe: 'yellow', unavailable: 'red'};
-
-  if (participants.length === 0) {
-    return 'Nobody has voted yet.';
-  }
 
   // eslint-disable-next-line react/prop-types
   const renderName = ({name, status, id}) => (
@@ -75,6 +72,18 @@ function AvailabilityRow({
   const duration = useSelector(getNewdleDuration);
   const startTime = toMoment(startDt, 'YYYY-MM-DDTHH:mm');
 
+  const isMobile = useIsMobile();
+  const renderParticipants = () => {
+    if (!numberOfParticipants) {
+      return 'There are no participants yet.';
+    } else if (participants.length === 0) {
+      return 'Nobody has voted yet.';
+    } else if (!isMobile || active) {
+      return <ParticipantNames participants={participants} />;
+    }
+    return null;
+  };
+
   return (
     <Table.Row
       className={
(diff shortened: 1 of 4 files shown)


newdle/client/src/components/ParticipantTable.js
```diff
@@ -83,41 +110,55 @@ function AvailabilityRow({
...
       style={!finalized || active ? null : {opacity: '0.3'}}
-      onClick={() => (finalized || !isCreator ? null : setActiveDate(startDt))}
+      onClick={() => (finalized || (!isMobile && !isCreator) ? null : setActiveDate(startDt))}
       active={active}
     >
       <Table.Cell width={3}>
-        <div className={styles['date']}>{startTime.format('D MMM')}</div>
-        <div className={styles['time']}>{formatMeetingTime(startTime, duration)}</div>
+        {isMobile ? (
+          <div style={{display: 'flex', justifyContent: 'space-evenly', alignItems: 'center'}}>
```
ThiefMaster, line 118: this could go in the scss file

newdle/client/src/components/ParticipantTable.module.scss
```diff
@@ -2,7 +2,9 @@
   max-width: 700px;
 
   .participant-row {
-    height: 150px;
+    height: 100%;
+    padding-top: 0em;
+    padding-bottom: 0em;
```
ThiefMaster, line 7:
```suggestion
    padding-top: 0;
    padding-bottom: 0;
```

newdle/client/src/components/ParticipantTable.module.scss
```diff
@@ -70,3 +73,27 @@
     }
   }
 }
+
+@media only screen and (max-width: 767px) {
+  .participant-table {
```
ThiefMaster, line 78: do we need this selector? everything is inside `.participant-row` anyway

panagiotappl: Without it the `margin-bottom: 0` wasn't being applied as the value was taken from the normal css.. but this margin might not even be needed now with the latest changes in layout.

panagiotappl: Ah, but also other CSS seems to not get applied on mobile mode as well (flex-box related stuff)

ThiefMaster: ah the joys of selector specificity :/

ThiefMaster:
image

- Can we move both the circle and the participants more to the left, and maybe align them a bit so they start on the same horizontal position? Or if the header with the circle is supposed to be centered, just moving the participants a bit more to the left may be enough.
- The question mark icon for people who haven't voted yet looks a bit weird on top. Maybe we can always keep this on the side?

ThiefMaster:
BTW, I just tested on my actual phone and the new responsive UI is already really nice!

newdle/client/src/components/ParticipantTable.js
```diff
@@ -75,6 +72,36 @@ function AvailabilityRow({
...
+  const renderParticipants = () => {
+    if (!numberOfParticipants) {
+      return 'There are no participants yet.';
+    }
+    if (participants.length === 0) {
+      return 'Nobody has voted yet.';
+    }
+    if (availableCount === 0) {
+      return 'Nobody is available.';
+    }
```
ThiefMaster, line 85: This isn't a good idea because it prevents you from seeing *who* "nobody" actually is.

newdle/client/src/components/ParticipantTable.js
```diff
@@ -75,6 +72,36 @@ function AvailabilityRow({
...
+    if (!isMobile || active) {
+      return <ParticipantNames participants={participants} />;
+    }
+    return null;
+  };
+
+  const renderParticipantLabel = () => {
+    if (availableCount > 0 && (!isMobile || !active || finalized)) {
```
ThiefMaster, line 93: Together with my comment above I would remove the `availableCount` condition here and always show a label in the mobile view, but use a red one with a different icon in case nobody is available.

ThiefMaster: Also, better to structure it like this:

```js
if (...) {
  return null;
}
return ...;
```

newdle/client/src/components/ParticipantTable.js
panagiotappl: Why does it prevent you from seeing that? You can see it in the message underneath, right?

I added this mainly because there was an awkward empty white space under the grey area (of the circle + date). Now that everything is in one color it can also be removed.

ThiefMaster: But this is for the case where you have named participants, they are just all "red".

panagiotappl: Ah, true. I mixed it up with "Nobody has voted yet"

newdle/client/src/components/ParticipantTable.js
panagiotappl: If we always want to show a label, then the if condition can be removed. Isn't that what you meant?

ThiefMaster (approved):

newdle/client/src/components/ParticipantTable.js
ThiefMaster: I played around a bit with the label and honestly I'm not sure when it is actually useful... we always have the circle which already shows this information. But I'd leave it as it is for now, this can be revisited later if it turns out to be wasting space or people dislike it...

Issue #208, opened 2020-08-19
Display options in local timezone


hannahshort:
It would be helpful for international recipients to see the Newdle slot options displayed in their local timezone.

---

It looks like we can nowadays get the current timezone from the browser (we do not have a "user profile" yet where such a thing could be stored, and people may answer as guests anyway) via [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), so ideally we'd default to that timezone, but the user should be able to change it (this could also be done in a second step).

It should also be clear to the user that they are seeing it in their own timezone, maybe with an option to view it in the default timezone as chosen by the newdle creator.

An important case to take into account is that the start *date* may be different in the user's timezone - we have to be careful that we still submit it correctly for the original slot (ie time zone conversion should be purely a display thing).

Slightly tricky part: `getCalendarDates` (possibly other selectors as well, but on a quick glance I only found this one - best to eventually test it) are used both to fetch free-busy information from the API (always needs to use the raw value without timezone conversion) and to display the dates in the calendar (which would need to be adjusted using the user's timezone).

ThiefMaster:
related: #127

Diego-Zulu:
Hi there! I would like to work on this issue

I've been going through the code, and I believe I should be able to do it by refactoring the functions that end up applying the `.format` function on a moment object. If I apply the timezone to the moment object before formatting, I should be able to get the correct way of printing the dates

I'll reach out if I have any questions or problems!

ThiefMaster:
Great! 👍 
In this task we only need it on the main "answer newdle" page, i.e. the page on `/newdle/XXX` ([example](https://newdle.cern.ch/newdle/jJvWYRSm)). Adding it on the summary page would be a good second step (separate issue+pr) especially now that the summary is also visible to participants, but no need to add more work to this particular issue!

I think we don't send the newdle's timezone back to the client yet, so you'll have to include it when returning the newdle data from the API.

For the client-side implementation my approach would probably involve selectors and combining selectors (createSelector) (so you get the data from one selector, and the newdle's timezone from another selector, and then doing some "timezone magic" in your new selector).

I would in fact start with these two things so newdle can show the answer page and timeslots in the client's timezone.
Even if submission (or showing previously-selected values in case of someone already having answered) likely doesn't work yet, that's probably already a good point to open a draft PR.

PR #218, opened 2020-08-31
Displaying timezone on answers page

Diego-Zulu:
## Preview:

Screen Shot 2020-09-08 at 21 54 43
Screen Shot 2020-09-02 at 23 37 10

## Description:

Adding timezone functionality to the newdle answer page as mentioned on #208 . The display timezone used by default is the one obtained from the user's browser. Dates are kept both in the format received by the server and in a timezone sensitive format to show the user

## Todo:

- Active date marked on calendar seems to ignore user timezone, need to look more into it

## Checks:

- All tests pass ✅ 
- All linting rules pass ✅

## Manual testing:

Environment setup:

```
make
source ./.venv/bin/activate
make flask-server
make react-server
```

- Newdle with same timezone as user's has it's dates displayed without changes ✅
- Newdle with option 17:00 - 17:30 on CEST displays as 12:00 - 12:30 on Sao Paulo time ✅
- Newdle displays date correctly with 12 hour difference (e.g. 10:00 August 31st on Japan should be 22:00 August 30th on Sao Paulo) ✅
- Same 12 hour difference check for busy times ✅
- Active date on calendar matches event date correctly when 12 hour difference ✅
- Active date on calendar matches event date correctly when 12 hour difference ✅
- "Accept all options where I'm available" toggle highlights only the aforementioned slots ✅

## Questions:

N/A

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -205,6 +205,7 @@ def _get_busy_times(date, tz, uid):
         [
             ['{:02}:{:02}'.format(*r[0]), '{:02}:{:02}'.format(*r[1])]
             for r in merged_ranges
+            if r[0] != r[1]
         ]
     )
 
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -6,13 +6,19 @@ from flask import Blueprint, current_app, g, jsonify, request, url_for
 from itsdangerous import BadData, SignatureExpired
 from marshmallow import fields
 from marshmallow.validate import OneOf
-from pytz import common_timezones_set
+from pytz import common_timezones_set, timezone
 from sqlalchemy.orm import selectinload
 from werkzeug.exceptions import Forbidden, ServiceUnavailable, UnprocessableEntity
 
 from .core.auth import search_users, user_info_from_app_token
 from .core.db import db
-from .core.util import DATE_FORMAT, format_dt, range_union, sign_user
+from .core.util import (
+    DATE_FORMAT,
+    change_dt_timezone,
+    format_dt,
+    range_union,
+    sign_user,
+)
 from .core.webargs import abort, use_args, use_kwargs
 from .models import Newdle, Participant
 from .notifications import notify_newdle_participants
@@ -187,7 +193,12 @@ def get_participant_busy_times(date, code, tz, participant_code=None):
     ).first_or_404('Specified participant does not exist')
     if participant.auth_uid is None:
         abort(422, messages={'participant_code': ['Participant is an unknown user']})
-    if not any(date == ts.date() for ts in participant.newdle.timeslots):
+    targetTz = timezone(tz)
+    newdleTz = timezone(participant.newdle.timezone)
+    if not any(
+        date == change_dt_timezone(ts, newdleTz, targetTz).date()
+        for ts in participant.newdle.timeslots
+    ):
         abort(422, messages={'date': ['Date has no timeslots']})
     return _get_busy_times(date, tz, participant.auth_uid)
 
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -193,10 +193,10 @@ def get_participant_busy_times(date, code, tz, participant_code=None):
     ).first_or_404('Specified participant does not exist')
     if participant.auth_uid is None:
         abort(422, messages={'participant_code': ['Participant is an unknown user']})
-    targetTz = timezone(tz)
-    newdleTz = timezone(participant.newdle.timezone)
+    target_tz = timezone(tz)
+    newdle_tz = timezone(participant.newdle.timezone)
     if not any(
-        date == change_dt_timezone(ts, newdleTz, targetTz).date()
+        date == change_dt_timezone(ts, newdle_tz, target_tz).date()
         for ts in participant.newdle.timeslots
     ):
         abort(422, messages={'date': ['Date has no timeslots']})
--- a/newdle/client/src/components/ParticipantTable.js
+++ b/newdle/client/src/components/ParticipantTable.js
@@ -6,7 +6,12 @@ import {Radio, Icon, Label, Table} from 'semantic-ui-react';
 import AvailabilityRing from './AvailabilityRing';
 import {serializeDate, toMoment} from '../util/date';
 import {useIsMobile} from 'src/util/hooks';
-import {getNewdleDuration, getNumberOfParticipants, getParticipantAvailability} from '../selectors';
+import {
+  getNewdleDuration,
+  getNewdleTimezone,
+  getNumberOfParticipants,
+  getParticipantAvailability,
+} from '../selectors';
 import styles from './ParticipantTable.module.scss';
 
 const MAX_PARTICIPANTS_SHOWN = 4;
@@ -69,6 +74,7 @@ function AvailabilityRow({
   children,
 }) {
   const numberOfParticipants = useSelector(getNumberOfParticipants);
+  const newdleTimezone = useSelector(getNewdleTimezone);
   const duration = useSelector(getNewdleDuration);
   const startTime = toMoment(startDt, 'YYYY-MM-DDTHH:mm');
 
@@ -109,6 +115,7 @@ function AvailabilityRow({
             <div>
               <div className={styles['date']}>{startTime.format('D MMM')}</div>
               <div className={styles['time']}>{formatMeetingTime(startTime, duration)}</div>
+              <div className={styles['timezone']}>{newdleTimezone}</div>
             </div>
             {!finalized && isCreator && <Radio name="slot-id" value={startDt} checked={active} />}
           </div>
@@ -116,6 +123,7 @@ function AvailabilityRow({
           <>
             <div className={styles['date']}>{startTime.format('D MMM')}</div>
             <div className={styles['time']}>{formatMeetingTime(startTime, duration)}</div>
+            <div className={styles['timezone']}>({newdleTimezone})</div>
           </>
         )}
       </Table.Cell>
--- a/newdle/client/src/components/answer/AnswerPage.js
+++ b/newdle/client/src/components/answer/AnswerPage.js
@@ -150,9 +150,10 @@ export default function AnswerPage() {
 
   useEffect(() => {
     if ((participantCode && !participantUnknown) || (!participantCode && user)) {
-      dispatch(fetchBusyTimesForAnswer(newdleCode, participantCode || null, dates, newdleTz));
+      // Fetching busy times for user's timezone so no timezone conversion needed later
+      dispatch(fetchBusyTimesForAnswer(newdleCode, participantCode || null, dates, userTz));
     }
-  }, [dates, newdleCode, participantCode, participantUnknown, user, newdleTz, dispatch]);
+  }, [dates, newdleCode, participantCode, participantUnknown, user, userTz, dispatch]);
 
   if (!newdle || (participantCode && !participant)) {
     return null;
--- a/newdle/client/src/components/answer/AnswerPage.js
+++ b/newdle/client/src/components/answer/AnswerPage.js
@@ -101,7 +101,7 @@ export default function AnswerPage() {
   const participantUnknown = useSelector(isParticipantUnknown);
   const participantAnswersChanged = useSelector(haveParticipantAnswersChanged);
   const busyTimesLoaded = useSelector(hasBusyTimes);
-  const tz = useSelector(getNewdleTimezone);
+  const newdleTz = useSelector(getNewdleTimezone);
   const userTz = useSelector(getUserTimezone);
   usePageTitle(newdle && newdle.title, true);
 
@@ -150,9 +150,9 @@ export default function AnswerPage() {
 
   useEffect(() => {
     if ((participantCode && !participantUnknown) || (!participantCode && user)) {
-      dispatch(fetchBusyTimesForAnswer(newdleCode, participantCode || null, dates, tz));
+      dispatch(fetchBusyTimesForAnswer(newdleCode, participantCode || null, dates, newdleTz));
     }
-  }, [dates, newdleCode, participantCode, participantUnknown, user, tz, dispatch]);
+  }, [dates, newdleCode, participantCode, participantUnknown, user, newdleTz, dispatch]);
 
   if (!newdle || (participantCode && !participant)) {
     return null;
@@ -211,7 +211,10 @@ export default function AnswerPage() {
                 />
               </Segment>
             )}
-            <Header as="h3" className={styles.date}>
+            <Header as="h4" className={styles.date}>
+              Newdle timezone: {newdleTz}
+            </Header>
+            <Header as="h4" className={styles.date}>
               Displayed timezone: {userTz}
             </Header>
           </Grid.Column>
--- a/newdle/client/src/components/answer/AnswerPage.js
+++ b/newdle/client/src/components/answer/AnswerPage.js
@@ -1,14 +1,4 @@
-import {
-  Button,
-  Checkbox,
-  Container,
-  Grid,
-  Icon,
-  Input,
-  Message,
-  Segment,
-  Header,
-} from 'semantic-ui-react';
+import {Button, Checkbox, Container, Grid, Icon, Input, Message, Segment} from 'semantic-ui-react';
 import React, {useEffect, useState} from 'react';
 import {useDispatch, useSelector} from 'react-redux';
 import {useParams} from 'react-router-dom';
@@ -212,12 +202,12 @@ export default function AnswerPage() {
                 />
               </Segment>
             )}
-            <Header as="h4" className={styles.date}>
-              Newdle timezone: {newdleTz}
-            </Header>
-            <Header as="h4" className={styles.date}>
-              Displayed timezone: {userTz}
-            </Header>
+            <div className={styles.timezone}>
+              <span className={styles['timezone-title']}>Newdle timezone:</span> {newdleTz}
+            </div>
+            <div className={styles.timezone}>
+              <span className={styles['timezone-title']}>Displayed timezone:</span> {userTz}
+            </div>
           </Grid.Column>
           <Grid.Column computer={11} tablet={8}>
             <Calendar />
--- a/newdle/client/src/components/answer/AnswerPage.js
+++ b/newdle/client/src/components/answer/AnswerPage.js
@@ -1,4 +1,14 @@
-import {Button, Checkbox, Container, Grid, Icon, Input, Message, Segment} from 'semantic-ui-react';
+import {
+  Button,
+  Checkbox,
+  Container,
+  Grid,
+  Icon,
+  Input,
+  Message,
+  Segment,
+  Header,
+} from 'semantic-ui-react';
 import React, {useEffect, useState} from 'react';
 import {useDispatch, useSelector} from 'react-redux';
 import {useParams} from 'react-router-dom';
@@ -21,7 +31,7 @@ import {
   haveParticipantAnswersChanged,
   hasBusyTimes,
 } from '../../answerSelectors';
-import {getUserInfo} from '../../selectors';
+import {getUserInfo, getUserTimezone} from '../../selectors';
 import {
   chooseAllAvailable,
   fetchBusyTimesForAnswer,
@@ -92,6 +102,7 @@ export default function AnswerPage() {
   const participantAnswersChanged = useSelector(haveParticipantAnswersChanged);
   const busyTimesLoaded = useSelector(hasBusyTimes);
   const tz = useSelector(getNewdleTimezone);
+  const userTz = useSelector(getUserTimezone);
   usePageTitle(newdle && newdle.title, true);
 
   const [submitAnswer, submitting, , submitResult] = participantCode
@@ -200,6 +211,9 @@ export default function AnswerPage() {
                 />
               </Segment>
             )}
+            <Header as="h3" className={styles.date}>
+              Displayed timezone: {userTz}
+            </Header>
           </Grid.Column>
           <Grid.Column computer={11} tablet={8}>
             <Calendar />
--- a/newdle/client/src/components/answer/Calendar.js
+++ b/newdle/client/src/components/answer/Calendar.js
@@ -130,7 +130,7 @@ function calculateOptionsPositions(options, duration, minHour, maxHour, answers,
     options.map(slot =>
       getSlotProps(slot, duration, minHour, maxHour, answers[slot], newdleTz, userTz)
     ),
-    slot => serializeDate(toMoment(slot.slot, DEFAULT_FORMAT))
+    slot => serializeDate(slot.groupDateKey, HTML5_FMT.DATE, userTz)
   );
 
   return Object.entries(optionsByDate).map(([date, options]) => {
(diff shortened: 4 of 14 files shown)


requirements.txt
```diff
@@ -12,7 +12,7 @@ Jinja2==2.10.3
 marshmallow-enum==1.5.1
 marshmallow-sqlalchemy==0.19.0
 marshmallow==3.2.1
-psycopg2==2.8.4
+psycopg2-binary==2.8.4
```
Diego-Zulu, line 15: I wanted to leave this change in for a discussion: Do we need to compile psycopg2 when installing dependencies? If not, we should start using `psycopg2-binary`

newdle/core/app.py
```diff
@@ -86,7 +86,8 @@ def create_app(config_override=None, use_env_config=True):
     _configure_app(app, from_env=use_env_config)
     if config_override:
         app.config.update(config_override)
-    _configure_multipass(app)
+    if not app.config['SKIP_LOGIN']:
```
Diego-Zulu, line 89: SKIP_LOGIN was failing because `_configure_multipass` redirects were in effect, so had to add this to be able to use this option

newdle/client/src/util/date.js
```diff
@@ -2,28 +2,56 @@ import moment, {HTML5_FMT} from 'moment';
...
 export function overlaps([start1, end1], [start2, end2]) {
   return start1.isBefore(end2) && start2.isBefore(end1);
 }
 
 export function getHourSpan(input) {
-  const {timeSlots, defaultHourSpan, defaultMinHour, defaultMaxHour, duration, format} = input;
-  const timeSlotsMoment = timeSlots.map(c => toMoment(c, format));
+  const {
+    timeSlots,
+    busyTimes,
```
Diego-Zulu, line 20: I belive busyTimes were not being considered for the answers page, so we wouldn't see them on the answers page even though react would try to renderize them. We should either: 
1 - Include them correctly on the display between a correct min and max hours (this approach)
2 - Don't show busy times that are outside min and max hours range

newdle/client/src/components/creation/timeslots/Timeline.js
```diff
@@ -71,7 +71,7 @@ function getCandidateSlotProps(startTime, duration, minHour, maxHour) {
   const endTime = toMoment(startTime, DEFAULT_TIME_FORMAT)
     .add(duration, 'm')
     .format(DEFAULT_TIME_FORMAT);
-  return getSlotProps(startTime, endTime, minHour, maxHour, duration);
+  return getSlotProps(startTime, endTime, minHour, maxHour);
```
Diego-Zulu, line 74: I believe this `duration` argument was not being captured by the function so I removed it

newdle/auth.py
```diff
@@ -17,4 +17,8 @@ def login():
 
 @auth.route('/logout/')
 def logout():
-    return multipass.logout(url_for('index', _external=True), clear_session=True)
+    after_logout = url_for('index', _external=True)
```
ThiefMaster, line 20: `after_logout_url` would be a better name here

newdle/client/src/selectors.js
```diff
@@ -15,6 +15,7 @@ export const isAcquiringToken = state => !!state.auth.acquiringToken;
 
 // user
 export const getUserInfo = state => state.user;
+export const getUserTimezone = state => (state.user && state.user.timezone) || moment.tz.guess();
```
ThiefMaster, line 18: Since we never have a user timezone in the state, i would only return the guessed one.

We can add it back if that changes at some point in the future.

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -200,6 +211,9 @@ export default function AnswerPage() {
                 />
               </Segment>
             )}
+            <Header as="h3" className={styles.date}>
+              Displayed timezone: {userTz}
```
ThiefMaster, line 215: I think it'd be important to show the newdle's timezone as well, so the user knows that the event the newdle is about is taking place in timezone X, but the times are shown to them in timezone Y.

newdle/client/src/util/date.js
ThiefMaster: I don't think we care much about busy times outside the displayed timespan here. So I'd just hide those.

newdle/client/src/components/creation/timeslots/Timeline.js
ThiefMaster: Indeed, thanks!

newdle/client/src/util/date.js
```diff
@@ -2,28 +2,56 @@ import moment, {HTML5_FMT} from 'moment';
...
+    defaultHourSpan,
+    defaultMinHour,
+    defaultMaxHour,
+    duration,
+    format,
+    newdleTz,
+    userTz,
+  } = input;
+  let timeSlotsMoment = [];
+  if (userTz && newdleTz) {
```
ThiefMaster, line 30: I don't think we can have any case with no `newdleTz` being set.

Actually, can `userTz` even be empty? I don't see anything about it in [the docs](https://momentjs.com/timezone/docs/#/using-timezones/guessing-user-timezone/).

So `userTz !== newdleTz` may be a better check here. But then I have the feeling that the logic here would simply not change anything, so maybe we don't need this check since we can always run the code here?

ThiefMaster:
> Regarding busy times: I understand that (because I'm using the skip sign in option) I'm getting random busy times. I'm I correct to assume that the returned busy times' timezone is on the same timezone we provide as a parameter?

Yes. Here's an example with real data:

```
$ http -pb get 'https://newdle.blackhole.cern.ch/api/newdle/7kYbqYp9/participants/xxx/busy' 'date==2020-09-09' 'tz==UTC'
[
    [
        "08:30",
        "10:00"
    ],
    [
        "14:00",
        "15:00"
    ],
    [
        "22:00",
        "23:59"
    ]
]

$ http -pb get 'https://newdle.blackhole.cern.ch/api/newdle/7kYbqYp9/participants/xxx/busy' 'date==2020-09-09' 'tz==Europe/Zurich'
[
    [
        "10:30",
        "12:00"
    ],
    [
        "16:00",
        "17:00"
    ]
]
```

This particular user has the following busy times (in Europe/Zurich):

- Sep 9: 10:30-12:00, 16:00-17:00
- Sep 10: Full day (this is what causes the 22:00-23:59 entry for the 9th in ?UTC)

newdle/client/src/util/date.js
ThiefMaster: BTW, this indeed looks a bit weird (in master):

image

```
$ http -pb get 'https://newdle.blackhole.cern.ch/api/newdle/7kYbqYp9/participants/xprFxn2f/busy' 'date==2020-09-11' 'tz==Europe/Zurich'
[
    [
        "00:00",
        "00:00"
    ],
    [
        "14:00",
        "16:00"
    ]
]
```

However, I think this is actually a bug in the API. We should not return a zero-duration busy slot at all (this is coming from the all-day-busy on the day before).

newdle/client/src/components/answer/Calendar.js
```diff
@@ -97,17 +99,28 @@ function getAnswerProps(slot, answer) {
...
 
-function getSlotProps(slot, duration, minHour, maxHour, answer) {
+function getSlotProps(slot, duration, minHour, maxHour, answer, newdleTz, userTz) {
   const start = toMoment(slot, DEFAULT_FORMAT);
   const end = toMoment(start).add(duration, 'm');
   const answerProps = getAnswerProps(slot, answer);
   const height = calculateHeight(start, end, minHour, maxHour);
-  const pos = calculatePosition(start, minHour, maxHour);
+
+  let startMomentForTimezone = toMoment(slot, DEFAULT_FORMAT, newdleTz).tz(userTz);
```
ThiefMaster, line 108: `...ForTimezone` sounds a bit strange. In Indico we usually use "local" to indicate times in a local timezone, e.g. `startMomentLocal`. Or `startMomentUserTz` / `startMomentUser`? `...Local` sounds least strange to me though.

Anyway, naming discussions are not really important right now (and they tend to be harder than the actual technical discussions, because naming is a [hard problem](https://twitter.com/secretGeek/status/7269997868))

newdle/client/src/util/date.js
ThiefMaster: BTW, taking busy times into account here in any regard would probably be something for a separate issue/PR. Especially when it comes to timezone logic which is already tricky by itself, it does not make PRs easier to review and test if they include other, unrelated things as well. 

But nonetheless, thanks for realizing that this is a point which should probably be improved!

ThiefMaster:
I was having another look at the free-busy timezone behavior and since the API returns it in the specified timezone,. you do not need to perform any client-side timezone conversions there - you just need to pass the `userTz` instead of the `newdleTz` when fetching fetching the free-busy information from the server. This means some of the new logic you added can actually be simplified a bit (and some of it maybe even removed again, such as the code added to `getBusySlotProps`). :)

ThiefMaster:
FYI, I just opened #219 which correctly handles timezone information in the random free-busy data the API returns. I will merge this in a few minutes once CI passed, so once you rebased you should be able to test properly with the random data.

requirements.txt
Diego-Zulu: Sounds good! I did not know psycopg2-binary was not recommended for prod environment. I'll remove the changes on this requirements.txt 👍

newdle/client/src/util/date.js
Diego-Zulu: Ok, I agree, I'll leave it out for now and create a separate issue

newdle/auth.py
```diff
@@ -17,4 +17,8 @@ def login():
 
 @auth.route('/logout/')
 def logout():
-    return multipass.logout(url_for('index', _external=True), clear_session=True)
```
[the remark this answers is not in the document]
Diego-Zulu: Ok, I'll create a separate issue and PR, so I can move this code and the redirect problems when skipping login

requirements.txt
ThiefMaster: It's mentioned in its [readme on pypi](https://pypi.org/project/psycopg2-binary/) ;)

> The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

<details>
FWIW, they may be a bit overly cautious - in older psycopg2 versions binary wheels were the default so many people ended up using those implicitly in the normal psycopg2 package.
</details>

newdle/client/src/selectors.js
Diego-Zulu: Sounds good, I did it just to show how we could receive it in the future, but I agree "code speculation" is confusing. I'll leave only moment.tz.guess

newdle/client/src/components/answer/AnswerPage.js
Diego-Zulu: Good idea, thanks!

newdle/auth.py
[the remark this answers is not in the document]
ThiefMaster: No need for an issue, PR is enough for something as simple as this. :)

newdle/client/src/components/answer/Calendar.js
Diego-Zulu: Ok, I'll revisit the naming. I think startTimeLocal should work 👍

Diego-Zulu:
Ok, thanks for the comments and for the random provider free-busy data fix! I'll get to all of the changes mentioned above later today

newdle/client/src/util/date.js
Diego-Zulu: I forgot to address this comment. The problem here is that this same method is used during creation, so it won't receive any userTz or newdleTz. I'll revisit this to see if I can have a cleaner logic for both the creation and the answer pages

newdle/client/src/components/answer/AnswerPage.js
Diego-Zulu: Added a simple "Newdle timezone" header, will update the PR's preview image ASAP

newdle/client/src/components/answer/Calendar.js
Diego-Zulu: Immplemented startTimeLocal and startMomentLocal

newdle/core/app.py
Diego-Zulu: Moved this code to https://github.com/indico/newdle/pull/220

newdle/auth.py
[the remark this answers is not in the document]
Diego-Zulu: Moved to https://github.com/indico/newdle/pull/220

newdle/auth.py
Diego-Zulu: Done on https://github.com/indico/newdle/pull/220

newdle/client/src/util/date.js
Diego-Zulu: I noticed that I could just move the timezone for the outputted hours from the original `getHourSpan` method instead of modifying it

newdle/client/src/util/date.js
Diego-Zulu: Hey, I realized this discussion all steamed from me not understanding that busy slots below minHour will be assigned pos 0 and no height. I believed the minHour should be the minimum of the busy slots and options slots just because I didn't understand that I was rendering the busy slots wrongly. I now believe my timezone aware busySlots implementation is correct and they render correctly (pos and height wise), so I believe this discussion can be dropped

Diego-Zulu: Although I can open a new issue if we want to fix the red line that appears at the top when the busy slot is outside our display range

ThiefMaster: i think they are fine - to indicate that there is *something* earlier, without cluttering the vertical space.

(i'd just fix the 00:00-00:00 entries sent e.g. by the exchange backend, because those are useless since they come from the previous day)

newdle/client/src/components/answer/Option.js
```diff
@@ -25,6 +28,8 @@ Option.propTypes = {
   className: PropTypes.string,
   icon: PropTypes.string.isRequired,
   styles: PropTypes.object,
+  startTimeLocal: PropTypes.string.isRequired,
+  endTimeLocal: PropTypes.string.isRequired,
```
ThiefMaster, line 32: `startTime` and `endTime` are no longer used so I would either:

- remove them from `propTypes` as well or
- keep using the original `startTime` and `endTime` props since AFAICT nothing changed inside this component besides the removal of the serialize+toMoment calls. BTW, the only reason these calls exist is to strip leading zeroes from hours before 10am, and that logic should probably be preserved since it's more human-friendly

newdle/client/src/components/answer/Calendar.js
```diff
@@ -116,33 +127,55 @@ function getSlotProps(slot, duration, minHour, maxHour, answer) {
...
+    );
+  });
+
+  const busyByDate = _.groupBy(busySlots, slot =>
+    serializeDate(slot.groupDateKey, HTML5_FMT.DATE, userTz)
+  );
+
+  return Object.entries(busyByDate).map(([date, times]) => {
+    return {date, times};
   });
```
ThiefMaster, line 179: I think you don't need any of the changes here if you make this change instead:

```patch
diff --git a/newdle/client/src/components/answer/AnswerPage.js b/newdle/client/src/components/answer/AnswerPage.js
index c25342d..87b7212 100644
--- a/newdle/client/src/components/answer/AnswerPage.js
+++ b/newdle/client/src/components/answer/AnswerPage.js
@@ -150,9 +150,9 @@ export default function AnswerPage() {

   useEffect(() => {
     if ((participantCode && !participantUnknown) || (!participantCode && user)) {
-      dispatch(fetchBusyTimesForAnswer(newdleCode, participantCode || null, dates, newdleTz));
+      dispatch(fetchBusyTimesForAnswer(newdleCode, participantCode || null, dates, userTz));
     }
-  }, [dates, newdleCode, participantCode, participantUnknown, user, newdleTz, dispatch]);
+  }, [dates, newdleCode, participantCode, participantUnknown, user, userTz, dispatch]);

   if (!newdle || (participantCode && !participant)) {
     return null;
```

(you probably need to still change some params or undo changes related to the new props you added )

newdle/client/src/util/date.js
Diego-Zulu: I think I fixed this 00:00-00:00 shortcoming, please check it out

newdle/client/src/components/answer/Option.js
Diego-Zulu: You are right! I believed startTime and endTime were still being used as part of posting the availability to the backend, but then I noticed the slot has its own action already associated for that purpose. I deleted `startTimeLocal` and `endTimeLocal`, and went back to `startTime` and `endTime`

newdle/client/src/selectors.js
```diff
@@ -15,6 +15,7 @@ export const isAcquiringToken = state => !!state.auth.acquiringToken;
 
 // user
 export const getUserInfo = state => state.user;
+export const getUserTimezone = _state => moment.tz.guess();
```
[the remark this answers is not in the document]
Diego-Zulu: I'll leave it as `export const getUserTimezone = () => moment.tz.guess();`. In the original issue it was discussed that it would be nice to allow the user to change this timezone, and I agree. So I want to leave it like this so a future maintainer doesn't need to worry about deprecating and removing function on `util/date.js`

newdle/client/src/components/answer/Calendar.js
Diego-Zulu: You were totally right, so much code simplified, thanks for that!

newdle/client/src/selectors.js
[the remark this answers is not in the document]
ThiefMaster: good point! i wasn't thinking about that case but it makes sense to store the custom timezone in the state so keeping the selector is indeed a good idea :)

newdle/api.py
```diff
@@ -187,7 +193,12 @@ def get_participant_busy_times(date, code, tz, participant_code=None):
     ).first_or_404('Specified participant does not exist')
     if participant.auth_uid is None:
         abort(422, messages={'participant_code': ['Participant is an unknown user']})
-    if not any(date == ts.date() for ts in participant.newdle.timeslots):
+    targetTz = timezone(tz)
+    newdleTz = timezone(participant.newdle.timezone)
+    if not any(
+        date == change_dt_timezone(ts, newdleTz, targetTz).date()
+        for ts in participant.newdle.timeslots
+    ):
```
Diego-Zulu, line 201: Hey, let's discuss this piece of code. You see, when updating the active date code to correctly reflect the active date on the answer calendar when dealing with 12 hour diff timezones (japan - sao paulo for example), I noticed a problem here. My test case is:

- Newdle on Sept 4th 10AM Japan
- I use Sao Paulo, so I see displayed Sept 3rd 10PM

In this case, I end up asking my busy slots for Sept 3rd, and this check would fail because all slots were for Sept 4th. The solutions I thought about were:

1. This solution I wrote: Cast the slots timezone before asking for the date. Depending on the demand of this particular endpoint, this could be a costly operation
2. Ask for busy time on the timezone and date of the newdle, and then cast and aggregate them on the frontend as I did in a previous commit we discussed
3. Remove this check and call `_get_busy_times` even if there are no `participant.newdle.timeslots`

What do you think we should do?

ThiefMaster: I like your solution. It's simple, and timezone conversions are so cheap that they are perfectly fine. Every database query etc is likely more expensive than even a few timezone conversions.

Just one thing: We do not use camelCase in Python code - please use the more common `snake_case` in the code above.

newdle/client/src/components/answer/Option.js
```diff
@@ -1,17 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {Icon} from 'semantic-ui-react';
-import {serializeDate, toMoment} from '../../util/date';
 import styles from './answer.module.scss';
 
 export default function Option({startTime, endTime, icon, onClick, className, styles: moreStyles}) {
-  const start = serializeDate(toMoment(startTime, 'H:mm'), 'H:mm');
-  const end = serializeDate(toMoment(endTime, 'H:mm'), 'H:mm');
```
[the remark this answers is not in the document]
Diego-Zulu: Ok 👍 . I just fixed this, I decided to just serialize `startTime` at `getSlotProps` with single-digit hours there, as we aren't using it anywhere else

newdle/api.py
Diego-Zulu: Ok, sounds good! And I just fixed the camelCase on Python, I was pretty tired when I wrote that code and the camelCase totally passed me by. Thanks for the catch!

Diego-Zulu:
Hey, I'm removing the Draft status form this PR. I have run over a couple examples myself and I all of them rendered correctly, unbeknownst to the timezone difference

Please feel free to continue commenting until you think things look good enough to merge! I'll continue addressing the comments ASAP

newdle/client/src/components/ParticipantTable.js
```diff
@@ -109,13 +115,15 @@ function AvailabilityRow({
             <div>
               <div className={styles['date']}>{startTime.format('D MMM')}</div>
               <div className={styles['time']}>{formatMeetingTime(startTime, duration)}</div>
+              <div className={styles['timezone']}>{newdleTimezone}</div>
```
ThiefMaster, line 118: It looks good, but I wonder if it wouldn't make more sense to display the timezone in a single place instead of for every single timeslot (since it's always the same anyway).

Not a blocker for this PR though!

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -200,6 +212,12 @@ export default function AnswerPage() {
                 />
               </Segment>
             )}
+            <Header as="h4" className={styles.date}>
+              Newdle timezone: {newdleTz}
+            </Header>
+            <Header as="h4" className={styles.date}>
```
ThiefMaster, line 218: Should we display this in case it's the same as `newdleTz`? And I'm not sure if `Header` (especially when rendered as a h4 or any other hX element) is semantically correct - it's not really a header..

ThiefMaster:
Thanks a lot, pretty much everything works great now! I only noticed one small but in the "accept all options when i'm available" feature. This only works you are in the same timezone as the newdle, as it uses newdle-tz timeslots with user-tz busy times.

Here's a fix that works for me - maybe you have a nicer idea, otherwise feel free to just commit that (ideally with the `--author` set to me ;))

```patch
diff --git a/newdle/client/src/answerSelectors.js b/newdle/client/src/answerSelectors.js
index 2877e14..d639194 100644
--- a/newdle/client/src/answerSelectors.js
+++ b/newdle/client/src/answerSelectors.js
@@ -32,6 +32,12 @@ export const getLocalNewdleTimeslots = createSelector(
       )
     )
 );
+/** A mapping from user-tz timeslots to newdle-tz timeslots */
+const getLocalNewdleTimeslotsMap = createSelector(
+  getNewdleTimeslots,
+  getLocalNewdleTimeslots,
+  (timeslots, localTimeslots) => _.zipObject(localTimeslots, timeslots)
+);
 export const getCalendarDates = createSelector(
   getLocalNewdleTimeslots,
   timeslots =>
@@ -81,17 +87,17 @@ export const isAllAvailableSelectedExplicitly = state => state.answer.allAvailab
 /** All time slots during which the person is free */
 const getFreeTimeslots = createSelector(
   getFlatBusyTimes,
-  getNewdleTimeslots,
+  getLocalNewdleTimeslotsMap,
   getNewdleDuration,
-  (busyTimes, timeslots, duration) => {
+  (busyTimes, timeslotMap, duration) => {
     busyTimes = busyTimes.map(pair => pair.map(t => toMoment(t, moment.HTML5_FMT.DATETIME_LOCAL)));
-    timeslots = timeslots.map(t => [
+    const localTimeslots = Object.keys(timeslotMap).map(t => [
       toMoment(t, moment.HTML5_FMT.DATETIME_LOCAL),
       toMoment(t, moment.HTML5_FMT.DATETIME_LOCAL).add(duration, 'm'),
     ]);
-    return timeslots
+    return localTimeslots
       .filter(ts => !busyTimes.some(bt => overlaps(ts, bt)))
-      .map(([start]) => serializeDate(start, moment.HTML5_FMT.DATETIME_LOCAL))
+      .map(([start]) => timeslotMap[serializeDate(start, moment.HTML5_FMT.DATETIME_LOCAL)])
       .sort();
   }
 );
```

newdle/client/src/components/ParticipantTable.js
Diego-Zulu: I adds a lot of clutter in the case you have many slots, but I believe that it also helps people not loosing track about the original timestamp. But I do also see a use case for rendering it only once, specially because in the future we may decide to display this time in the ParticipantTable in the user's local timezone and we could use that same place to indicate the difference in timezones

For now, I noticed that I can display an abbreviation for a given timezone, so I'll go with that to reduce clutter

ThiefMaster: What kind of abbreviation do you mean? If it's `UTC+xx` then it may be confusing for people, so I'd rather keep the repeated long `Europe/Zurich` etc

newdle/client/src/components/answer/AnswerPage.js
Diego-Zulu: 1) Yes, I believe we should still display both of them, to prevent confusion from people
2) I agree the semantical meaning is not correct, so I'll change it to something that semantically makes more sense

newdle/client/src/components/ParticipantTable.js
Diego-Zulu: Okay, I agree, abbreviation are not as cool as I thought 😄 . Sao Paulo timezone's abbreviation is just `(-03)`, so I agree that leaving the long repeated timezone is better

ThiefMaster: yeah, let's keep it as-is then :)

newdle/client/src/components/answer/AnswerPage.js
Diego-Zulu: Done, take a look at it now

ThiefMaster (approved):
Great PR, thanks a lot!

PR #220, opened 2020-09-01
Fixing SKIP_LOGIN's logout and multipass redirects

Diego-Zulu:
## Description:

This PR fixes two problems that appeared if using `SKIP_LOGIN = True` option without any multipass credentials in your `newdle.cfg`:

1. Couldn't sign in because I was getting redirected into `/login/newdle-sso` when trying to reach `/login/`, and newdle-sso would fail due to missing credentials
2. Couldn't sign out without getting flask error. This was simply because the logout endpoint had no logic reflecting `SKIP_LOGIN = True` option

## Todo:

N/A

## Checks:

- All tests pass ✅ 
- All linting rules pass ✅

## Manual testing:

Environment setup:

```
make
source ./.venv/bin/activate
make flask-server
make react-server
```
- `SKIP_LOGIN = True` on `newdle.cfg`
- No multipass credentials provided on `newdle.cfg`

1. I can sign in correctly into the test user ✅ 
2. I can sign out from the test user ✅ 

## Questions:

N/A

### DIFF
--- a/newdle/auth.py
+++ b/newdle/auth.py
@@ -1,4 +1,4 @@
-from flask import Blueprint, current_app, render_template, url_for
+from flask import Blueprint, current_app, redirect, render_template, session, url_for
 
 from .core.auth import app_token_from_dummy, multipass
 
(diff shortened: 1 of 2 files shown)


ThiefMaster (approved):
works great, thanks for the fix!

PR #229, opened 2020-09-14
Add confirmation modal when emailing participants

### LINKED ISSUE #135: Add confirmation step to "e-mail participants"
It's quite easy to press it by mistake. A possible solution would be a confirmation modal.
We should also clarify in that dialog:
 * how many participants will receive an e-mail (so that the user can understand how much spam they will generate);
 * whether any of the participants (and how many) do not have an e-mail address (and possible also list them) - those are the "unknown" participants who were added ad-hoc.

image

### LINKED ISSUE #227: Mention activating the virtualenv in README

hannahshort:
- Update README to include venv for running db migrations (closes #227)
- Add confirmation modal when emailing participants on summary page 
- List participants without email that will not be contacted
- Display total that will be emailed (is this clear enough?)

Screenshot 2020-09-15 at 11 37 05

closes #135
closes #227

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ $ PYTHON=/usr/bin/python3.8 make
 
 ## Database schema
 
-Before running the alembic migrations make sure you  have created a database called `newdle` (or adjust the config file). Having done that, run `flask db upgrade` to upgrade the schema.
+Before running the alembic migrations make sure you  have created a database called `newdle` (or adjust the config file). Having done that, enter your virtual environment, ``source .venv/bin/activate``, and run `flask db upgrade` to upgrade the schema.
 
 ## Running the development server
 
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ decided (still work in progress!).
 
 ## Development
 
-We chose Python 3.8 as the backend language, so make sure you have it installed. To prepare the development environment it is enough to run `make` which takes care of installing all required dependencies inside a new virtualenv. Typically that will be the `.venv` directory unless you override the environment variable `VENV` *e.g.* `VENV=.virtualenv make`.
+We chose Python 3.8 as the backend language, so make sure you have it installed. To prepare the development environment it is enough to run `make` which takes care of installing all required dependencies inside a new virtualenv. Typically that will be the `.venv` directory unless you override the environment variable `VENV` *e.g.* `VENV=.virtualenv make`. Activate your virtualenv using ``source .venv/bin/activate`` since this is required to run the various `flask` comments that come later.
 
 Make sure you have the `python3.8` binary in your PATH. You can also use the `PYTHON` environment variable to override the location of the
 `python` binary. *e.g.:*
@@ -55,7 +55,7 @@ $ PYTHON=/usr/bin/python3.8 make
 
 ## Database schema
 
-Before running the alembic migrations make sure you  have created a database called `newdle` (or adjust the config file). Having done that, enter your virtual environment, ``source .venv/bin/activate``, and run `flask db upgrade` to upgrade the schema.
+Before running the alembic migrations make sure you  have created a database called `newdle` (or adjust the config file). Having done that, run `flask db upgrade` to upgrade the schema.
 
 ## Running the development server
 
--- a/newdle/client/src/components/RecipientList.js
+++ b/newdle/client/src/components/RecipientList.js
@@ -33,7 +33,7 @@ ListItems.propTypes = {
 export default function RecipientList({recipients, color, icon}) {
   const [renderAll, setRenderAll] = useState(false);
 
-  if (renderAll || recipients.length <= DEFAULT_RECIPIENTS_SHOWN + 1) {
+  if (renderAll || recipients.length <= DEFAULT_RECIPIENTS_SHOWN) {
     return <ListItems recipients={recipients} color={color} icon={icon} />;
   } else {
     return (
@@ -44,6 +44,7 @@ export default function RecipientList({recipients, color, icon}) {
           icon={icon}
         />
         <Label
+          as="a"
           className={styles['more-participants']}
           color="blue"
           onClick={evt => {
@@ -54,6 +55,7 @@ export default function RecipientList({recipients, color, icon}) {
         >
           show all
         </Label>
+        <br />
       </>
     );
   }
--- a/newdle/client/src/components/RecipientList.js
+++ b/newdle/client/src/components/RecipientList.js
@@ -8,7 +8,6 @@ const DEFAULT_RECIPIENTS_SHOWN = 10;
 function ListItems({recipients, color, icon}) {
   return (
     <>
-      <br />
       <List>
         {recipients.map(p => {
           return (
@@ -24,7 +23,9 @@ function ListItems({recipients, color, icon}) {
 }
 
 ListItems.propTypes = {
-  recipients: PropTypes.array.isRequired,
+  recipients: PropTypes.arrayOf(
+    PropTypes.shape({id: PropTypes.number.isRequired, name: PropTypes.string.isRequired})
+  ).isRequired,
   color: PropTypes.string.isRequired,
   icon: PropTypes.string.isRequired,
 };
@@ -51,7 +52,7 @@ export default function RecipientList({recipients, color, icon}) {
           }}
           circular
         >
-          {`show all`}
+          show all
         </Label>
       </>
     );
@@ -59,7 +60,9 @@ export default function RecipientList({recipients, color, icon}) {
 }
 
 RecipientList.propTypes = {
-  recipients: PropTypes.array.isRequired,
+  recipients: PropTypes.arrayOf(
+    PropTypes.shape({id: PropTypes.number.isRequired, name: PropTypes.string.isRequired})
+  ).isRequired,
   color: PropTypes.string.isRequired,
   icon: PropTypes.string.isRequired,
 };
--- a/newdle/client/src/components/RecipientList.js
+++ b/newdle/client/src/components/RecipientList.js
@@ -24,7 +24,10 @@ function ListItems({recipients, color, icon}) {
 
 ListItems.propTypes = {
   recipients: PropTypes.arrayOf(
-    PropTypes.shape({id: PropTypes.number.isRequired, name: PropTypes.string.isRequired})
+    PropTypes.shape({
+      id: PropTypes.number.isRequired,
+      name: PropTypes.string.isRequired,
+    })
   ).isRequired,
   color: PropTypes.string.isRequired,
   icon: PropTypes.string.isRequired,
--- a/newdle/client/src/components/RecipientList.js
+++ b/newdle/client/src/components/RecipientList.js
@@ -66,7 +66,10 @@ export default function RecipientList({recipients, color, icon}) {
 
 RecipientList.propTypes = {
   recipients: PropTypes.arrayOf(
-    PropTypes.shape({id: PropTypes.number.isRequired, name: PropTypes.string.isRequired})
+    PropTypes.shape({
+      id: PropTypes.number.isRequired,
+      name: PropTypes.string.isRequired,
+    })
   ).isRequired,
   color: PropTypes.string.isRequired,
   icon: PropTypes.string.isRequired,
--- a/newdle/client/src/components/RecipientList.js
+++ b/newdle/client/src/components/RecipientList.js
@@ -23,6 +23,12 @@ function ListItems({recipients, color, icon}) {
   );
 }
 
+ListItems.propTypes = {
+  recipients: PropTypes.array.isRequired,
+  color: PropTypes.string.isRequired,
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -55,7 +55,7 @@ $ PYTHON=/usr/bin/python3.8 make
 
 ## Database schema
 
-Before running the alembic migrations make sure you  have created a database called `newdle` (or adjust the config file). Having done that, run `flask db upgrade` to upgrade the schema.
+Before running the alembic migrations make sure you  have created a database called `newdle` (or adjust the config file). Having done that, enter your virtual environment, ``source .venv/bin/activate``, and run `flask db upgrade` to upgrade the schema.
```
ThiefMaster, line 58: I would document this in the general `## Development` section, since it makes sense to do it when doing any kind of development.

    Activate your virtualenv using ``...`` since this is required to run the various `flask` comments.

or something like this

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -65,6 +88,48 @@ export default function SummaryPage() {
...
+              </Modal.Header>
+              <Modal.Content>
+                {hasParticipantsWithoutEmail && (
+                  <>
+                    Some of your recipients do not have e-mail addresses and will not be contacted:
+                    <br />
+                    <List>
+                      {participantsWithoutEmail.map((p, index) => {
+                        return (
+                          <List.Item key={index}>
```
ThiefMaster, line 113: It is discouraged to use the `index` as a key. Check if we have something better that uniquely identifies participants (`p.id` maybe)

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -65,6 +88,48 @@ export default function SummaryPage() {
...
+                            <List.Icon color="red" name="close" />
+                            <List.Content>{p.name}</List.Content>
+                          </List.Item>
+                        );
+                      })}
+                    </List>
+                    <br />
+                  </>
+                )}
+                {participantsWithEmail.length} participants will be e-mailed.
```
ThiefMaster, line 123: Should we show the actual list of participants who will be emailed there?

ThiefMaster:
Just two small things:

- the README change is unrelated to the rest so while including it in this PR is perfectly fine (it's quite small), I would really use a separate commit for it
- we usually write commit messages using present tense, e.g. "Add confirmation modal when emailing participants" (also, max 50 chars in the first line - see [this page](https://chris.beams.io/posts/git-commit/) for some good tips on writing great commit messages)

newdle/client/src/components/summary/SummaryPage.js
hannahshort: As you prefer - happy to add both lists. Should we cut off the list at a certain point if it is very long?

ThiefMaster: We could do it similar how we do it in the participant list - ie cut off after a certain number with a "show all" link.

newdle/client/src/selectors.js
```diff
@@ -96,10 +96,18 @@ export const newdleHasParticipantsWithEmail = createSelector(
   getNewdleParticipants,
   participants => participants.some(p => p.email !== null)
 );
+export const newdleParticipantsWithEmail = createSelector(
```
plourenco, line 99: What do you think about composing the selectors to avoid repeating the condition `p => p.email !== null`?

Such as:
```
export const newdleParticipantsWithEmail = createSelector(
  getNewdleParticipants,
  participants => participants.filter(p => p.email !== null)
);
export const newdleHasParticipantsWithEmail = createSelector(
  newdleParticipantsWithEmail,
  participants => participants.length
);
```

ThiefMaster: I was actually thinking about commenting the exact same thing, but then decided against it because in the end it's very small and - not that it matters even for large newdles - `some()` is still more efficient since it stops at the first successful check.

Anyway, if you decide to change it, make sure to convert to bool, i.e. `!!participants.length` or `participants.length !== 0`

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -36,14 +54,52 @@ export default function SummaryPage() {
...
+  }, [setMailModalOpen, newdle, _sendResultEmails]);
 
   if (!newdle) {
     return <Loader active />;
   }
 
   const mailSent = sendMailResponse !== null;
 
+  const recipientList = (list, color, icon, limit, limitSetter) => {
+    var sliced_list = list.slice(0, limit);
```
ThiefMaster, line 73: `var` is a no-no nowadays (always use `const` or `let` depending on whether you reassign or not)

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -36,14 +54,52 @@ export default function SummaryPage() {
...
+    return (
+      <List>
+        {sliced_list.map((p, index) => {
```
ThiefMaster, line 76: `index` is unused

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -36,14 +54,52 @@ export default function SummaryPage() {
...
+    _sendResultEmails(newdle.code);
+  }, [setMailModalOpen, newdle, _sendResultEmails]);
 
   if (!newdle) {
     return <Loader active />;
   }
 
   const mailSent = sendMailResponse !== null;
 
+  const recipientList = (list, color, icon, limit, limitSetter) => {
```
ThiefMaster, line 72: I wonder if this shouldn't be a new component instead

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -36,14 +54,52 @@ export default function SummaryPage() {
...
+              <List.Content>{p.name}</List.Content>
+            </List.Item>
+          );
+        })}
+        {limit < list.length && (
+          <List.Item>
+            <List.Icon color="black" name="ellipsis horizontal" />
+            <List.Content>
+              <a
+                href="#!"
```
ThiefMaster, line 89: I think this adds an ugly `#!` to the URL - if you remove the href clicking the link should not change the URL at all.

Otherwise (in case you want a href for some reason) you could also use `evt.preventDefault()` in the onClick handler to prevent it from changing the URL.

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -36,14 +54,52 @@ export default function SummaryPage() {
...
+                onClick={() => {
+                  limitSetter(list.length);
```
ThiefMaster, line 91: Since you switch from "limited" to "everything" you can simplify this by simply storing a bool whether the list is expanded or not

newdle/client/src/selectors.js
hannahshort: Good spot, changed

newdle/client/src/components/summary/SummaryPage.js
hannahshort: The problem here is that I think having a link in a List.Item looks nicer and is more intuitive for Users, but you can no longer have a <a /> without a valid href, and I couldn't find a nice way to format a Button as a link. Changed in this latest version to use a Button that you're already using. If you're happy with that we could leave it

ThiefMaster: `<a>` should work fine without a href.. or is the styling different in that case?

newdle/client/src/components/RecipientList.js
```diff
@@ -23,6 +23,12 @@ function ListItems({recipients, color, icon}) {
   );
 }
 
+ListItems.propTypes = {
+  recipients: PropTypes.array.isRequired,
```
ThiefMaster, line 27: `PropTypes.arrayOf(PropTypes.shape({id: PropTypes.number.isRequired, name: PropTypes.string.isRequired})).isRequired,`

newdle/client/src/components/summary/SummaryPage.js
hannahshort: Screenshot 2020-09-15 at 13 53 55

newdle/client/src/components/RecipientList.js
```diff
@@ -0,0 +1,70 @@
...
+import PropTypes from 'prop-types';
+import {Label, List} from 'semantic-ui-react';
+import styles from './ParticipantTable.module.scss';
+
+const DEFAULT_RECIPIENTS_SHOWN = 10;
+
+function ListItems({recipients, color, icon}) {
+  return (
+    <>
+      <br />
```
ThiefMaster, line 11: I would rather keep this outside (ie right before `<RecipientList />`)

newdle/client/src/components/summary/SummaryPage.js
hannahshort:
```
Line 56:9:  The href attribute is required for an anchor to be keyboard accessible. Provide a valid, navigable address as the href value. If you cannot provide an href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid
```

newdle/client/src/components/RecipientList.js
```diff
@@ -0,0 +1,70 @@
...
+        <Label
+          className={styles['more-participants']}
+          color="blue"
+          onClick={evt => {
+            evt.stopPropagation();
+            setRenderAll(true);
+          }}
+          circular
+        >
+          {`show all`}
```
ThiefMaster, line 54: Just `show all` without anything around should do the same ;)

newdle/client/src/components/RecipientList.js
```diff
@@ -0,0 +1,70 @@
...
+        </Label>
+      </>
+    );
+  }
+}
+
+RecipientList.propTypes = {
+  recipients: PropTypes.array.isRequired,
```
ThiefMaster, line 62: same as above

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -65,6 +88,34 @@ export default function SummaryPage() {
...
+              <Modal.Content>
+                {hasParticipantsWithoutEmail && (
+                  <>
+                    Some of your recipients do not have e-mail addresses and will not be contacted:
+                    <RecipientList recipients={participantsWithoutEmail} color="red" icon="close" />
+                    <br />
+                  </>
+                )}
+                {participantsWithEmail.length} participant
+                {participantsWithEmail.length > 1 && <>s</>} will be e-mailed:
```
ThiefMaster, line 108: I like that you take care of pluralization but I think it would  be better to do that only once we add i18n - that'll allow doing it nicely without "hacks" like this. For now I would just always use the plural form.

If you prefer to keep it: `{participantsWithEmail.length > 1 && 's'}` would be a more concise version

newdle/client/src/components/summary/SummaryPage.js
ThiefMaster: ah that warning :/ i think going for the label like in the other place was a good choice :)

hannahshort:
@ThiefMaster I think all your points are addressed

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -65,6 +88,35 @@ export default function SummaryPage() {
               )}
             </Message>
           )}
+          {
```
ThiefMaster, line 91: I don't think these curly braces are needed

newdle/client/src/components/RecipientList.js
```diff
@@ -0,0 +1,73 @@
...
+          );
+        })}
+      </List>
+    </>
+  );
+}
+
+ListItems.propTypes = {
+  recipients: PropTypes.arrayOf(
+    PropTypes.shape({id: PropTypes.number.isRequired, name: PropTypes.string.isRequired})
```
ThiefMaster, line 27: I thought prettier would take care of it but it didn't :(

```suggestion
    PropTypes.shape({
      id: PropTypes.number.isRequired,
      name: PropTypes.string.isRequired,
    })
```

ThiefMaster:
image

the margins of the "show all" button are a bit weird - they are closer to the second list than the one they actually apply to

ThiefMaster:
Besides those minor stylistic issues it looks very nice and works great! :)

hannahshort:
> image
> 
> the margins of the "show all" button are a bit weird - they are closer to the second list than the one they actually apply to
> 
> we also need `cursor: pointer;` on it (just like on `.more-participants`)

Weirdly the CSS already contains `cursor: pointer` but doesn't seem to render properly. Fixed by adding `as='a'` to the Label.

hannahshort:
btw - sorry I don't have time to do the backend stuff :(

ThiefMaster:
no problem, i think it makes perfect sense to open a separate issue for this

ThiefMaster (approved):

PR #230, opened 2020-09-14
Create a comments field for participants

### LINKED ISSUE #142: Allow participants to leave comments
Not sure how this one will look UI-wise, but the goal is to allow participants to clarify some of their choices. *E.g. "I'm sorry, I cannot on day X because I'm on holiday"*

A UI approach option could be to add a small "message" icon next to the user's name on the newdle summary page.

plourenco:
Closes #142 

A proposal for participant comments. 
These can easily be set, unset or replaced in the answers page.

image

### DIFF
--- a/newdle/client/src/components/answer/AnswerPage.js
+++ b/newdle/client/src/components/answer/AnswerPage.js
@@ -89,7 +89,7 @@ export default function AnswerPage() {
   const participantHasAnswers = !!Object.keys(participantAnswers).length;
   const participant = useSelector(getParticipant);
   const [_comment, setComment] = useState(null);
-  const comment = _comment === null && participant ? participant.comment : _comment;
+  const comment = _comment === null && participant ? participant.comment : _comment || '';
   const participantUnknown = useSelector(isParticipantUnknown);
   const participantAnswersChanged = useSelector(haveParticipantAnswersChanged);
   const busyTimesLoaded = useSelector(hasBusyTimes);
@@ -230,7 +230,7 @@ export default function AnswerPage() {
             type="text"
             placeholder="Leave a comment..."
             className={styles['comment-submit']}
-            value={comment || ''}
+            value={comment}
             onChange={(__, {value}) => setComment(value)}
             action={!isSmallScreen}
           >
--- a/newdle/client/src/components/answer/AnswerPage.js
+++ b/newdle/client/src/components/answer/AnswerPage.js
@@ -29,7 +29,7 @@ import {
   setParticipantCode,
 } from '../../actions';
 import client from '../../client';
-import {useIsMobile, useIsSmallScreen, usePageTitle} from '../../util/hooks';
+import {useIsSmallScreen, usePageTitle} from '../../util/hooks';
 import styles from './answer.module.scss';
 
 function ParticipantName({unknown, setName, onSubmit, disabled}) {
@@ -88,8 +88,8 @@ export default function AnswerPage() {
   const participantAnswers = useSelector(getParticipantAnswers);
   const participantHasAnswers = !!Object.keys(participantAnswers).length;
   const participant = useSelector(getParticipant);
-  const [_comment, setComment] = useState(undefined);
-  const comment = _comment == null && participant ? participant.comment : _comment;
+  const [_comment, setComment] = useState(null);
+  const comment = _comment === null && participant ? participant.comment : _comment;
   const participantUnknown = useSelector(isParticipantUnknown);
   const participantAnswersChanged = useSelector(haveParticipantAnswersChanged);
   const busyTimesLoaded = useSelector(hasBusyTimes);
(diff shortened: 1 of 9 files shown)


newdle/models.py
```diff
@@ -90,6 +90,7 @@ class Participant(db.Model):
         unique=True,
     )
     _answers = db.Column('answers', JSONB, nullable=True, default={})
+    comment = db.Column('comment', db.String, nullable=False, default='')
```
ThiefMaster, line 93: also add `server_default=''`

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -88,11 +88,14 @@ export default function AnswerPage() {
   const participantAnswers = useSelector(getParticipantAnswers);
   const participantHasAnswers = !!Object.keys(participantAnswers).length;
   const participant = useSelector(getParticipant);
+  const [_comment, setComment] = useState(undefined);
+  const comment = _comment == null && participant ? participant.comment : _comment;
```
ThiefMaster, line 92: why not use `null` for the default state? then you can use `===` here instead of using the null/undefined coercion of `==`

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -223,22 +226,34 @@ export default function AnswerPage() {
...
-            onClick={() => {
-              answerNewdle();
+          <Input
+            type="text"
+            placeholder="Leave a comment..."
+            className={styles['comment-submit']}
+            value={comment || ''}
+            onChange={(__, data) => {
+              setComment(data.value);
             }}
```
ThiefMaster, line 236:
```suggestion
            onChange={(__, {value}) => setComment(value)}
```

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -223,22 +226,34 @@ export default function AnswerPage() {
...
+            <input />
+            <Button
+              size="large"
+              color={participantHasAnswers ? 'teal' : 'violet'}
+              content={participantHasAnswers ? 'Update your answer' : 'Send your answer'}
+              disabled={
+                submitting ||
+                !canSubmit ||
+                (participantCode && !participant) ||
+                (participantHasAnswers && !participantAnswersChanged && _comment == null)
```
ThiefMaster, line 248: `===` (see above)

newdle/client/src/components/answer/AnswerPage.js
plourenco: I had null initially. But I wanted to avoid the unpredictable case of having the input `value={null}`. 
It will complain something like: "set as undefined instead of null..."

But if you agree that this is not worth the concern, I can change it.

ThiefMaster: won't this be covered by `comment || ''` in the input?

plourenco: Yes, it will be. I added that shortly after and eventually forgot it was even there.

ThiefMaster:
image

Could we make the input field a bit wider on desktops? Maybe align the text to the left, the button to the right, and just let flexbox grow the input field to fill the remaining space in the middle?


---

In the mobile view there's an issue with the border radiuses (there shouldn't be rounded corners between two items that are touching each other), but I would probably just put the field above the button instead of next to it (again to have more space for it).

image

newdle/client/src/components/answer/AnswerPage.js
```diff
@@ -223,22 +226,32 @@ export default function AnswerPage() {
...
+            <input />
+            <Button
+              size="large"
+              color={participantHasAnswers ? 'teal' : 'violet'}
+              content={participantHasAnswers ? 'Update your answer' : 'Send your answer'}
+              disabled={
+                submitting ||
+                !canSubmit ||
+                (participantCode && !participant) ||
+                (participantHasAnswers && !participantAnswersChanged && _comment === null)
```
ThiefMaster, line 246: This will keep the button enabled if you change the comment and then revert that change. Better check if the comment is not equal to the stored comment.

plourenco: Good idea. Since the condition on the top was simplified, this can be enhanced.

plourenco:
> Could we make the input field a bit wider on desktops? Maybe align the text to the left, the button to the right, and just let flexbox grow the input field to fill the remaining space in the middle?
> 
> In the mobile view there's an issue with the border radiuses (there shouldn't be rounded corners between two items that are touching each other), but I would probably just put the field above the button instead of next to it (again to have more space for it).

I increased the responsiveness and decoupled the elements. 
Also made the input slightly larger. I tried to make it full-width but it didn't seem as nice as having all this information right-aligned in one place.

ThiefMaster:
I like it now!

But it looks like the display of the comment in the summary broke due to the changes in #235:

image

plourenco:
> I like it now!
> 
> But it looks like the display of the comment in the summary broke due to the changes in #235:

Thanks, fixed in the latest.

ThiefMaster (approved):

PR #231, opened 2020-09-14
Attach icalendar invite for participants

### LINKED ISSUE #143: Attach iCalendar file (*.ics) to result e-mail
The result e-mail is currently very simple and it would be great if we could send an `*.ics` file along with it, so that people can just add the new meeting to their calendars. Integration with Indico will make this unnecessary in organizations using it, but it's still a nice thing to have in places where Indico is not an option.

image

saibot94:
Closes #143 .

I tested it out with gmail and it seems to parse the information properly.

image

This is the generated ICS file for the above, in case you want to take a look:

```
BEGIN:VCALENDAR
BEGIN:VEVENT
SUMMARY:Something like this
DTSTART;TZID=Europe/Paris;VALUE=DATE-TIME:20200914T110000
DTEND;TZID=Europe/Paris;VALUE=DATE-TIME:20200914T113000
DTSTAMP;VALUE=DATE-TIME:20200914T130547Z
ORGANIZER;CN="Cristi Sch":MAILTO:REDACTED
END:VEVENT
END:VCALENDAR
```

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -10,6 +10,7 @@ from pytz import common_timezones_set, timezone
 from sqlalchemy.orm import selectinload
 from werkzeug.exceptions import Forbidden, ServiceUnavailable, UnprocessableEntity
 
+from .calendar import create_calendar_event
 from .core.auth import search_users, user_info_from_app_token
 from .core.db import db
 from .core.util import (
@@ -399,6 +400,8 @@ def send_result_emails(code):
     date = newdle.final_dt.strftime('%-d %B %Y')
     start_time = newdle.final_dt.strftime('%H:%M')
     end_time = (newdle.final_dt + newdle.duration).strftime('%H:%M')
+    ical_data = create_calendar_event(newdle)
+    attachments = [('invite.ics', ical_data, 'text/calendar')]
     notify_newdle_participants(
         newdle,
         f'Result: {newdle.title}',
@@ -412,5 +415,6 @@ def send_result_emails(code):
             'end_time': end_time,
             'timezone': newdle.timezone,
         },
+        attachments,
     )
     return '', 204
--- a/newdle/calendar.py
+++ b/newdle/calendar.py
@@ -1,0 +1,43 @@
+from datetime import datetime
+
+from flask import g
+from icalendar import Calendar, Event, vCalAddress, vText
+from pytz import timezone
+
+
+def _email_to_vcal_address(email):
+    return vCalAddress(f'MAILTO:{email}')
+
+
+def _add_participant_to_ical_event(event, participant):
+    """Adds a participant as an attendee to an ical event."""
+    attendee = _email_to_vcal_address(participant.email)
+    attendee.params['cn'] = vText(participant.name)
+    attendee.params['ROLE'] = vText('REQ-PARTICIPANT')
+    event.add('attendee', attendee, encode=0)
+
+
+def create_calendar_event(newdle):
+    """Create an icalendar event based on a newdle."""
+    calendar = Calendar()
+    event = Event()
+    tz = timezone(newdle.timezone)
+    start_dt = tz.localize(newdle.final_dt)
+    end_dt = start_dt + newdle.duration
+
+    event.add('summary', newdle.title)
+    event.add('dtstart', start_dt)
+    event.add('dtend', end_dt)
+    event.add('dtstamp', datetime.utcnow())
+
+    organizer = _email_to_vcal_address(g.user['email'])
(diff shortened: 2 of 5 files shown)


newdle/api.py
```diff
@@ -399,6 +400,9 @@ def send_result_emails(code):
     date = newdle.final_dt.strftime('%-d %B %Y')
     start_time = newdle.final_dt.strftime('%H:%M')
     end_time = (newdle.final_dt + newdle.duration).strftime('%H:%M')
+    event = create_calendar_event(newdle)
+    # Attachments have to be in the format: (filename, content_bytes, MIMEType)
```
ThiefMaster, line 404: not sure if that comment is super useful here, it' kind of obvious that this is what the library expects..

if we want to document this, a better place would be the docstring of `notify_newdle_participants` which doesn't exist so for now i'd just not document it ;)

newdle/api.py
```diff
@@ -399,6 +400,9 @@ def send_result_emails(code):
     date = newdle.final_dt.strftime('%-d %B %Y')
     start_time = newdle.final_dt.strftime('%H:%M')
     end_time = (newdle.final_dt + newdle.duration).strftime('%H:%M')
+    event = create_calendar_event(newdle)
```
ThiefMaster, line 403: i'd rename this to `ics` or `ical_data` - it's a string, not an ical Event object

newdle/calendar.py
```diff
@@ -0,0 +1,26 @@
+from datetime import datetime
+
+from flask import g
+from icalendar import Calendar, Event, vCalAddress, vText
+from pytz import timezone
+
+
+def create_calendar_event(newdle):
+    """Create an icalendar event based on a newdle"""
```
ThiefMaster, line 9:
```suggestion
    """Create an icalendar event based on a newdle."""
```

newdle/calendar.py
```diff
@@ -0,0 +1,26 @@
...
+    calendar = Calendar()
+    event = Event()
+    tz = timezone(newdle.timezone)
+    start_tz_aware = tz.localize(newdle.final_dt)
+    end_tz_aware = tz.localize(newdle.final_dt + newdle.duration)
```
ThiefMaster, line 14: This should be enough:

```suggestion
    end_tz_aware = start_tz_aware + newdle.duration
```

ThiefMaster: I'd just use `start_dt` and `end_dt` for these variable names

ThiefMaster:
Would it make sense to include the list of participants in the ical event as well if the newdle is not set to private?

newdle/api.py
saibot94: Thanks, removed it.

newdle/api.py
saibot94: Changed to `ical_data`.

newdle/calendar.py
saibot94: Changed

saibot94:
> Would it make sense to include the list of participants in the ical event as well if the newdle is not set to private?

I think it's a good idea. Trying it out now.

newdle/calendar.py
```diff
@@ -0,0 +1,43 @@
...
+from flask import g
+from icalendar import Calendar, Event, vCalAddress, vText
+from pytz import timezone
+
+
+def _email_to_vcal_address(email):
+    return vCalAddress(f'MAILTO:{email}')
+
+
+def add_participant_to_ical_event(event, participant):
```
ThiefMaster, line 12:
```suggestion
def _add_participant_to_ical_event(event, participant):
```

just like `_email_to_vcal_address` it's only used in here

newdle/calendar.py
```diff
@@ -0,0 +1,43 @@
...
+    event.add('dtstamp', datetime.utcnow())
+
+    organizer = _email_to_vcal_address(g.user['email'])
+    organizer.params['cn'] = vText(newdle.creator_name)
+    event['organizer'] = organizer
+
+    if not newdle.private:
+        participants = [p for p in newdle.participants if p.email]
+        for participant in participants:
+            add_participant_to_ical_event(event, participant)
```
ThiefMaster, line 40:
```suggestion
        for participant in newdle.participants:
            if not p.email:
                continue
            add_participant_to_ical_event(event, participant)
```

Your option is fine as well though so if you prefer that feel free to ignore this suggestion

saibot94: I think I prefer the first variant, thanks.

ThiefMaster (approved):
Very nice PR, thanks!

PR #232, opened 2020-09-14
Add Github actions for CI 

### LINKED ISSUE #207: Use GitHub actions for CI
Some stuff can probably be copied from the [Indico CI config](https://github.com/indico/indico/blob/master/.github/workflows/ci.yml)

ParthS007:
Screenshot 2020-09-15 at 17 25 28
Closes #207

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,63 @@
+name: CI
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8]
+    
+    services:
+      postgres:
+        image: postgres:12
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
+        ports:
+          - 5432
+        options: --health-cmd pg_isready --health-interval 5s --health-timeout 5s --health-retries 10
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+
+      - name: Setup newdle
+        run: |
+          export PYTHON=python
+          make
+
+      - name: Setup Database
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
+name: CI
+
+on:
+  push:
+    branches: [master, actions]
```
ThiefMaster, line 5: I think we can skip the `actions` branch here;  that was just something I used when developing the action for indico and then forgot to remove

ThiefMaster (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
```
ThiefMaster, line 14: let's keep running on 3.7 as well

.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+          python-version: ${{ matrix.python-version }}
+
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+
+      - name: Setup newdle
+        run: |
+          make
```
ThiefMaster, line 40:
```suggestion
        run: make
```

should be enough

.github/workflows/ci.yml
ParthS007: In the `Makefile`, 3.8 is hardcoded, Is there any way to make it dynamic?

Since if CI runs in `python 3.7` env, there is no `python 3.8` there and CI fails.

ThiefMaster: you can set `PYTHON=python` in the env, that should use the default python

Makefile
```diff
@@ -1,4 +1,4 @@
-PYTHON ?= python3.8
+PYTHON ?= python
```
ThiefMaster, line 1: When we discussed this I was more thinking about `export PYTHON=python` in the YAML. Because if you change it here you'll also have to change the README which currently mentions this:

> Make sure you have the `python3.8` binary in your PATH.

And of course this doesn't apply if the Makefile suddenly looks for a `python` binary.

ThiefMaster (changes requested):

ThiefMaster (approved):

PR #233, opened 2020-09-15
Fix bottommost slots position in answer page (#221)

### LINKED ISSUE #221: Bottommost slots on answer page don't render at correct position
I have noticed that slots have a slight shift in positioning the closer they are to the bottom of the timeline. Here's an example:
Screen Shot 2020-09-05 at 19 44 57
23:30 appears just to the right of 24. To my understanding, this slot should have started a little higher, right?

I quickly debugged it and it seems to be related to the fact that the timeline overflows a little bit from the Grid Row, although please dig into more detail when finding a solution

asieraguado:
image

This change makes the hours and answers columns with the same height, fixing the bottommost answers position. I used a header with a white space for having the same fixed-height elements in the hours and answers columns, but I'm not sure if it's the prettiest solution.

closes #221

### DIFF
--- a/newdle/client/src/components/answer/Calendar.js
+++ b/newdle/client/src/components/answer/Calendar.js
@@ -153,6 +153,11 @@ function calculateBusyPositions(busyTimes, minHour, maxHour) {
   });
 }
 
+function calculateHourPositions(minHour, maxHour, hourStep) {
+  const hourSpan = maxHour - minHour;
+  return _.range(0, hourSpan + hourStep, hourStep).map(i => (i / hourSpan) * 100);
+}
+
 function Hours({minHour, maxHour, hourStep}) {
   const hourSeries = hourRange(minHour, maxHour, hourStep);
   const hourSpan = maxHour - minHour;
@@ -243,6 +248,11 @@ export default function Calendar() {
                 options={item}
                 busySlots={busyByDay.find(busySlot => busySlot.date === item.date)}
                 selected={isActiveDay(item.date)}
+                hourPositions={calculateHourPositions(
+                  minHour,
+                  maxHour,
+                  Hours.defaultProps.hourStep
+                )}
               />
             </Grid.Column>
           )}
--- a/newdle/client/src/components/answer/answer.module.scss
+++ b/newdle/client/src/components/answer/answer.module.scss
@@ -22,7 +22,7 @@
   position: relative;
   width: 100%;
   height: 750px;
-  margin-top: 25px;
+  margin-top: 35px;
 
   .hour {
     position: absolute;
--- a/newdle/client/src/components/answer/answer.module.scss
+++ b/newdle/client/src/components/answer/answer.module.scss
@@ -27,6 +27,9 @@
   .hour {
     position: absolute;
     color: $grey;
+    width: 25px;
+    background-color: white;
+    z-index: 1;
   }
 }
 
@@ -171,11 +174,12 @@
 
   .hours-separator {
     position: absolute;
-    border: 0px;
-    width: 100%;
+    border: 0;
+    width: 114%;
     height: 1px;
-    margin: 0px;
+    margin: 0;
     background-color: #e6e6e6;
+    left: -7%;
   }
 }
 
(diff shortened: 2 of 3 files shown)


newdle/client/src/components/answer/Calendar.js
```diff
@@ -156,19 +156,24 @@ function Hours({minHour, maxHour, hourStep}) {
...
-          style={{top: `${(i / hourSpan) * 100}%`}}
-        >
-          {n === 0 && hourSeries[n] === 0
-            ? '0'
-            : serializeDate(toMoment({hours: hourSeries[n]}), 'k')}
-        </div>
-      ))}
-    </div>
+    <>
+      <Header as="h3" className={styles.date}>
```
pferreir, line 160: Is this really necessary? Couldn't we add some padding/margin in `hours-column` instead?

pferreir (commented):
Thanks for the PR! I am not sure abour the "empty header". Maybe there's some other way to do that?

asieraguado:
Thank you for your review @pferreir . Replaced the empty header with a different margin property.

plourenco (commented):
Thanks @asieraguado ! I was working on another PR and this problem was affecting me too.

However, I noticed that the current height and margin doesn't seem to work very well with hour-long blocks closer to the bottom (24 mid-night).

image

Same result in mobile:

image

Perhaps we can find another balance that works both for half and an-hour blocks?

asieraguado:
> Thanks @asieraguado ! I was working on #230 and this affected me too.
> 
> However, I noticed that the current height and margin doesn't seem to work very well with hour-long blocks closer to the bottom (24 mid-night).
> 
> image
> 
> Same result in mobile:
> 
> image
> 
> Perhaps we can find another balance that works both for half and an-hour blocks?

The height looks correct to me, they would look the same at the top. Maybe you mean that the selection blocks should start and end right in the middle of the hour number? (instead of starting the bottom).

plourenco (approved):
What I mentioned earlier now seems a bit opinion-based. I've checked with the team and they agree with the current positioning. So, all good on my side!

ThiefMaster:
Now that I've seen it for a time slot starting right at an even hour I'm less sure what's the proper way to align this...

master:
image

your pr:
image

ThiefMaster:
What do you think about adding some thin horizontal lines for each hour? A bit like the ones in indico's room booking calendar, just horizontal instead of vertical:

image

With those lines added we could nicely align the timeslot boxes and it would make any misalignment (like the one this PR fixes) very obvious to notice

asieraguado:
> What do you think about adding some thin horizontal lines for each hour? A bit like the ones in indico's room booking calendar, just horizontal instead of vertical:
> 
> image
> 
> With those lines added we could nicely align the timeslot boxes and it would make any misalignment (like the one this PR fixes) very obvious to notice

Sounds good. I can update the PR with something like this, if you are interested.

ThiefMaster:
i think it's a good idea :)

asieraguado:
Added a new commit to the PR with the hour separators. A timeline with a few slots (some of them overlapping with the separator) would look like this now:

image

ThiefMaster:
I would make the separators wider tan the slots - that way it'll look less strange in case of a slot starting/ending right on top of a separator

asieraguado:
> I would make the separators wider than the slots - that way it'll look less strange in case of a slot starting/ending right on top of a separator

Thanks for the idea, I made the slots a bit narrower so that the separator is wider. This looks better than widening the separator when there are multiple days as below.

image

A few _squeezed_ slots would have a bit less space, but I think it's still usable:

image

ThiefMaster:
Instead of making the slots narrower (I think not even seeing the whole start time isn't amazing), what do you think about making the hour separator span over all three columns instead of having separate ones in each column?

asieraguado:
The previous columns would look like below after the last commit.

image

newdle/client/src/components/answer/answer.module.scss
```diff
@@ -160,6 +163,16 @@
     opacity: 0.3;
     width: 100%;
   }
+
+  .hours-separator {
+    position: absolute;
+    border: 0px;
```
ThiefMaster, line 169: `0` doesn't need a unit (same below for the margin)

ThiefMaster:
Looks really nice now! I'll merge it once you rebased and fixed the conflicts .

ThiefMaster (approved):

PR #236, opened 2020-09-15
Delete old newdles CLI and model updates

### LINKED ISSUE #179: Delete newdles after a time
Newdles are something ephemeral; there's no need for long-term archival.

We should delete them after a certain time, and maybe indicate when they will happen (at least to the creator) when viewing it.

Also need to decide when to start counting against the deadline... some options:

- final date
- date when the newdle was finalized
- latest option in the newdle
- date when the newdle was created (probably not a good idea depending on how long in the future the available options are)

Also, what to do with unfinalized newdles? Delete them some time after the date of the latest choice?

saibot94:
Closes #179  

```python
# Cleanup configuration (values are in days)
# Days after which an incomplete newdle is cleaned up, after the last update.
LAST_ACTIVITY_CLEANUP_INTERVAL = 90
# Days after which a newdle with a final date is cleaned up.
FINAL_DATE_CLEANUP_INTERVAL = 30
```

I added the intervals in days because `timedelta` doesn't have a `months=` argument.


newdle/api.py
```diff
@@ -292,6 +293,8 @@ def update_newdle(args, code):
         raise Forbidden
     for key, value in args.items():
         setattr(newdle, key, value)
+    if newdle.final_dt:
+        newdle.last_update = datetime.utcnow()
```
ThiefMaster, line 297: What about adding a method like `newdle.update_lastmod()` and call that whenever something is changed that needs to bump it?

newdle/core/app.py
```diff
@@ -81,6 +83,34 @@ def _handle_exception(exc):
         return jsonify(error='internal_error'), 500
 
 
+def _configure_cli(app):
+    @app.cli.command('cleanup_newdles')
```
ThiefMaster, line 87: I'd prefer to have the actual implementation elsewhere, e.g. in a `cli.py`. Easiest option to do it would be adding a `cli` blueprint so you can define the cli functions there and then just register the blueprint in here.

newdle/core/app.py
```diff
@@ -81,6 +83,34 @@ def _handle_exception(exc):
...
+    def _cleanup_newdles():
+        """Removes old newdles from the database."""
+        last_activity_cleanup_interval = timedelta(
+            days=app.config['LAST_ACTIVITY_CLEANUP_INTERVAL']
+        )
+        final_date_cleanup_interval = timedelta(
+            days=app.config['FINAL_DATE_CLEANUP_INTERVAL']
+        )
+        now = datetime.utcnow()
+        final_newdles_to_delete = Newdle.query.filter(
```
ThiefMaster, line 97: IIRC we said that we want to require both criteria to be met to delete - so if there's a final dt we still take into account whether it's still within the lastmod interval and in that case don't delete it yet.

newdle/core/app.py
```diff
@@ -81,6 +83,34 @@ def _handle_exception(exc):
...
+            Newdle.final_dt.isnot(None),
+            Newdle.final_dt + final_date_cleanup_interval <= now,
```
ThiefMaster, line 99: `final_dt` isn't UTC but in the newdle's timezone. Not that it really matters here... 

I think something like this should work to convert it to UTC (untested):

    final_dt_utc = func.timezone('UTC', func.timezone(Newdletimezone, Newdle.final_dt))

and then use that in the query here.

newdle/migrations/versions/20200915_1105_239d7862e2fe_add_last_update_field_to_newdle_model.py
```diff
@@ -0,0 +1,36 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '239d7862e2fe'
+down_revision = '93a638b96375'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
ThiefMaster, line 20: please remove those comments ;)

tests/cli_test.py
```diff
@@ -0,0 +1,51 @@
...
+from newdle.models import Newdle
+
+
+@pytest.mark.usefixtures('db_session')
+def test_cleanup_newdles_uncompleted_tasks(cli_runner, dummy_newdle):
+    dummy_newdle.last_update = datetime.utcnow() - timedelta(
+        days=current_app.config['LAST_ACTIVITY_CLEANUP_INTERVAL'] + 1
+    )
+    db.session.commit()
+    assert 1 == Newdle.query.count()
```
ThiefMaster, line 16: Yoda conditions we don't do.

newdle/core/app.py
ThiefMaster, line 87: I'd prefer `cleanup-newdles` with a dash

tests/cli_test.py
```diff
@@ -0,0 +1,51 @@
...
+    cli_runner.invoke(args=['cleanup_newdles'])
+    assert 0 == Newdle.query.count()
+
+
+@pytest.mark.usefixtures('db_session')
+def test_cleanup_newdles_non_fitting_final_tasks(cli_runner, dummy_newdle):
+    dummy_newdle.final_dt = datetime.utcnow() - timedelta(
+        days=current_app.config['FINAL_DATE_CLEANUP_INTERVAL'] - 5
```
ThiefMaster, line 46: If black lets you do it, I'd prefer this since it looks a bit weird when the value of a kwarg has spaces and there aren't parentheses around it:

```suggestion
        days=(current_app.config['FINAL_DATE_CLEANUP_INTERVAL'] - 5)
```

tests/cli_test.py
```diff
@@ -0,0 +1,51 @@
+from datetime import datetime, timedelta
+
+import pytest
+from flask import current_app
+
+from newdle.core.db import db
+from newdle.models import Newdle
+
+
```
ThiefMaster, line 9: A test case with multiple newdles at the beginning where only some of them get deleted would be nice.

tests/api_test.py
```diff
@@ -608,10 +618,12 @@ def test_create_unknown_participant(flask_client):
     code = data.pop('code')
     id_ = data.pop('id')
     participant = Participant.query.filter_by(name=name).first()
+    newdle = Newdle.query.filter_by(code='dummy').first()
     assert data == {'answers': {}, 'auth_uid': None, 'email': None, 'name': name}
     assert Participant.query.count() == num_participants + 1
     assert participant.code == code
     assert participant.id == id_
+    assert newdle.last_update > before_update
```
ThiefMaster, line 626: I wonder if it wouldn't be better to add some separate test cases which only test that the last update timestamp gets updated whenever relevant actions are performed.

Right now this is tested by test cases that aren't really related to this..

ThiefMaster (commented):
Very nice PR!

newdle/core/app.py
saibot94: Hm, I don't think I get it. Is there any reason to check the `final_dt` interval then? Because the `lastmod` interval will always be greater since setting the `final_dt` updates the `last_update` column to the current timestamp.

tests/api_test.py
saibot94: Sure, I'll remove them from here and add separate tests.

newdle/core/app.py
ThiefMaster: the `final_dt` may be a few months in the future. see this part in my comment on the issue:

> newdles get deleted only if both criteria are met, so if i finalize a newdle taking place in 12M, it'd be deleted after 13M (with the defaeults)

ThiefMaster: so i think the logic you want here (can be done in a single query is):

    (now - lastmod > lastmod_threshold) AND (final_dt IS NULL OR now - final_dt > final_threshold)

newdle/core/app.py
saibot94: Is it worth complicating it then or just leaving it as is?

ThiefMaster: I'd be fine with leaving it as-is and just adding a comment

```
# XXX: This does not take newdle timezones into account, but a
# few hours are not significant here
``

or something like that

File: newdle/core/app.py (hunk 3 above)
Comment (saibot94) [reply newdle/core/app.py]: Updated, thanks.

**Comment by saibot94** [comment]
@ThiefMaster I think you can have another look now.

File: conftest.py [hunk 10]
```diff
@@ -110,3 +120,38 @@ def dummy_newdle(db_session, dummy_uid):
     db_session.add(newdle)
     db_session.flush()
     return newdle
+
+
+@pytest.fixture
+def multiple_dummy_newdles(db_session, dummy_uids):
```
ThiefMaster, line 126: A `create_newdle` fixture  that returns a function creating a dummy newdle would be more flexible.. so you could do something like this in your test:

```python
def test_whatever(create_newdle):
    newdle1 = create_newdle(1)
    newdle2 = create_newdle(2)
```

At least that's what we do [in indico](https://github.com/indico/indico/blob/master/indico/testing/fixtures/event.py).

If also removes the need for a `dummy_uids` fixture (you can default to the `dummy_uid` and just provide a custom uid if you want a different one)

---

In any case, nothing wrong with your option so if you want to keep it like this for now I'm fine with it as well.

newdle/api.py
```diff
@@ -293,6 +293,8 @@ def update_newdle(args, code):
         raise Forbidden
     for key, value in args.items():
         setattr(newdle, key, value)
+    if newdle.final_dt:
```
ThiefMaster, line 296: `if args:` so we set the lastmod whenver something is changed? right now that's just the `final_dt` anyway...

newdle/newdle.cfg.example
```diff
@@ -89,3 +89,9 @@ EXCHANGE_DOMAIN = ''
 EXCHANGE_PROVIDER_SERVER = ''
 EXCHANGE_PROVIDER_ACCOUNT = ''
 EXCHANGE_PROVIDER_CREDENTIALS = ('', '')
+
+# Cleanup configuration (values are in days)
+# Days after which an incomplete newdle is cleaned up after its last update.
+LAST_ACTIVITY_CLEANUP_INTERVAL = 90
```
ThiefMaster, line 95: Should we allow (and maybe even default to) `None` to disable the cleanup?

That would also allow people to set only one of the two settings to clean up based on final date or last activity *only*.

tests/api_test.py
```diff
@@ -379,6 +381,68 @@ def test_update_newdle(flask_client, dummy_newdle, dummy_uid):
     assert resp.json == expected_json
 
 
+@pytest.mark.usefixtures('dummy_newdle')
+def test_update_newdle_changes_last_update(flask_client, dummy_uid, dummy_newdle):
+    final_dt = '2019-09-12T13:30'
```
ThiefMaster, line 386: It's used only once so I'd just define it directly in the JSON data below

newdle/models.py
```diff
@@ -63,6 +71,10 @@ class Newdle(db.Model):
         cascade='all, delete-orphan',
     )
 
+    def update_lastmod(self):
+        """Sets the last_update column to the current UTC timestamp"""
```
ThiefMaster, line 75:
```suggestion
        """Set the last_update time of the newdle to the current time."""
```

newdle/cli.py
```diff
@@ -0,0 +1,34 @@
...
+
+from .core.db import db
+from .models import Newdle
+
+
+cli = Blueprint('newdle-cli', __name__, cli_group=None)
+
+
+@cli.cli.command('cleanup-newdles')
+def cleanup_newdles():
```
ThiefMaster, line 14: A `-n / --dry-run` flag that only prints what would be deleted would be nice

conftest.py
saibot94: Added, sounded like a better idea. Thanks.

newdle/api.py
saibot94: Fixed.

newdle/newdle.cfg.example
saibot94: I changed the logic a bit around, please take a look.

tests/api_test.py
saibot94: Changed.

conftest.py
```diff
@@ -70,16 +70,83 @@ def flask_client(app):
...
+
+
 @pytest.fixture
 def dummy_uid():
     return 'user123'
 
 
+@pytest.fixture
+def dummy_uids():
+    return ['user124', 'user125', 'user126']
```
ThiefMaster, line 85: unused

conftest.py
```diff
@@ -70,16 +70,83 @@ def flask_client(app):
...
+            'timeslots',
+            [
+                datetime(2019, 9, 11, 13, 0),
+                datetime(2019, 9, 11, 14, 0),
+                datetime(2019, 9, 12, 13, 0),
+                datetime(2019, 9, 12, 13, 30),
+            ],
+        )
+        kwargs.setdefault(
+            'code', 'dummy#{}'.format(id_) if id_ is not None else u'dummy'
```
ThiefMaster, line 119:
```suggestion
            'code', f'dummy#{id_}' if id_ is not None else 'dummy'
```

conftest.py
```diff
@@ -70,16 +70,83 @@ def flask_client(app):
...
 @pytest.fixture
 def dummy_participant_uid():
     return 'pig'
 
 
+@pytest.fixture
+def create_newdle(dummy_uid, db_session):
+    """Returns a callable which lets you create dummy newdles"""
+
+    def _create_newdle(id_=None, **kwargs):
```
ThiefMaster, line 97: i'm tempted to just use `id` and not bother about shadowing the builtin... it's not like we use this in here anyway

conftest.py
```diff
@@ -70,16 +70,83 @@ def flask_client(app):
...
+                datetime(2019, 9, 11, 14, 0),
+                datetime(2019, 9, 12, 13, 0),
+                datetime(2019, 9, 12, 13, 30),
+            ],
+        )
+        kwargs.setdefault(
+            'code', 'dummy#{}'.format(id_) if id_ is not None else u'dummy'
+        )
+        kwargs.setdefault(
+            'title', 'Test event {}'.format(id_) if id_ is not None else u'Test event'
```
ThiefMaster, line 122:
```suggestion
            'title', f'Test event {id_}' if id_ is not None else 'Test event'
```

conftest.py
```diff
@@ -70,16 +70,83 @@ def flask_client(app):
...
+
+    return _create_newdle
+
+
+@pytest.fixture
+def override_config(app):
+    """Returns a callable which lets you override app config variables."""
+
+    def _override_config(**kwargs):
+        app.config.update(**kwargs)
```
ThiefMaster, line 145: I think this persists after the test finished (or failed)...

this should avoid it (untested!):

```python
@pytest.fixture
def override_config(app):
    """Returns a callable which lets you override app config variables."""
    orig_config = dict(app.config)

    def _override_config(**kwargs):
        app.config.update(**kwargs)

    try:
        yield _override_config
    finally:
        app.config.update(orig_config)
```

newdle/cli.py
```diff
@@ -0,0 +1,57 @@
...
+            or_(
+                Newdle.final_dt.is_(None),
+                now - Newdle.final_dt > final_date_cleanup_interval,
+            )
+        )
+
+    # XXX: This does not take the newdle's timezone into account, but a
+    # few hours are not significant here
+    newdles_to_delete = Newdle.query.filter(*filters)
+    for newdle in newdles_to_delete:
```
ThiefMaster, line 48:
```suggestion
    newdles_to_delete = Newdle.query.filter(*filters)
    for newdle in Newdle.query.filter(*filters):
```

newdle/cli.py
```diff
@@ -0,0 +1,57 @@
...
+        final_date_cleanup_interval = timedelta(days=final_date_cleanup_days)
+        filters.append(
+            or_(
+                Newdle.final_dt.is_(None),
+                now - Newdle.final_dt > final_date_cleanup_interval,
+            )
+        )
+
+    # XXX: This does not take the newdle's timezone into account, but a
+    # few hours are not significant here
```
ThiefMaster, line 46: maybe move this up to the place where we actually check the `final_dt`?

newdle/migrations/versions/20200915_1105_239d7862e2fe_add_last_update_field_to_newdle_model.py
```diff
@@ -0,0 +1,36 @@
...
+            'last_update',
+            sa.DateTime(),
+            server_default=sa.text("(now() at time zone 'utc')::timestamp"),
+            nullable=False,
+        ),
+    )
+    op.execute(
+        'UPDATE newdles SET last_update = final_dt WHERE final_dt IS NOT NULL '
+        "AND final_dt < now() AT TIME ZONE 'utc';"
+    )
```
ThiefMaster, line 32:
```suggestion
    op.execute(
        '''
        UPDATE newdles
        SET last_update = final_dt
        WHERE final_dt IS NOT NULL AND final_dt < now() AT TIME ZONE 'utc';
        '''
    )
```

newdle/models.py
```diff
@@ -1,11 +1,13 @@
 import random
+from datetime import datetime
 from enum import auto
 
 from flask import current_app
 from sqlalchemy.dialects.postgresql import ARRAY, JSONB
 from sqlalchemy.event import listens_for
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.schema import CheckConstraint
+from sqlalchemy.sql import text
```
ThiefMaster, line 10: nothing wrong with importing it directly from SA, but `db.text()` would work as well (flask-sqlalchemy exposes various commonly used sqlalchemy symbols via `db`)

newdle/newdle.cfg.example
```diff
@@ -89,3 +89,9 @@ EXCHANGE_DOMAIN = ''
 EXCHANGE_PROVIDER_SERVER = ''
 EXCHANGE_PROVIDER_ACCOUNT = ''
 EXCHANGE_PROVIDER_CREDENTIALS = ('', '')
+
+# Cleanup configuration (values are in days)
+# Days after which an incomplete newdle is cleaned up after its last update.
```
ThiefMaster, line 94: I'd use "is deleted" here to be clear that this doesn't just "clean something up" but really deletes those newdles

newdle/cli.py
```diff
@@ -0,0 +1,57 @@
...
+def cleanup_newdles(dry_run):
+    """Removes old newdles from the database."""
+    last_activity_cleanup_days = current_app.config['LAST_ACTIVITY_CLEANUP_INTERVAL']
+    final_date_cleanup_days = current_app.config['FINAL_DATE_CLEANUP_INTERVAL']
+    now = datetime.utcnow()
+    if not last_activity_cleanup_days and not final_date_cleanup_days:
+        current_app.logger.warn(
+            'LAST_ACTIVITY_CLEANUP_INTERVAL and FINAL_DATE_CLEANUP_INTERVAL are not set.'
+        )
+        current_app.logger.warn('Nothing to do.')
```
ThiefMaster, line 30: I would only use one warning. Otherwise it's quite ugly if we ever start using sentry - then this would be logged as two separate events.

conftest.py
saibot94: Yep, works like that. Tested it and:

Before:
```
 'FINAL_DATE_CLEANUP_INTERVAL': 30,
 'LAST_ACTIVITY_CLEANUP_INTERVAL': 90,
```

After:
```
 'FINAL_DATE_CLEANUP_INTERVAL': None,
 'LAST_ACTIVITY_CLEANUP_INTERVAL': None,
```

newdle/cli.py
saibot94: Moved above.

ThiefMaster (approved):

conftest.py
```diff
@@ -80,6 +85,65 @@ def dummy_participant_uid():
     return 'pig'
 
 
+@pytest.fixture
+def create_newdle(dummy_uid, db_session):
+    """Returns a callable which lets you create dummy newdles"""
```
Leats, line 90:
```suggestion
    """Return a callable which lets you create dummy newdles."""
```

conftest.py
```diff
@@ -80,6 +85,65 @@ def dummy_participant_uid():
...
+        db_session.add(newdle)
+        db_session.flush()
+        return newdle
+
+    return _create_newdle
+
+
+@pytest.fixture
+def override_config(app):
+    """Returns a callable which lets you override app config variables."""
```
Leats, line 135:
```suggestion
    """Return a callable which lets you override app config variables."""
```

newdle/cli.py
```diff
@@ -0,0 +1,56 @@
...
+
+@cli.cli.command('cleanup-newdles')
+@click.option(
+    '-n',
+    '--dry-run',
+    is_flag=True,
+    help='If enabled, does not commit changes to the DB.',
+)
+def cleanup_newdles(dry_run):
+    """Removes old newdles from the database."""
```
Leats, line 22:
```suggestion
    """Remove old newdles from the database."""
```

newdle/newdle.cfg.example
```diff
@@ -89,3 +89,9 @@ EXCHANGE_DOMAIN = ''
 EXCHANGE_PROVIDER_SERVER = ''
 EXCHANGE_PROVIDER_ACCOUNT = ''
 EXCHANGE_PROVIDER_CREDENTIALS = ('', '')
+
+# Cleanup configuration (values are in days)
+# Days after which an incomplete newdle is delete after its last update.
```
Leats, line 94:
```suggestion
# Days after which an incomplete newdle is deleted after its last update.
```

ThiefMaster (approved):

Issue #141, opened 2019-10-24
Optionally notify newdle creator when a new answer comes in


pferreir:
This could exist in the form of a on option that is set in Step 3:

![image](https://user-images.githubusercontent.com/2699/67484365-42d81580-f668-11e9-9268-235ae31776d1.png)

(a switch/checkbox, something like "keep me notified about new answers", which could be turned on by default). It could also be modifiable from the newdle summary page.

alcalaperez:
@ThiefMaster

alcalaperez:
I'm on it (again)

alcalaperez:
![UI](https://user-images.githubusercontent.com/9081709/93086263-276a6480-f697-11ea-85d4-65c5d8c629d8.png)

UI set

PR #237, opened 2020-09-15
Optionally notify newdle creator when a new answer comes in #141

alcalaperez:
New email template
Captura de pantalla 2020-09-15 a las 21 34 41

UI
Captura de pantalla 2020-09-15 a las 21 28 41


newdle/api.py
```diff
@@ -244,16 +244,18 @@ def get_newdles_participating():
 
 @api.route('/newdle/', methods=('POST',))
 @use_kwargs(NewNewdleSchema(), locations=('json',))
-def create_newdle(title, duration, timezone, timeslots, participants, private):
+def create_newdle(title, duration, timezone, timeslots, participants, private, notify):
     newdle = Newdle(
         title=title,
         creator_uid=g.user['uid'],
         creator_name=f'{g.user["first_name"]} {g.user["last_name"]}',
+        creator_email=g.user["email"],
```
ThiefMaster, line 252: single quotes please ;)

newdle/api.py
```diff
@@ -355,6 +357,21 @@ def update_participant(args, code, participant_code):
     for key, value in args.items():
         setattr(participant, key, value)
     db.session.commit()
+    if participant.newdle.notify:
+        notify_newdle_creator(
+            participant.newdle,
+            participant,
+            f'{participant.name} joined to {participant.newdle.title}',
```
ThiefMaster, line 364: "joined to" sounds rather strange. What about "{name} replied to {title}" or something like that?

newdle/api.py
```diff
@@ -355,6 +357,21 @@ def update_participant(args, code, participant_code):
...
+            'joined_email.txt',
+            'joined_email.html',
+            lambda p: {
+                'participant': participant.name,
+                'title': participant.newdle.title,
+                'answer_link': url_for(
+                    'newdle', code=participant.newdle.code, participant_code=participant.code, _external=True
+                ),
+            },
```
ThiefMaster, line 373: Unlike `notify_newdle_participants` this one is always about a single participant, so we don't need the lambda and can instead pass the context directly.

newdle/api.py
```diff
@@ -387,7 +404,7 @@ def create_participant(code):
             name=name, email=g.user['email'], auth_uid=g.user['uid']
         )
         newdle.participants.add(participant)
-        db.session.commit()
+        db.session.commit()
```
ThiefMaster, line 407: I'm very surprised this wasn't removed by `black` during the pre-commit autoformatting... In any case, you may want to configure your editor to remove trailing whitespace since it's pretty much never wanted.

newdle/client/src/actions.js
```diff
@@ -148,6 +149,10 @@ export function setPrivate(isPrivate) {
   return {type: SET_PRIVATE, private: isPrivate};
 }
 
+export function setNotification(isNotified) {
+  return {type: SET_NOTIFICATION, notify: isNotified};
+}
```
ThiefMaster, line 154:
```suggestion
export function setNotification(notify) {
  return {type: SET_NOTIFICATION, notify};
}
```

In `setPrivate` we only need the different name because using `private` there caused some error (since `private` is considered a reserved word in some contexts).

newdle/client/src/components/creation/creation.module.scss
```diff
@@ -81,10 +81,16 @@
...
     justify-content: space-between;
     margin: 1em 0 0 1em;
     color: #808080;
     padding-bottom: 15px;
   }
+  .options div {
+    margin: 0em 0.5em 1em 0;
+  }
+
+  .advanced-checkbox:global(.ui.checkbox) {
```
ThiefMaster, line 93: Just `.advanced-checkbox` didn't work?

newdle/client/src/components/creation/creation.module.scss
```diff
@@ -81,10 +81,16 @@
...
 
   .options {
-    display: flex;
     justify-content: space-between;
     margin: 1em 0 0 1em;
     color: #808080;
     padding-bottom: 15px;
   }
+  .options div {
+    margin: 0em 0.5em 1em 0;
```
ThiefMaster, line 90:
```suggestion
    margin: 0 0.5em 1em 0;
```

You don't need a unit for `0` in most cases.

newdle/client/src/reducers/creation.js
```diff
@@ -147,4 +148,15 @@ export default combineReducers({
         return state;
     }
   },
+  notify: (state = false, action) => {
```
ThiefMaster, line 151: I think this is useful enough to default to `true`

newdle/client/src/reducers/creation.js
```diff
@@ -147,4 +148,15 @@ export default combineReducers({
...
+    switch (action.type) {
+      case ABORT_CREATION:
+      case NEWDLE_CREATED:
+        return false;
```
ThiefMaster, line 155: same here

newdle/migrations/versions/20200914_1600_c161fdcfab19_.py
```diff
@@ -0,0 +1,31 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'c161fdcfab19'
+down_revision = '93a638b96375'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
ThiefMaster, line 20: Please remove these comments :)

newdle/migrations/versions/20200914_1600_c161fdcfab19_.py
```diff
@@ -0,0 +1,31 @@
...
+    op.add_column(
+        'newdles',
+        sa.Column('notify', sa.Boolean(), server_default='false', nullable=False),
```
ThiefMaster, line 23: after adding the column (which applies the default to all existing rows), alter it to change the default to `true` (see the revision adding private for an example)

newdle/migrations/versions/20200914_1640_aa8c14c9e867_.py
```diff
@@ -0,0 +1,28 @@
+"""Add creator email column to db model
```
ThiefMaster, line 1: Both this and the other file lack the title slug in the filename (usually it's set when doing `flask db migrate -m 'title here'`).

newdle/notifications.py
```diff
@@ -27,6 +27,25 @@ def notify_newdle_participants(
...
+            sender_name,
+            participant.email,
+            creator_email,
+            subject,
+            text_template,
+            html_template,
+            get_context(participant),
+        )
+    ]
+    return send_emails(emails)
```
ThiefMaster, line 46: IMO this is cleaner:

```python
email = create_creator_email(...)
return send_emails([email])
```

newdle/notifications.py
```diff
@@ -55,3 +74,28 @@ def create_participant_email(
     )
     msg.attach_alternative(html_content, 'text/html')
     return msg
+
+def create_creator_email(
```
ThiefMaster, line 78: This function is almost identical with `create_participant_email` - see if you can't use one function that can cover both cases

newdle/templates/joined_email.html
```diff
@@ -0,0 +1,9 @@
+<body>
+  <p>Hi,</p>
+  <p>
+    <strong>{{ participant }}</strong> has joined to "<strong>{{ title }}</strong>".
```
ThiefMaster, line 4: see my comment about the email subject - this applies here as well

newdle/templates/joined_email.html
```diff
@@ -0,0 +1,9 @@
...
+  </p>
+  <p>
+    You can get more info using your <a href="{{ answer_link }}"> personal link</a>.
```
ThiefMaster, line 7: I don't think including the user's answer link makes sense here.

Better link to the newdle summary instead. Any maybe we could even include the user's availability information in this email?

ThiefMaster (changes requested):

newdle/api.py
```diff
@@ -356,6 +358,29 @@ def update_participant(args, code, participant_code):
     for key, value in args.items():
         setattr(participant, key, value)
     db.session.commit()
+    formatted_answers = []
+    for date, state in participant._answers.items():
+        formatted_answers.append(
+            "Date: {} ({})".format(
+                date.replace("T", " at "), state.replace("ifneedbe", "if needed")
+            )
```
ThiefMaster, line 366: I don't think this is a good way of doing this.  I would rather use strftime to format the datetime that's in `answers`.

For the state, define a dict mapping state names to human-friendly strings. either here in python or in a Jinja macro used in the templates.

newdle/api.py
```diff
@@ -356,6 +358,29 @@ def update_participant(args, code, participant_code):
     for key, value in args.items():
         setattr(participant, key, value)
     db.session.commit()
+    formatted_answers = []
+    for date, state in participant._answers.items():
```
ThiefMaster, line 362: You can use a list comprehension instead

newdle/api.py
```diff
@@ -356,6 +358,29 @@ def update_participant(args, code, participant_code):
...
+            participant,
+            f'{participant.name} replied to {participant.newdle.title}',
+            'joined_email.txt',
+            'joined_email.html',
+            {
+                'participant': participant.name,
+                'title': participant.newdle.title,
+                'answers': formatted_answers,
+                'summary_link': url_for(
+                    'newdle', code=participant.newdle.code + '/summary', _external=True
```
ThiefMaster, line 380: define a new `build_only` url rule in `create_app` and use that in the `url_for` call here instead of adding path segments to the `code`

newdle/migrations/versions/20200914_1640_aa8c14c9e867__add_creator_email_column_to_newdle_model.py
```diff
@@ -0,0 +1,29 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.add_column('newdles', sa.Column('creator_email', sa.String(), nullable=False))
+    op.alter_column(
+        'newdles',
+        'notify',
+        server_default='',
+    )
```
ThiefMaster, line 25: This shouldn't be here

newdle/migrations/versions/20200914_1640_aa8c14c9e867__add_creator_email_column_to_newdle_model.py
```diff
@@ -0,0 +1,29 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'aa8c14c9e867'
+down_revision = 'c161fdcfab19'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('newdles', sa.Column('creator_email', sa.String(), nullable=False))
```
ThiefMaster, line 20: You need `server_default=''` or it will fail for existing rows.

newdle/models.py
```diff
@@ -39,6 +39,7 @@ class Newdle(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     creator_uid = db.Column(db.String, nullable=False, index=True)
     creator_name = db.Column(db.String, nullable=False)
+    creator_email = db.Column(db.String, nullable=False)
```
ThiefMaster, line 42: `server_default=''`

newdle/templates/joined_email.txt
```diff
@@ -0,0 +1,10 @@
+Hi,
+
+{{ participant }} replied to "{{ title }}".
+
+Selected dates:
+{% for item in answers %}
+    {{ item }}
+{% endfor %}
```
ThiefMaster, line 8: I think you get some ugly extra linebreaks here. `{% for item in answers -%}` may be enough to avoid it, but you need to test this ;)

newdle/templates/joined_email.txt
```diff
@@ -0,0 +1,10 @@
...
+
+You can get more info using the {{ summary_link }} summary link
```
ThiefMaster, line 10: This doesn't look particularly good in a plaintext mail. What about something like this?

```
Check all results on the summary page:
{{ summary_link }}
```

newdle/templates/joined_email.html
```diff
@@ -0,0 +1,15 @@
...
+    <strong>{{ participant }}</strong> replied to "<strong>{{ title }}</strong>".
+  </p>
+  <p>Selected dates:</p>
+  <ul>
+    {% for item in answers %}
+      <li>{{ item }}</li>
+    {% endfor %}
+  </ul>
+  <p>
+    You can get more info using the <a href="{{ summary_link }}"> summary link</a>.
```
ThiefMaster, line 13: Check all results on the <a href="...">summary page</a>.

ThiefMaster (changes requested):

newdle/templates/joined_email.txt
alcalaperez: @ThiefMaster 
```
{% for item in answers %-}
    {{ item }}
{% endfor %-}
```
Returns
```
  <p>Selected dates:</p>
  <ul>
    <li>Date: 18/09/2020, 10:00:00 (Available)</li>
    <li>Date: 18/09/2020, 13:30:00 (If needed)</li>
    </ul>
  <p>
```
Notice the ul closing tag displaced.

While
```
{% for item in answers %-}
    {{ item }}
{% endfor %}
```

Converts into 
```
  <p>Selected dates:</p>
  <ul>
    <li>Date: 18/09/2020, 10:00:00 (Available)</li>
    <li>Date: 18/09/2020, 13:30:00 (Available)</li>

  </ul>
  <p>
```

Notice the extra line. Some ideas?

ThiefMaster: for the HTML version it doesn't matter, since linebreaks are pretty much meaningless in HTML.

only the plaintext version is where all whitespace is significant.

alcalaperez: It is good something like this?

```
{% for item in answers -%}
    • {{ item }}
{% endfor %}
```

returning

> Selected dates:
> • Date: 18/09/2020, 10:00:00 (If needed)
> • Date: 18/09/2020, 13:30:00 (Available)
> 
> Check all results on the summary page:
> ...../newdle/XXXXXX/summary

ThiefMaster: I think I'd prefer simple dashes (`-`), but having some kind of list item makes sense for sure

newdle/api.py
alcalaperez: Interesting feature 🤔

newdle/api.py
```diff
@@ -38,6 +42,11 @@
     UserSearchResultSchema,
 )
 
+statesDict = {
+    Availability.available: "Available",
+    Availability.ifneedbe: "If needed",
+    Availability.unavailable: "Unavailable",
```
ThiefMaster, line 48:
```suggestion
    Availability.available: 'Available',
    Availability.ifneedbe: 'If needed',
    Availability.unavailable: 'Unavailable',
```

newdle/api.py
```diff
@@ -38,6 +42,11 @@
     UserSearchResultSchema,
 )
 
+statesDict = {
```
ThiefMaster, line 45: no camelcase please. `state_titles`?

```suggestion
state_titles = {
```

newdle/api.py
```diff
@@ -356,6 +367,28 @@ def update_participant(args, code, participant_code):
     for key, value in args.items():
         setattr(participant, key, value)
     db.session.commit()
+
+    formatted_answers = [
+        "Date: {} ({})".format(
+            date.strftime("%d/%m/%Y, %H:%M:%S"), statesDict.get(state)
```
ThiefMaster, line 373:
```suggestion
        'Date: {} ({})'.format(
            date.strftime('%d/%m/%Y, %H:%M:%S'), state_titles[state]
```

newdle/api.py
```diff
@@ -356,6 +367,28 @@ def update_participant(args, code, participant_code):
...
+        ) for date, state in args['answers'].items()
+    ]
+
+    if participant.newdle.notify:
+        notify_newdle_creator(
+            participant,
+            f'{participant.name} replied to {participant.newdle.title}',
+            'joined_email.txt',
+            'joined_email.html',
```
ThiefMaster, line 382: I'd rename those to `replied_email.*`

newdle/migrations/versions/20200914_1600_c161fdcfab19_add_notify_column_to_newdle_model.py
```diff
@@ -0,0 +1,32 @@
+"""Add notify column to Newdle model
+
+Revision ID: c161fdcfab19
+Revises: 93a638b96375
```
ThiefMaster, line 4:
```suggestion
Revises: 679eab4aab01
```

this is just metadata for humans, but let's be consistent :)

newdle/notifications.py
```diff
@@ -30,15 +30,35 @@ def notify_newdle_participants(
     return send_emails(emails)
 
 
+def notify_newdle_creator(participant, subject, text_template, html_template, context, attachments=None):
+    if not attachments:
+        attachments = []
+    creator_email = participant.newdle.creator_email
+    sender_name = current_app.config['NOREPLY_ADDRESS']
```
ThiefMaster, line 37: What about using the name of the participant here? Having an email address in the name part of the email is rather strange IMO.

newdle/core/app.py
```diff
@@ -97,4 +97,5 @@ def create_app(config_override=None, use_env_config=True):
     app.add_url_rule('/', 'index', build_only=True)
     app.add_url_rule('/newdle/<code>', 'newdle', build_only=True)
     app.add_url_rule('/newdle/<code>/<participant_code>', 'newdle', build_only=True)
+    app.add_url_rule('/newdle/<newdleCode>/summary', 'newdle', build_only=True)
```
ThiefMaster, line 100: `<code>` to be consistent

newdle/templates/joined_email.txt
```diff
@@ -0,0 +1,10 @@
+Hi,
+
+{{ participant }} replied to "{{ title }}".
+
+Selected dates:
+{% for item in answers -%}
+    - {{ item }}
+{% endfor %}
```
ThiefMaster, line 8: If the user left a comment when answering the newdle including this might be nice as well...

ThiefMaster (changes requested):
Some tests are failing, most likely because we aren't sending a value for `notify` when calling the newdle creation endpoint.

newdle/core/app.py
alcalaperez: @ThiefMaster if it is changed to code, the url_for will go for ` app.add_url_rule('/newdle/<code>', 'newdle', build_only=True)` instead of ` app.add_url_rule('/newdle/<code>/summary', 'newdle', build_only=True)` 

Am I missing a way to specify the last part of the url in the url_for?

ThiefMaster: oh, you have to use a different endpoint name, e.g. `summary` or `newdle_summary`.

(for the first two using the same one is fine because it's clear which one to build from the params)

newdle/templates/joined_email.txt
alcalaperez: Something like this? @ThiefMaster 

`'comments': participant.comment if participant.comment else 'No comments added'`

```
Selected dates:
Hi,

{{ participant }} replied to "{{ title }}".
{{ comments }}

Selected dates:
{% for item in answers -%}
    - {{ item }}
{% endfor %}
```

ThiefMaster: I would only include something if there is a comment. Something like this (example not taking whitespace into account:

```jinja
{% if comment %}
Comment: {{ comment }}
{% endif %}
```

newdle/api.py
```diff
@@ -356,9 +365,32 @@ def update_participant(args, code, participant_code):
                 },
             )
         participant.newdle.update_lastmod()
+        formatted_answers = [
+            'Date: {} ({})'.format(
+                date.strftime('%d/%m/%Y, %H:%M:%S'), states_titles.get(state)
```
ThiefMaster, line 370:
```suggestion
                date.strftime('%d/%m/%Y, %H:%M:%S'), states_titles[state]
```

you never have a state that's not in the titles dict

newdle/migrations/versions/20200922_1759_43f0bf77bf97_add_creator_email_and_notify_columns_to_.py
```diff
@@ -0,0 +1,36 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '43f0bf77bf97'
+down_revision = '239d7862e2fe'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
ThiefMaster, line 20: please remove those :)

newdle/templates/replied_email.html
```diff
@@ -0,0 +1,20 @@
+<body>
+  <p>Hi,</p>
+  <p>
+    <strong>{{ participant }}</strong> replied to "<strong>{{ title }}</strong>".
+  </p>
+  {%- if comment %}
+  <p>
+    Comment: {{ comment }}
+  </p>
```
ThiefMaster, line 9:
```suggestion
    <p>
      Comment: {{ comment }}
    </p>
```

indentation

alcalaperez: @ThiefMaster this converts into:

```
  <p>
    <strong>Guinea Pig</strong> replied to "<strong>adsfasfsdfadsfa</strong>".
  </p>
    <p>
      Comment: asdfsdfasdf
    </p>

  <p>Selected dates:</p>
```

breaks the indentation for some reason

ThiefMaster: it does, but in an html email this doesn't matter!

ThiefMaster (approved):

PR #239, opened 2020-09-16
Clarify invitation possibilities of participants

lgatting:
I have not added the current user automatically as participant since it didn't seem necessary and, to me, it seemed to change the newdle creation flow significantly. Let me know if that is a desired feature which should be implemented.

### DIFF
--- a/newdle/client/src/components/creation/CreationHeader.js
+++ b/newdle/client/src/components/creation/CreationHeader.js
@@ -11,7 +11,8 @@ export default function CreationHeader() {
   const steps = [
     {
       title: 'Choose the participants',
-      description: 'You can also "Skip" this step and add participants later by sharing link to your Newdle',
+      description:
+        'You can also "Skip" this step and add participants later by sharing link to your Newdle',
     },
     {
       title: 'Choose the time slots',
--- a/newdle/client/src/components/creation/CreationHeader.js
+++ b/newdle/client/src/components/creation/CreationHeader.js
@@ -12,7 +12,7 @@ export default function CreationHeader() {
     {
       title: 'Choose the participants',
       description:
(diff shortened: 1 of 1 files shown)


newdle/client/src/components/creation/CreationHeader.js
```diff
@@ -11,7 +11,8 @@ export default function CreationHeader() {
   const steps = [
     {
       title: 'Choose the participants',
-      description: 'You can also "Skip" this step and add them later',
+      description:
+        'You can also "Skip" this step and add participants later by sharing link to your Newdle',
```
ThiefMaster, line 15:
```suggestion
        'You can also "Skip" this step and simply share the link to your newdle',
```

newdle/client/src/components/creation/CreationHeader.js
```diff
@@ -11,7 +11,8 @@ export default function CreationHeader() {
   const steps = [
     {
       title: 'Choose the participants',
-      description: 'You can also "Skip" this step and add them later',
+      description:
+        'You can also "Skip" this step and simply share the link to your newdle',
```
ThiefMaster, line 15:
```suggestion
      description: 'You can also "Skip" this step and simply share the link to your newdle',
```

newdle/client/src/components/creation/CreationHeader.js
lgatting: I'd then suggest to modify the "Finalize your Newdle" to "Finalize your new newdle" to keep it consistent

ThiefMaster: "new newdle" sounds pretty strange ;o

ThiefMaster:
btw i just noticed that in step 3 we use "Newdle" instead of "newdle" - maybe you could fix that as well? :)

PR #240, opened 2020-09-16
Add 'get_busy_times' API test

tomasr8:
Ensures that the given endpoint requires authentication.
The API should respond with `200` for valid auth and `401` when no auth is provided

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -6,6 +6,7 @@ import pytest
 from flask import url_for
 from werkzeug.exceptions import Forbidden
 
+from newdle import api
 from newdle.core.auth import app_token_from_multipass
 from newdle.models import Newdle, Participant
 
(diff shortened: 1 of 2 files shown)


tests/api_test.py
```diff
@@ -159,6 +160,26 @@ def test_create_newdle_participant_email_sending(flask_client, dummy_uid, mail_q
     assert resp.status_code == 200
 
 
+def test_get_busy_times(flask_client, dummy_uid):
+    _get_busy_times = api._get_busy_times
+    api._get_busy_times = Mock(return_value={})
```
ThiefMaster, line 165: Please don't do manual monkeypatching. You can use pytest's [monkeypatch](https://docs.pytest.org/en/stable/monkeypatch.html) fixture. See `test_code_generation` for an example where it's used with mocks

tomasr8: Thanks for the suggestion! This is definitely the better way to do it 👍

tests/api_test.py
```diff
@@ -159,6 +159,28 @@ def test_create_newdle_participant_email_sending(flask_client, dummy_uid, mail_q
     assert resp.status_code == 200
 
 
+def test_get_busy_times(flask_client, dummy_uid, monkeypatch):
+    def mock_get_busy(date, tz, uid):
+        return {}
+
+    from newdle import api
+
+    monkeypatch.setattr(api, '_get_busy_times', mock_get_busy)
```
ThiefMaster, line 168:
```python
monkeypatch.setattr(api, '_get_busy_times', lambda date, tz, uid: {})
```

Or better, use a real mock - that will let you actually assert that it has been called with the proper data from the request! That way the test also becomes more useful than just checking if the access check works.

You can either use `monkeypatch.setattr` for it or better add [`pytest-mock`](https://pypi.org/project/pytest-mock/) to the dev dependencies and use that here.

tomasr8: Thanks again! I'll get on it

ThiefMaster (approved):

ThiefMaster:
Thanks! If you have time, maybe you could also add a test for `get_participant_busy_times`?

tomasr8:
> 
> 
> Thanks! If you have time, maybe you could also add a test for `get_participant_busy_times`?

Yup, on it!

PR #243, opened 2020-09-17
Call Indico API for event creation

### LINKED ISSUE #128: Integration with Indico
This can only be done once we have an endpoint on the Indico side that can receive the data and forward the user to the event creation dialog.

adl1995:
closes #128

Corresponding PR in Indico: https://github.com/indico/indico/pull/4628

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -126,6 +126,12 @@ def user_avatar(payload):
     return Response(resp.content, resp.status_code, headers)
 
 
+@api.route('/config/')
+@allow_anonymous
+def config():
+    return jsonify(has_event_creation=bool(current_app.config['CREATE_EVENT_URL']))
+
+
 @api.route('/ping')
 @allow_anonymous
 def ping():
@@ -284,6 +290,26 @@ def get_newdles_participating():
         .order_by(Newdle.id.desc())
     )
     return NewdleParticipantSchema(many=True).jsonify(newdle)
+
+
+@api.route('/newdle/<code>/create-event', methods=('POST',))
+def create_event(code):
+    newdle = Newdle.query.filter_by(code=code, creator_uid=g.user['uid']).first_or_404(
+        'Specified newdle does not exist'
+    )
+    url = current_app.config['CREATE_EVENT_URL']
+    res = requests.post(
+        url,
+        data={
+            'title': newdle.title,
+            'start_dt': newdle.final_dt.isoformat(),
+            # Indico accepts the duration in minutes.
+            'duration': int(newdle.duration.total_seconds()) // 60,
+            'tz': newdle.timezone,
+        },
+    )
+    res.raise_for_status()
+    return jsonify(url=res.json()['url'])
 
 
 @api.route('/newdle/', methods=('POST',))
--- a/newdle/client/src/components/summary/SummaryPage.js
+++ b/newdle/client/src/components/summary/SummaryPage.js
@@ -48,6 +48,8 @@ export default function SummaryPage() {
     client.sendResultEmails
   );
   const [_setFinalDate, submitting] = client.useBackendLazy(client.setFinalDate);
+  const [_createEvent] = client.useBackendLazy(client.createEvent);
+  const [config, loading] = client.useBackend(() => client.getConfig(), []);
   usePageTitle(newdle && `Summary: ${newdle.title}`, true);
 
   const update = async () => {
@@ -82,6 +84,13 @@ export default function SummaryPage() {
 
   const mailSent = sendMailResponse !== null;
   const editUrl = `/newdle/${newdle.code}/edit`;
+
+  const createEvent = async (event, {value}) => {
+    const resp = await _createEvent(newdle.code);
+    if (resp && resp.url) {
+      window.location.href = resp.url;
+    }
+  };
 
   return (
     <Container text>
@@ -172,11 +181,15 @@ export default function SummaryPage() {
                       <Trans>E-mail participants</Trans>
                     </Button>
                   )}
-                  {/*
-                    <Button className={styles['create-event-button']}>
+                  {config.has_event_creation && (
+                    <Button
+                      className={styles['create-event-button']}
+                      onClick={createEvent}
+                      loading={loading}
+                    >
                       <Trans>Create event</Trans>
                     </Button>
-                  */}
+                  )}
                 </div>
               )}
             </ParticipantTable>
--- a/newdle/newdle.cfg.example
+++ b/newdle/newdle.cfg.example
@@ -97,3 +97,8 @@ LAST_ACTIVITY_CLEANUP_DELAY = None
 FINAL_DATE_CLEANUP_DELAY = None
 # Days after which a newdle that was deleted by its creator is cleaned up.
 DELETED_CLEANUP_DELAY = None
+
+# If set, newdle will show a button to create an Indico event using the title
+# and time from the newdle.  Other applications besides Indico (3.0+) could be
+# used as well as long as they implement the same API as Indico has.
+CREATE_EVENT_URL = ''
--- a/newdle/client/src/client.js
+++ b/newdle/client/src/client.js
@@ -80,7 +80,7 @@ class Client {
   /**
    * `useBackendLazy` works fine for methods that are invoked manually (e.g. clicking a button) but may cause problems
    * when used in conjunction with `useEffect` which relies on the identity of its dependencies. `useBackend` solves
-   * this problem by calling the passed `func` immediatelly and returning relevant data.
+   * this problem by calling the passed `func` immediately and returning relevant data.
    *
    * @param {Function} func - function that will be called every time `deps` change.
    * @param {Array} deps - dependencies passed to the `useEffect`.
@@ -123,6 +123,12 @@ class Client {
 
   searchUsers(name, email) {
     return this._request(flask`api.users`({name, email}));
+  }
+
+  createEvent(code) {
+    return this._request(flask`api.create_event`({code}), {
+      method: 'POST',
+    });
   }
 
   createNewdle(title, duration, timezone, timeslots, participants, isPrivate, notify) {
@@ -242,6 +248,10 @@ class Client {
         comment: comment,
       }),
     });
+  }
+
+  getConfig() {
+    return this._request(flask`api.config`(), {anonymous: true});
   }
 
   async _request(url, options = {}, withStatus = false, isRetry = false) {
--- a/newdle/client/src/locales/es/messages.po
+++ b/newdle/client/src/locales/es/messages.po
@@ -67,7 +67,7 @@ msgstr ""
 
 #: src/components/creation/userSearch/UserSearch.js:146
 #: src/components/login/LoginPrompt.js:45
-#: src/components/summary/SummaryPage.js:146
+#: src/components/summary/SummaryPage.js:155
 msgid "Cancel"
 msgstr ""
 
@@ -98,7 +98,7 @@ msgstr ""
 #: src/components/creation/ParticipantsStep.js:49
 #: src/components/creation/timeslots/TimeslotsStep.js:103
 #: src/components/creation/userSearch/UserSearch.js:143
-#: src/components/summary/SummaryPage.js:143
+#: src/components/summary/SummaryPage.js:152
 msgid "Confirm"
 msgstr ""
 
@@ -118,7 +118,7 @@ msgstr ""
 msgid "Copy to clipboard"
 msgstr ""
 
-#: src/components/summary/SummaryPage.js:176
+#: src/components/summary/SummaryPage.js:190
 msgid "Create event"
 msgstr ""
 
@@ -131,8 +131,8 @@ msgid "Create your newdle!"
 msgstr ""
 
 #: src/components/summary/DeleteModal.js:96
-#: src/components/summary/SummaryPage.js:188
-#: src/components/summary/SummaryPage.js:246
+#: src/components/summary/SummaryPage.js:203
+#: src/components/summary/SummaryPage.js:261
 msgid "Delete newdle"
 msgstr ""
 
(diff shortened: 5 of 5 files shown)


newdle/api.py
```diff
@@ -245,6 +247,21 @@ def get_newdles_participating():
     return NewdleParticipantSchema(many=True).jsonify(newdle)
 
 
+@api.route('/newdle/event', methods=('POST',))
+@use_kwargs(NewEventSchema(), locations=('json',))
+def create_event(title, duration, final_dt):
```
ThiefMaster, line 252: What about POSTing to `/newdle/<code>/create-event` and then taking the data from the newdle's data?

newdle/client/src/components/creation/FinalStep.js
```diff
@@ -29,15 +29,7 @@ export default function FinalStep() {
...
-    const newdle = await _createNewdle(
-      title,
-      duration,
-      timezone,
-      timeslots,
-      participants,
-      isPrivate,
-      notify
-    );
+    const newdle = await _createNewdle(title, duration, timezone);
```
ThiefMaster, line 32: This change doesn't look correct...

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -103,6 +104,11 @@ export default function SummaryPage() {
 
   const mailSent = sendMailResponse !== null;
 
+  const createEvent = async (event, {value}) => {
+    const resp = await _createEvent(newdle.title, newdle.duration, newdle.final_dt);
+    window.open(resp['url'], '_blank');
```
ThiefMaster, line 109: - Isn't `_blank` the default?
- I think this may trigger popup blockers depending on how long the request takes. I'd just redirect tbh.
- `resp.url`

newdle/api.py
adl1995: Do you mean to create the UUID (`<code>`) here then pass it on to /newdle/<code>/create-event (I'm not sure if that will be useful)? Can you clarify this?

newdle/client/src/components/creation/FinalStep.js
adl1995: Reverted, thanks.

newdle/client/src/components/summary/SummaryPage.js
adl1995: It indeed is. Updated `resp['url']` to `resp.url`.

adl1995:
@ThiefMaster - I don't think the failing test case is coming from my code.

newdle/client/src/locales/es/messages.js
```diff
@@ -0,0 +1 @@
... was created and invitation e-mails have been sent. You can send the following link to everyone you would like to invite:","Your newdle was created. You can now send the following to everyone you would like to invite:":"Your newdle was created. You can now send the following to everyone you would like to invite:","newdle is Open Source Software":"newdle es Software de c\xF3digo abierto","newdle logo":"newdle logo","originally created in the <0>{newdleTz}</0> timezone":["originally created in the <0>",["newdleTz"],"</0> timezone"],"switch back to the <0>{defaultUserTz}</0> timezone":["switch back to the <0>",["defaultUserTz"],"</0> timezone"],"{0, plural, one {# participant will be e-mailed:} other {# participants will be e-mailed:}}":[["0","plural",{one:["#"," participant will be e-mailed:"],other:["#"," participants will be e-mailed:"]}]],"{0} will take place on:":[["0"]," will take place on:"],"{numberOfTimeslots, plural, one {{numberOfAvailable} out of # option chosen} other {{numberOfAvailable} out of # options chosen}}":[["numberOfTimeslots","plural",{one:[["numberOfAvailable"]," out of ","#"," option chosen"],other:[["numberOfAvailable"]," out of ","#"," options chosen"]}]]};
```
ThiefMaster, line 1: this file shouldn't be here. run `make` since you probably have an older version of lingui that creates it (newer versions use .mjs which is on gitignore)

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -48,6 +48,8 @@ export default function SummaryPage() {
     client.sendResultEmails
   );
   const [_setFinalDate, submitting] = client.useBackendLazy(client.setFinalDate);
+  const [_deleteNewdle, deleting] = client.useBackendLazy(client.deleteNewdle);
```
ThiefMaster, line 51: Looks like something went wrong during your rebase - this was moved elsewhere.

newdle/schemas.py
```diff
@@ -172,6 +172,11 @@ def validate_timeslots(self, v):
 
 class UpdateNewdleSchema(NewNewdleSchema):
     participants = fields.List(fields.Nested(NewParticipantSchema, unknown=EXCLUDE))
+
+
+class NewEventSchema(mm.Schema):
+    title = fields.String(validate=validate.Length(min=3, max=80), required=True)
+    duration = fields.Integer()
     final_dt = fields.DateTime(format=DATETIME_FORMAT)
```
ThiefMaster, line 180: Look closely at the change here:  you are basically removing the `final_dt` from `UpdateNewdleSchema` which is what's breaking the tests. Add it back and the tests should pass. And maybe rename it to `start_dt` in here since this is more descriptive of what it is used for in NewEventSchema.

newdle/api.py
```diff
@@ -291,6 +292,21 @@ def get_newdles_participating():
...
+    url = current_app.config['INDICO_URL']
+    res = requests.post(
+        url,
+        data={
+            'title': title,
+            'start_dt': final_dt,
+            'duration': duration,
+        },
+    )
+    return res.text
```
ThiefMaster, line 307: Add `res.raise_for_status()` to fail if the request is not successful.

ThiefMaster, line 307: You are returning `text/html` here atm. I'd rather `return jsonify(url=res.text)` here. Or even better, make the Indico API also return JSON and then use `res.json()['url']`

newdle/api.py
ThiefMaster: This API is not specific to any newdle atm - what I meant is to get rid of `NewEventSchema` altogether and use the `final_dt` and `duration` of the newdle you can get through its code.

newdle/client/src/locales/es/messages.js
adl1995: Removed.

newdle/client/src/components/summary/SummaryPage.js
adl1995: Removed.

newdle/schemas.py
adl1995: Ah, I missed that! (I've now removed the `NewEventSchema`.)

newdle/api.py
adl1995: Thanks, added.

adl1995: Thanks, I'm returning `res.json()` now and accessing the URL with `resp.url` as before.

newdle/api.py
adl1995: That's much cleaner, thanks! I'm now fetching all fields (`title`, `duration`, `final_dt`) from the Newdle object.

newdle/newdle.cfg.example
```diff
@@ -97,3 +97,6 @@ LAST_ACTIVITY_CLEANUP_DELAY = None
 FINAL_DATE_CLEANUP_DELAY = None
 # Days after which a newdle that was deleted by its creator is cleaned up.
 DELETED_CLEANUP_DELAY = None
+
+# The INDICO_URL is used for creating an event from a Newdle. 
+INDICO_URL = ''
```
ThiefMaster, line 102: missing trailing linebreak

newdle/newdle.cfg.example
```diff
@@ -97,3 +97,6 @@ LAST_ACTIVITY_CLEANUP_DELAY = None
 FINAL_DATE_CLEANUP_DELAY = None
 # Days after which a newdle that was deleted by its creator is cleaned up.
 DELETED_CLEANUP_DELAY = None
+
+# The INDICO_URL is used for creating an event from a Newdle.
```
ThiefMaster, line 101: # If set, newdle will show a button to create an Indico event using data from the newdle.

ThiefMaster: We should hide that button btw if no URL is set.

newdle/api.py
```diff
@@ -291,6 +291,25 @@ def get_newdles_participating():
     return NewdleParticipantSchema(many=True).jsonify(newdle)
 
 
+@api.route('/newdle/<code>/create-event', methods=('POST',))
+def create_event(code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404(
+        'Specified newdle does not exist'
+    )
```
ThiefMaster, line 298: I think we can require that the user is the creator of the newdle. No harm in any case, but feels dirty to not restrict this API.

newdle/api.py
```diff
@@ -291,6 +291,25 @@ def get_newdles_participating():
...
+    url = current_app.config['INDICO_URL']
+    res = requests.post(
+        url,
+        data={
+            'title': newdle.title,
+            'start_dt': newdle.final_dt,
+            # Newdle accepts the duration in minutes.
```
ThiefMaster, line 305: Don't you mean `Indico`?

newdle/api.py
```diff
@@ -291,6 +291,25 @@ def get_newdles_participating():
...
+            'duration': (newdle.duration.seconds // 60) % 60,
```
ThiefMaster, line 306: `int(newdle.duration.total_seconds()) // 60` and I the modulus isn't needed - we can have meetings longer than 60 minutes! ATM this actually returns 0 if it's exactly 1 hour...

newdle/newdle.cfg.example
ThiefMaster, line 101: IIRC we said we wanted to make this generic at some point, ie `CREATE_EVENT_URL` with no mention of Indico or just as an example.

> If you use Indico, the URL to use is https://yourindico/event/prepare

Or something like that.

newdle/api.py
```diff
@@ -291,6 +291,25 @@ def get_newdles_participating():
...
+            'tz': newdle.timezone,
+        },
+    )
+    return res.json()
```
ThiefMaster, line 310:
```python
res.raise_for_status()
return jsonify(url=res.json()['url'])
```

that way we fail more clearly if we get bad data back from the server

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -83,6 +84,11 @@ export default function SummaryPage() {
   const mailSent = sendMailResponse !== null;
   const editUrl = `/newdle/${newdle.code}/edit`;
 
+  const createEvent = async (event, {value}) => {
+    const resp = await _createEvent(newdle.code);
+    window.open(resp.url);
```
ThiefMaster, line 89: This triggers the popup blocker. It also fails badly if `resp` is undefined due to an error. Easiest solution would be this:

```js
if (resp && resp.url) {
  location.href = resp.url
}
```

newdle/api.py
```diff
@@ -291,6 +291,25 @@ def get_newdles_participating():
...
+def create_event(code):
+    newdle = Newdle.query.filter_by(code=code).first_or_404(
+        'Specified newdle does not exist'
+    )
+    url = current_app.config['INDICO_URL']
+    res = requests.post(
+        url,
+        data={
+            'title': newdle.title,
+            'start_dt': newdle.final_dt,
```
ThiefMaster, line 304: `.isoformat()`

ThiefMaster (changes requested):

newdle/client/src/components/summary/SummaryPage.js
adl1995: This results in this error:
```
/home/adeel/software/newdle/newdle/client/src/components/summary/SummaryPage.js
  90:7  error  Unexpected use of 'location'  no-restricted-globals
```
Should I suppress it?

ThiefMaster: yes, perfectly valid here...

adl1995: Thanks, had to use `window.location.href` otherwise got an error when committing.

newdle/api.py
adl1995: Updated the query to `Newdle.query.filter_by(code=code, creator_uid=g.user['uid'])...`.

newdle/api.py
adl1995: Ah, right. Thanks.

newdle/api.py
adl1995: Didn't realize that before, thanks.

newdle/newdle.cfg.example
adl1995: Alright, updated to `CREATE_EVENT_URL`.

newdle/api.py
adl1995: Thanks, updated.

newdle/api.py
ThiefMaster: explicit check and raising `Forbidden` would be cleaner IMO. i think we do that in other places as well...

newdle/newdle.cfg.example
adl1995: I'm not sure how to hide the button. Is there a way to import the config in `.js` files?

ThiefMaster: hm i thought we already had an endpoint to return settings but we don't..

so you need to add something like this:

```
@api.route('/config/')
@allow_anonymous
def config():
    return jsonify(has_event_creation=bool(current_app.config.CREATE_EVENT_URL))
```

and then call it to determine whether to show the button or not

adl1995: Added, thanks.

ThiefMaster (approved):
Nice PR! Works fine :)

Issue #214, opened 2020-08-25
Create basic project site


pferreir:
Should include:
 * Project description
 * Main features
 * Link to install instructions

ParthS007:
Do we have any preference for the theme(like Jekyll or something similar)?

ThiefMaster:
I don't think so... something that works out of the box with GH pages would be great though!

Caetan95:
Hey! I am new here. Michal told me about the sprint (I am not registered) and I would like to collaborate. I think I can try to do this part with GH pages.

ThiefMaster:
Not sure if @ParthS007 started with it already, or if you two may want to work together on it (not sure how to best split it on two people though). Best to check with him directly :)

Caetan95:
I have talked with @ParthS007 and he didn't start it yet, so I can deal with this issue :)

pferreir:
Guys, thanks for your interest in this. I just want to add that we don't need anything very complicated at this point: just something which is easy to update and which doesn't look too bad.

Caetan95:
Yeap, pretty and simple

Caetan95:
How can I find the Link to install instructions?

ThiefMaster:
It's in the README of the project itself. But I don't think you need to install newdle for this task..

pferreir:
> It's in the README of the project itself. But I don't think you need to install newdle for this task..

The ticket does mention he needs to link to the install instructions, though 😉

Caetan95:
> > It's in the README of the project itself. But I don't think you need to install newdle for this task..
> 
> The ticket does mention he needs to link to the install instructions, though wink

yeap, that's why I asked haha I understand that install instructions are just the Development section of the README, just to be sure

Caetan95:
I'm having a problem to make the github page reachable. As far as I know, to be able to make a webpage with GH I have to create a repo called "owner_name.github.io", but of course is not possible due to there is no newdle owner. I have several ideas to deal with it:

1. To copy the generated HTML code and create another webpage separated from github (ugly idea)
2. To create a newdle account in github to have a newdle owner and be able to create the page (also ugly idea)
3. To include in https://getindico.io (indico's github page) a section for newdle

Any idea about what should I do?

This is the page I did, but is in my repo, so if it is necessary to update nobody can do it

[site](https://caetan95.github.io)

ThiefMaster:
[...]

PR #244, opened 2020-09-18
Create basic project site #214

Caetan95:
_config.yml URL and baseurl lines are commented, UNCOMMENT them once it is in Indico's repo!!


ThiefMaster:
You have some commits with a wrong email (`caetan@caetanpc.dyndns.cern.ch`), but since we'll probably squash anyway and cherry-pick into a new gh-pages branch (that has no ancestry from master) this shouldn't be a problem in the end.

Could you enable gh-pages on your fork though so we can preview it via https://caetan95.github.io/newdle?

ThiefMaster:
Regarding the content of the site: I don't think repeating what we have in the README again in detail is a good idea - I would rather focus on what newdle is, and then link to the readme for the install docs

Caetan95:
I am teleworking connected via ssh to my CERN's computer, so the wrong email was because of that (I don't know why though). I did the changes requested, so now is reachable via [https://caetan95.github.io/newdle](https://caetan95.github.io/newdle ) and there is a new content of the site explaining what newdle is

ThiefMaster:
Please rebase your `gh-pages` branch on top of the newdle upstream/gh-pages branch (this just adds an empty config file since a branch cannot be completely empty) and send a PR against the newdle gh-pages branch. That way we can  actually comment on your latest version.

PR #245, opened 2020-09-18
[Answers] Highlight selected day

### LINKED ISSUE #158: Answer page: highlight currently selected day
Right now the currently selected day in the calendar is not highlighted on the right side. This can create confusion when using the arrows and the calendar to navigate. We should somehow indicate the "active" day, even if it doesn't have many consequences in pretty much anything else.

image

VirtualEvan:
closes #158 

This change indicates on the right side the day that has been selected in the Month Calendar.

image

### DIFF
--- a/newdle/client/src/components/answer/Calendar.js
+++ b/newdle/client/src/components/answer/Calendar.js
@@ -18,6 +18,7 @@ import {
 } from '../../answerSelectors';
 import {setAnswer, setAnswerActiveDate} from '../../actions';
 import DayCarousel from '../DayCarousel';
+
 import styles from './answer.module.scss';
 
 const OVERFLOW_HEIGHT = 0.5;
@@ -224,8 +225,9 @@ export default function Calendar() {
   );
   const numDaysVisible = isTabletOrMobile ? 1 : 3;
   const numColumns = isTabletOrMobile ? 14 : 5;
+  const isActiveDay = date => activeDate.isSame(date);
   return (
-    <Grid className={styles.calendar}>
+    <Grid>
       <Grid.Row>
         <Grid.Column width={1}>
           <Hours minHour={minHour} maxHour={maxHour} />
@@ -240,6 +242,7 @@ export default function Calendar() {
               <DayTimeline
                 options={item}
                 busySlots={busyByDay.find(busySlot => busySlot.date === item.date)}
+                selected={isActiveDay(item.date)}
               />
             </Grid.Column>
           )}
--- a/newdle/client/src/components/answer/answer.module.scss
+++ b/newdle/client/src/components/answer/answer.module.scss
@@ -6,15 +6,23 @@
   text-align: center;
 }
 
-.date {
+.date:global(.ui.header) {
   text-align: center;
+  padding-bottom: 5px;
+  padding-top: 5px;
+
+  &.date-selected {
+    background-color: $turquoise;
+    color: $white;
+    border-radius: 3px;
+  }
 }
 
 .hours-column {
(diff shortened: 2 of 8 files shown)


VirtualEvan:
I tryied to follow good practices, but maybe is a bit of an overkill and would be better to simplify like
```javascript
className={
  activeDate.isSame(item.date)
    ? `${styles['day-column']} ${styles['day-column-selected']}`
    : styles['day-column']
}
```
Let me know if any change is required :)

newdle/client/src/components/answer/Calendar.js
```diff
@@ -19,6 +19,7 @@ import {setAnswer, setAnswerActiveDate} from '../../actions';
 import DayCarousel from '../DayCarousel';
 import styles from './answer.module.scss';
 import {getUserTimezone} from '../../selectors';
+import classNames from 'classnames';
```
ThiefMaster, line 22: If you want to start using it, explicitly `npm install --save classnames` inside `newdle/client` so it gets added to `newdle/client/package.json` and we don't just use it because something else already install it.

ThiefMaster:
In general I think it looks very nice, but I would use a different and more neutral color that is not so similar to busy periods.

newdle/client/src/components/answer/Calendar.js
ThiefMaster: I just saw your comment - I don't have a strong opinion on whether to use `classNames` or not

VirtualEvan:
I have been testing different colors of the palette, but all of them end to be confusing at some point when there are multiple answers/periods + overlapping in a single column.

In my oppinion, these are the best ones. What do you think?

blue-0 1 light-grey-0 6 purple-0 10

Leats:
From these choices I like the blue and the grey most!

Though I've been thinking, maybe it is enough to just highlight the date itself and not the whole column? Like this (but maybe blue(?)):
Bildschirmfoto 2020-09-21 um 12 45 40
What do you think? Would that be noticeable enough or is highlighting the whole day more understandable?

VirtualEvan:
Highlighting just the date seems cleaner to me.
However, I feel exactly the same. I am not sure if it will be noticeable enough for the user or just another square in the column :/

Leats:
> However, I feel exactly the same. I am not sure if it will be noticeable enough for the user or just another square in the column :/

I think if the colors are bright enough it could work. (I am not sure how it would look, but maybe we could pick up the turquoise from the datepicker?) Also, I think having it clean like this is better than potentially confusing the user with too many colors laid over each other.

ThiefMaster:
yeah, sounds like a good idea - and it doesn't add noise to the area below

newdle/client/src/components/answer/Calendar.js
VirtualEvan: I just pushed this change

newdle/client/src/components/answer/answer.module.scss
```diff
@@ -6,15 +6,23 @@
...
 }
 
-.date {
+.date:global(.ui.header) {
   text-align: center;
+  padding-bottom: 5px;
+  padding-top: 5px;
+
+  &.date-selected {
+    background-color: rgba($turquoise, 1);
```
ThiefMaster, line 15: isn't that the same as just `background-color: $turquoise`? (opacity 1 is fully visible)

VirtualEvan: uh, yep, too many tests

ThiefMaster (approved):

PR #246, opened 2020-09-20
Add docker-compose files

### LINKED ISSUE #215: Add docker-compose.yml and instructions for Docker deployment
There is already a Dockerfile, but it would be great to provide a way for people to quickly test the system.

carantunes:
Closes #215 

Changes: 
- Add docker-compose
- Add some commands to make
- Update README


carantunes:
Need some feedback: 
Is this docker-compose mean't for development?

The one added mimics the production dockerfile, it's not really mean't for development bc it's not mapping the host root folder to the container.
For development I would probably need a new dockerfile, or to change the existing one. Correct me if I'm wrong but it looks like we're not copying the root folder to inside the container in the dockerfile.
And we would probably prefer to go with 2 containers, one running the `flask-server` and another `react-server`.

docker/development/newdle.cfg.docker
```diff
@@ -0,0 +1,92 @@
+# A random string that is secret.
+SECRET_KEY = 'jfAvAzpJRRfzk7ceanJzJpEIGqa3IzMs'
```
ThiefMaster, line 2: i wouldn't put an actual value in there but generate one ;)

ThiefMaster:
I think it's mostly meant for a quick "i want to see this thing in action".

If we want to support running a development version we should probably indeed map the folders (and make sure it runs as the same uid as your user) and (as you said) use separate containers for the flask and react servers... Maybe with Hacktoberfest coming up soon supporting a full-docker way to run it (while still having editable code locally) would be a nice thing to have.

Makefile
```diff
@@ -120,3 +120,27 @@ build:
 docker:
 	@printf "  \033[38;5;154mDOCKER\033[0m  \033[38;5;176mBuilding production docker image\033[0m\n"
 	@docker build -t newdle .
+
+
+################ Docker-compose commands ################
+
+env:
```
ThiefMaster, line 127: I would prefix all the makefile targets related to docker with `docker`...

Do we need anything except these btw?

- `make docker-run`: `docker-compose up` (without `-d`, this removes the need for the `logs` command)
- `make docker-clean`: `docker-compose down` and whatever else is needed to clean up the docker things
- `make docker-shell`: same as your `shell-env`

I've seen some packages (iirc mattermost does that for example) that run the docker stuff in the backgrond, but I think it makes it *very* easy to miss important error/warning output. I'm not sure how well foregrounded docker-compose with multiple container works. But if the output is readable it should be fine...

docker/development/newdle.cfg.docker
carantunes: well this is just for the sample docker-compose maybe a `secret` or `change-me` value would be more clear. Wdyt?

Makefile
carantunes: I usually do it this way, but I usually use the compose environment for development, so maybe not all is useful indeed.  I think you suggestion will be just fine, I'll update it. Regarding docker-compose in the foreground I never had any issues.

carantunes:
> I think it's mostly meant for a quick "i want to see this thing in action".
> 
> If we want to support running a development version we should probably indeed map the folders (and make sure it runs as the same uid as your user) and (as you said) use separate containers for the flask and react servers... Maybe with Hacktoberfest coming up soon supporting a full-docker way to run it (while still having editable code locally) would be a nice thing to have.

Do you wish me to work on it in this PR? Maybe a `docker-compose.development.yml`

pferreir:
I agree with @ThiefMaster, it's mostly to be able to run it quickly, surely not for production. @carantunes if you can/want to work in that, it would certainly be great!

docker/development/.env.postgres
```diff
@@ -0,0 +1,3 @@
+POSTGRES_USER=user
```
ThiefMaster, line 1: Since we never expect users to change this when testing I would just hardcode all 3 env vars in the compose file (but maybe use `newdle` instead of `user`

carantunes:
> I agree with @ThiefMaster, it's mostly to be able to run it quickly, surely not for production. @carantunes if you can/want to work in that, it would certainly be great!

I added a first working draft. Wdyt? Do you want me to extract some things? Maybe the entrypoints to scripts or the env variables..

docker-compose.development.yml
```diff
@@ -0,0 +1,81 @@
...
+      flask-server:
+        condition: service_started
+    working_dir: /newdle/client
+    environment:
+      - FLASK_URL=http://flask-server:5000
+    entrypoint:
+        - bash
+        - -c
+        - |
+            curl -sL https://deb.nodesource.com/setup_12.x | bash -
```
plourenco, line 57: I think we can upgrade this to node 14, as the node-sass version we're using doesn't seem to support it (4.14+):
```suggestion
            curl -sL https://deb.nodesource.com/setup_14.x | bash -
```

plourenco (approved):
Looks good to me! Just added a few comments regarding `node-sass` failures on my side.
Thank you 👍

README.md
```diff
@@ -68,6 +68,20 @@ We provide a couple of additional `make` targets that should streamline the deve
  - `make test` - runs Python and React tests
  - `make build` - builds a Python wheel which then could be used to install `newdle` in production
 
+## Develop with docker-compose [beta]
+
+ - `make docker-dev-run` - Build and runs the containers. Once everything is running, you can access the webapp on `http://127.0.0.1:3000`
+ - `make docker-dev-clean` - Stops and destroys the images and containers
+ - `make docker-dev-shell-react` - Bash the `react-server` container
+ - `make docker-dev-shell-flask` - Bash the `flask-server` container
```
ThiefMaster, line 76: "Bash the ..." sounds funny ;) Poor containers!

docker-compose.development.yml
```diff
@@ -0,0 +1,81 @@
+version: '3.7'
+
+services:
```
ThiefMaster, line 3: I don't see any `uid` in there. Is this currently running as root?

docker-compose.development.yml
```diff
@@ -0,0 +1,81 @@
...
+      - FLASK_URL=http://flask-server:5000
+    entrypoint:
+        - bash
+        - -c
+        - |
+            curl -sL https://deb.nodesource.com/setup_12.x | bash -
+            apt-get install -y nodejs
+            npm install
+            npm start --loglevel verbose
+    user: root
```
ThiefMaster, line 61: I do not like the idea of running these containers as root...

ThiefMaster: I guess it's needed to install nodejs... can we somehow avoid running the dev server itself as root afterwards?

docker/development/newdle.cfg.docker
```diff
@@ -0,0 +1,91 @@
+# A random string that is secret.
```
ThiefMaster, line 1: What do you think about using a minimal config file that only sets stuff that's actually needed (ie no SSO, no exchange config, random availability data, etc) and different from the default and also without all the comments?

I don't want to have another place to update whenever the example config changes.

plourenco: We could also point to the default one, and let users copy it and create a new one. As this one is also committed, I'm not sure how useful it becomes.

ThiefMaster: I guess for a development instance it doesn't really matter... The only alternative to a hardcoded one I see is generating a random one like we do during the non-docker `make`.

I think the whole idea behind being able to run a dev setup in Docker is that you can quickly do some basic development and testing (we still need a command to run the equivalent of `make lint test` in docker!) without having to do much setup... so having to manually create a config kind of defeats this goal.

docker-compose.development.yml
carantunes: Are you sure? your dockerfile is using 12, it may lead to inconsistencies

docker/development/newdle.cfg.docker
carantunes: I changed the config file. Wdyt? I think it's the minimum required to be able to run.

docker-compose.development.yml
carantunes: Not that I'm aware. But I've always used root as the standard user when developing, since this is not meant to be deployed anywhere and it's common to need to install thinks when bash the containers

docker/development/.env.postgres
carantunes: now it's used in both the docker-compose files

docker-compose.development.yml
ThiefMaster: feels super dirty to me but if others do it as well so be it ;)

let's just make sure that running the dockerfile does not write any root-owned file outside the container, volumes are fine of course, but i would for example not want it to create a `node_modules` in the source folder outside the container that's owned by root

carantunes:
@ThiefMaster @plourenco This has been in the limbo for a while. 😞  Do you need any more changes from me?

plourenco:
Sorry for the delay, some of these changes required further testing and we were distracted with Indico.
Thanks @carantunes!! 🐳

carantunes:
Great @plourenco! 🥳

Issue #242, opened 2020-09-17
Add footer


pferreir:
We should have a footer with a link to the actual project repo/site and optionally a privacy policy or any other legal information (configurable).

TimSmithCH:
Also, the footer should additionally contain a link to a "Terms of Use" (which we can draft from the Indico/Zenodo ones. And also a link to "Support" probably

pferreir:
> By "optionally" I guess you mean in the code-base for service owners to chose...

Yes, that's what I mean 😃

ThiefMaster:
Not sure if "Support" is needed - the amount of support requests we get for newdle is very low. But if we make the footer links configurable adding one would be easily possible.

ThiefMaster:
Let's also find a nice way to show the `LanguageSelector` in the footer.

Note: Since there are no translations available yet, it's hidden by default - add `#i18n` to the URL to show it.

VirtualEvan:
I can start working on this

VirtualEvan:
I made a couple of designs in Figma trying to follow some of the ideas you mention. What do you think?

ThiefMaster:
hm both look a big bulky to me :/

I prefer the one with the gradient though. Maybe we can distribute some things horizontally instead of vertically?

VirtualEvan:
I added new designs with horizontal distribution. There are also some color ideas for the elements and the language selector, trying to match with the rest of the web

ThiefMaster:
6 is my favorite, but some things I'd change:

- The octocat doesn't really fit there - newdle has nothing to do with Git/GitHub except that development happens there. If we want to keep a "Contribute now" (maybe too much?) then the logo should be only next to that
- Maybe instead of "Contribute now" we could put "made at CERN" there? 
- Putting a newdle logo there instead could work, but I'm not sure if that's not too much since we already have it in the header

(but let's see what the others think tomorrow!)

Caetan95:
Loving also 6! Maybe to add the Indico's logo, maintained by Indico Team or something like this could be nice? 😄

VirtualEvan:
Added a new page following the proposed ideas.
- Put "Made at CERN" as it is a representation of the organization and also solves the octocat location "problem"
- I would say the Newdle logo would be to much as the header and footer are both visible at the same time in most of the screens
- Not sure if the Indico logo/info fits there as they are non-related projects. It would have no meaning for people outside CERN who doesn't know Indico at all

[...]

PR #247, opened 2020-09-21
Add i18n

### LINKED ISSUE #136: Implement i18n
Still to see whether we want to go for an Indico-like solution (with [react-jsx-i18n](https://github.com/indico/react-jsx-i18n)) or to try out something like [react-intl](https://github.com/formatjs/react-intl).

PatriciaMaPe:
Closes #136 
Use lingui in two Components to check the i18n use.
As it was mentioned in the issue, this commit adds some example translations in two components to be reviewed. If it is ok with you I will continue working on a component to handle the change of the language in the web page.


newdle/client/src/components/home/HomePage.js
```diff
@@ -17,20 +18,26 @@ export default function HomePage() {
...
-            <span>Set the time slots based on their availability</span>
+            <span>
+              <Trans>Set the time slots based on their availability</Trans>
+            </span>
           </li>
           <li>
             <img src={answersIcon} alt="" className={styles.icon} />
             <span>
-              <strong>newdle</strong> will collect the answers!
+              <strong>newdle</strong> <Trans>will collect the answers!</Trans>
```
ThiefMaster, line 40: Can we somehow include newdle as a placeholder in the string? While it should never be translated, concatenating a translatable string won't work in any language where newdle isn't the first word in this sentence?

If it's not possible let's just include the `<strong>newdle</strong>` in the trans and hope translators don't try translating it ;)

newdle/client/src/locales/en/messages.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable */ module.exports = {
```
ThiefMaster, line 1: Is this a generated file? If yes, we should probably not include it in the repo and gitignore it.

newdle/client/src/locales/en/messages.po
```diff
@@ -0,0 +1,58 @@
...
+msgid "Welcome to newdle!"
+msgstr "Welcome to newdle!"
+
+#: src/components/home/HomeHeader.js:15
+msgid "You can use it to find out the best dates/times for your meetings."
+msgstr "You can use it to find out the best dates/times for your meetings."
+
+#: src/components/home/HomeHeader.js:11
+#~ msgid "is a collective meeting scheduling application."
+#~ msgstr "is a collective meeting scheduling application."
```
ThiefMaster, line 50: This probably shouldn't be in here anymore ;)

newdle/client/src/locales/es/messages.js
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable */ module.exports = {
```
ThiefMaster, line 1: same as for the english file

newdle/client/src/components/home/HomePage.js
PatriciaMaPe: If you are going to translate it with an automatic translator, I can create a variable named with numbers like _0, and use it in the string. 
In that way, the languages that not use 'newdle' as the first word could be correctly translated.
If I write the code like this: 
```js
const _0 = 'newdle';
<Trans><strong>{_0}</strong> is a collective meeting scheduling application.</Trans>
```
it will generate this:
`"<0>{_0}</0> is a collective meeting scheduling application."`
So you can move the variable inside the numbered tag wherever you want, for example, you could translate it like this:
`"A collective meeting scheduling application: <0>{_0}</0>"`
and it will generate:
`A collective meeting scheduling application: <strong>newdle</strong>`
Is this what you mean? :)

ThiefMaster: > If you are going to translate it with an automatic translator

we'll never do that :) I was more thinking about someone accidentally thinking it's a word that should be translated

I guess `<Trans><strong>{'newdle'}</strong> is a collective meeting scheduling application.</Trans>` won't work? In that case I'd just keep `newdle` as-is and assume translators realize that the product's name should not be translated

newdle/client/src/locales/en/messages.js
ThiefMaster: maybe check if we can automatically build it during `make react-server` (could just be done in the makefile before launching the actual react server - i don't think we need auto rebuilding there as long as the english locale works even if strings are missing in here) and of course also in `make build`

PatriciaMaPe: yes! I will add it to the gitignore then

newdle/client/.gitignore
```diff
@@ -18,6 +18,9 @@
 .env.test.local
 .env.production.local
 
+#i18n
```
ThiefMaster, line 21:
```suggestion
# i18n
```

newdle/client/package.json
```diff
@@ -28,7 +29,11 @@
     "semantic-ui-react": "^1.2.0"
   },
   "devDependencies": {
+    "@babel/core": "^7.11.6",
     "@jedmao/redux-mock-store": "^3.0.5",
+    "@lingui/cli": "^2.9.2",
+    "@lingui/macro": "^2.9.2",
```
ThiefMaster, line 35: What do you think about using lingui 3?

https://lingui.js.org/releases/migration-3.html

ThiefMaster: 3.0 docs are here: https://js-lingui-git-next.lingui-js.now.sh/

PatriciaMaPe: When I was through the installation setup, I didn't add any version so it installed the latest one. Maybe the lingui 3 is still in beta?
I had a look into the versions available and it's not the latest one.
https://www.npmjs.com/package/@lingui/cli
https://www.npmjs.com/package/@lingui/macro

ThiefMaster: Yeah, looks like it's only in `@next`... not super clear from their website nor issue tracker...

Let's see if they recommend using 2.x or 3.x in new projects: https://github.com/lingui/js-lingui/issues/334#issuecomment-697811138

PatriciaMaPe:
I have already created a component with a Select to choose the language. It's working now but I wonder where should I put the Select. I have been testing it in the footer of the HomePage like in the screenshot:
image

PatriciaMaPe:
Should I add to the Readme how to add a new language to the project?

newdle/client/package.json
ThiefMaster: some guy in the issue tracker:

> I've asked the same question and was recommended to use 3.0.0-13

ThiefMaster:
documenting it sounds like a good idea, but i'm not sure if README is the best place - it's not as important as how to get newdle running for example... not sure where's the best place to put it tbh. could add it to the description if this PR as well, moving it to some place in the docs is easily done later if needed

ThiefMaster:
About the language picker location: Not sure what's the best place. I think down there footer is OK for now until we have a proper footer (#242).

BTW, I'd try to select a good default language based on `navigator.languages` (this is basicallt the client-side version of the `Accept-Language` header)

newdle/client/package.json
PatriciaMaPe: Ok, I will try with lingui 3 then :)

ThiefMaster: due to lingui/js-lingui#707 (mentioned in a lter comment) it may be safer  to stick with 2 for now :) also since hacktoberfest starts soon: if we can get this PR ready and merged soon, we can open individual issues to add i18n to the various other areas, and benefit from people looking for small hacktoberfest issues that don't require much effort (ie no UI or new feature development)

PatriciaMaPe: ok, perfect! I have planned to push it today and make the needed changes before Hacktoberfest :)

PatriciaMaPe:
The lingui configuration will be in `package.json`:

```jsx
"lingui": {
    "localeDir": "src/locales/", // where the translations will be stored
    "srcPathDirs": [ // where the text will be taken to be translated
      "src/" 
    ],
    "format": "po", // format of the text to be translated
    "sourceLocale": "en" // default language
  },
```

All the translations will be stored in the directory `src/locales/<locale>`. Here, we will have two different files:

- messages.po: we will add the translations here.
- messages.js: the app will use this file to load the translations.

E.g: English language would have this files:

- `src/locales/en/messages.po`
- `src/locales/en/messages.js`

The component I18nProvider, which we use in the component App, sets the active language and loads the translations. The users will select the active language using the LanguageSelector component.

### How to add a new language

Run `npm run add-locale <locale>` with the locale code of the language you want to add to the app:

```jsx
$ npm run add-locale fr // Add french locale

Added locale fr.

(use "npm run extract" to extract messages)
```

Check the installation by running `npm run extract`. You will see something like this (with the other languages you have in the app):

```jsx
$ npm run extract

Catalog statistics:
┌─────────────┬─────────────┬─────────┐
│ Language    │ Total count │ Missing │
├─────────────┼─────────────┼─────────┤
│ en (source) │     10      │    -    │
│ es          │     10      │    0    │
│ fr          │     10      │   10    │
└─────────────┴─────────────┴─────────┘

(use "npm run add-locale <locale>" to add more locales)
(use "npm run extract" to update catalogs with new messages)
(use "npm run compile" to compile catalogs for production)
```

Also, we need to add the new language in the component LanguageSelector so the user can select it from the app. Go to `src/components/common/LanguageSelector.js` and add the new language in the constant `languageOptions` like this:

```jsx
const languageOptions = [
    {key: 'en', value: 'en', text: 'English'},
    {key: 'es', value: 'es', text: 'Spanish'},
    {key: 'fr', value: 'fr', text: 'French'},
  ];
```

Remember to use in `value` the locale code that you use with the `npm run add-locale` command.

### How to add translations

Wrap the text you want to be translated with Trans macro. Example:

```jsx
import { Trans } from '@lingui/macro'

<p><Trans>Hello {name}</Trans></p>
```

Now, we have to extract all the texts that should be translated into a file where we will add the translations. Use `npm run extract`:

```jsx
$ npm run extract

Catalog statistics:
┌─────────────┬─────────────┬─────────┐
│ Language    │ Total count │ Missing │
├─────────────┼─────────────┼─────────┤
│ en (source) │     10      │    -    │
│ es          │     10      │    0    │
│ fr          │     10      │   10    │
└─────────────┴─────────────┴─────────┘

(use "npm run add-locale <locale>" to add more locales)
(use "npm run extract" to update catalogs with new messages)
(use "npm run compile" to compile catalogs for production)
```

This will create a PO format file with all the texts between Trans macro for each language. Eg: 

```jsx
#: src/components/home/HomePage.js:29
msgid "Choose your participants"
msgstr "Choisissez les participants" // we add the french tranlsation
```

Each time we add a Trans macro we need to run the extract command to update the file.

In order to load the messages into the app we need to compile them. To do that, just run `npm run  compile`

```jsx
$ npm run compile

Compiling message catalogs…
Done!
```

This command will create or update the file `messages.js` for each locale. E.g: `src/locales/fr/messages.js`

Then, we need to import this file into the app in `src/components/App.js` and add it to the `catalogs` constant.

```jsx
import catalogEs from '../locales/es/messages.js';
import catalogFr from '../locales/fr/messages.js';

export default function App() {
	const catalogs = {es: catalogEs, fr: catalogFr};

	return (
	    <I18nProvider language={lang} catalogs={catalogs}>
			[...]
	)
}
```

### Useful links:

- Lingui tutorial - [https://lingui.js.org/tutorials/react.html](https://lingui.js.org/tutorials/react.html)
- Lingui configuration - [https://lingui.js.org/ref/conf.html?highlight=configuration](https://lingui.js.org/ref/conf.html?highlight=configuration)
- Catalogs formats (PO format) - [https://lingui.js.org/ref/catalog-formats.html](https://lingui.js.org/ref/catalog-formats.html)
- Semantic UI React Select - [https://react.semantic-ui.com/addons/select/](https://react.semantic-ui.com/addons/select/)

PatriciaMaPe:
I finished all the changes and added the documentation here. Let me know if its ok :)

ThiefMaster:
I fixed/improved various things I noticed while testing. Since having the language selector without a complete translation isn't particularly useful yet I disabled it by default; append `#i18n` to the URL to enable it.

ThiefMaster (approved):

PR #255, opened 2020-09-22
add 'get_participant_busy_times' API test

tomasr8:


tests/api_test.py
```diff
@@ -179,6 +179,109 @@ def test_get_busy_times(flask_client, dummy_uid, mocker):
...
+        **make_test_auth(dummy_uid),
+        json=json,
+    )
+    assert resp.status_code == 422
+    assert resp.json['messages']['date'][0] == 'Date has no timeslots'
+
+    json = {'date': dummy_newdle.timeslots[0].strftime('%Y-%m-%d'), 'tz': 'US/Pacific'}
+    participant = list(p for p in dummy_newdle.participants if p.auth_uid is not None)[
+        0
+    ]
```
ThiefMaster, line 269: wow i'm amazed what kind of abomination black created when formatting this...

maybe this turns out less ugly:

    participant = next(p for p in ... if ...)

tests/api_test.py
```diff
@@ -179,6 +179,109 @@ def test_get_busy_times(flask_client, dummy_uid, mocker):
...
+        url_for(
+            'api.get_participant_busy_times',
+            code=dummy_newdle.code,
+            participant_code=participant.code,
+        ),
+        **make_test_auth(dummy_uid),
+        json=json,
+    )
+    assert resp.status_code == 422
+    assert resp.json['messages']['date'][0] == 'Date has no timeslots'
```
ThiefMaster, line 264: Another test case with a different timezone might be useful to test that requesting busy dates for a different date (the date before/after) works as long as the newdle has timeslots on that date in the specified timezone.

Example: A newdle has a timeslot on 2020-09-22 from 00:00 to 01:00 in Europe/Zurich. When requesting busy times in UTC, it should fail for 2020-09-22 but work for 2020-09-21 because it's 22:00-23:00 in UTC.

tests/api_test.py
tomasr8: Yeah this one is something else.. I'll try changing it.

tests/api_test.py
tomasr8: Thanks! I added a test for this.

ThiefMaster (approved):

PR #257, opened 2020-09-23
Results overview

### LINKED ISSUE #190: Improve the newdle results overview
When evaluating the pool, having a tabular view, for example, (ideally with the option to rearrange the order in which respondents appear) could become more useful than a static list.

iokasti:
Closes #190

### DIFF
--- a/newdle/client/src/components/ParticipantGrid.js
+++ b/newdle/client/src/components/ParticipantGrid.js
@@ -1,0 +1,108 @@
+import React from 'react';
+import {useSelector} from 'react-redux';
+import _ from 'lodash';
+import moment from 'moment';
+import PropTypes from 'prop-types';
+import {Icon, Table} from 'semantic-ui-react';
+import {
+  getNewdleDuration,
+  getNewdleParticipants,
+  getNewdleTimezone,
+  getNewdleTimeslots,
+} from '../selectors';
+import {serializeDate, toMoment} from '../util/date';
+import UserAvatar from './UserAvatar';
+import styles from './ParticipantGrid.module.scss';
+
+function formatMeetingTime(startTime, duration) {
+  const endTime = moment(startTime)
+    .add(duration, 'm')
+    .format('HH:mm');
+  return `${serializeDate(startTime, 'HH:mm')} - ${endTime}`;
+}
+
+function DateCell({timeslot}) {
+  const newdleTimezone = useSelector(getNewdleTimezone);
+  const duration = useSelector(getNewdleDuration);
+
+  const startTime = toMoment(timeslot, 'YYYY-MM-DDTHH:mm');
+  return (
+    <Table.HeaderCell textAlign="center">
+      <div>
+        <div className={styles['date']}>{startTime.format('D MMM')}</div>
+        <div className={styles['time']}>{formatMeetingTime(startTime, duration)}</div>
+        <div className={styles['timezone']}>{newdleTimezone}</div>
+      </div>
+    </Table.HeaderCell>
+  );
+}
+
+DateCell.propTypes = {
+  timeslot: PropTypes.string,
+};
+
+function ParticipantRow({participant}) {
+  const timeslots = useSelector(getNewdleTimeslots);
+
+  return (
+    <Table.Row textAlign="center">
+      <Table.Cell style={{whiteSpace: 'nowrap'}}>
+        <UserAvatar user={participant} size={30} className={styles['participant-avatar']} />
+        <span className={styles['participant-avatar-label']}>{participant.name}</span>
+      </Table.Cell>
+      {timeslots.map(timeslot => {
+        const status = participant.answers[timeslot];
+        return (
+          <Table.Cell key={timeslot} textAlign="center">
+            {status ? (
+              status === 'available' ? (
+                <Icon color="green" name="checkmark" size="large" />
+              ) : status === 'unavailable' ? (
+                <Icon color="red" name="close" size="large" />
+              ) : (
+                <Icon color="yellow" name="question" size="large" />
+              )
+            ) : null}
+          </Table.Cell>
+        );
+      })}
+    </Table.Row>
+  );
+}
+
+ParticipantRow.propTypes = {
+  participant: PropTypes.shape({
+    answers: PropTypes.objectOf(PropTypes.oneOf(['unavailable', 'available', 'ifneedbe'])),
+    email: PropTypes.string,
+    name: PropTypes.string.isRequired,
+  }).isRequired,
+};
+
+export default function ParticipantGrid() {
+  const timeslots = useSelector(getNewdleTimeslots);
+  const participants = useSelector(getNewdleParticipants);
+
+  if (timeslots.length === 0) {
+    return null;
+  }
+
+  return (
+    <div className={styles['participant-grid']}>
+      <Table textAlign="center">
+        <Table.Header>
+          <Table.Row>
+            <Table.HeaderCell textAlign="center" className={styles['participant-grid-cell']}>
+              Participants
+            </Table.HeaderCell>
+            {timeslots.map(timeslot => (
+              <DateCell key={timeslot} timeslot={timeslot} />
+            ))}
+          </Table.Row>
+          {_.sortBy(participants, ['name']).map(participant => (
+            <ParticipantRow key={participant.id} participant={participant} />
+          ))}
+        </Table.Header>
+      </Table>
+    </div>
+  );
+}
--- a/newdle/client/src/components/ParticipantGrid.js
+++ b/newdle/client/src/components/ParticipantGrid.js
@@ -4,7 +4,7 @@ import {Trans} from '@lingui/macro';
 import _ from 'lodash';
 import moment from 'moment';
 import PropTypes from 'prop-types';
-import {Icon, Table, Radio} from 'semantic-ui-react';
+import {Icon, Table, Radio, Popup} from 'semantic-ui-react';
 import {
   getNewdleDuration,
   getNewdleParticipants,
@@ -13,7 +13,6 @@ import {
 } from '../selectors';
 import {serializeDate, toMoment} from '../util/date';
 import AvailabilityRing from './AvailabilityRing';
-import UserAvatar from './UserAvatar';
 import styles from './ParticipantGrid.module.scss';
 
 function formatMeetingTime(startTime, duration) {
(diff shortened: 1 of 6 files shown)


newdle/client/src/components/ParticipantTablev2.js
```diff
@@ -0,0 +1,137 @@
...
+import {Icon, Table} from 'semantic-ui-react';
+import {serializeDate, toMoment} from '../util/date';
+import {
+  getNewdleDuration,
+  getNewdleParticipants,
+  getNewdleTimezone,
+  getNewdleTimeslots,
+} from '../selectors';
+import UserAvatar from './UserAvatar';
+import styles from './creation/userSearch/UserSearch.module.scss';
```
ThiefMaster, line 13: I'd copy the rule you are using into a stylesheet related to this component. all it contains is `border-radius: 50%;` so i don't think this is worth adding a dependency on the style module of a much more specific component here

newdle/client/src/components/ParticipantTablev2.js
```diff
@@ -0,0 +1,137 @@
...
+        let status = null;
+        if (participant.answers[timeslot] === 'available') {
+          status = 'available';
+        } else if (participant.answers[timeslot] === 'unavailable') {
+          status = 'unavailable';
+        } else if (participant.answers[timeslot] === 'ifneedbe') {
+          status = 'ifneedbe';
+        }
+        return (
+          <Table.Cell key={`${status}-${index}`} textAlign="center">
```
ThiefMaster, line 73: using the index in the key is usually not a good idea, what about `key={timeslot}`?

newdle/client/src/components/ParticipantTablev2.js
```diff
@@ -0,0 +1,137 @@
...
+  if (timeslots.length === 0) {
+    return null;
+  }
+
+  return (
+    <div
+      style={{
+        overflowX: 'auto',
+        borderRight: '1px solid rgba(34, 36, 38, 0.15)',
+        borderLeft: '1px solid rgba(34, 36, 38, 0.15)',
```
ThiefMaster, line 114: these should probably go in a css module

newdle/client/src/components/ParticipantTablev2.js
```diff
@@ -0,0 +1,137 @@
...
+      }}
+    >
+      <Table textAlign="center">
+        <Table.Header>
+          <Table.Row>
+            <Table.HeaderCell
+              textAlign="center"
+              style={{paddingLeft: '40px', paddingRight: '40px'}}
```
ThiefMaster, line 122: same here

newdle/client/src/components/ParticipantTablev2.js
```diff
@@ -0,0 +1,137 @@
...
+            >
+              Participants
+            </Table.HeaderCell>
+            {timeslots.map((timeslot, index) => (
+              <DateCell key={`${timeslot}-${index}`} timeslot={timeslot} />
```
ThiefMaster, line 127: no need to include the index since timeslots should already be unique

newdle/client/src/components/ParticipantTablev2.js
```diff
@@ -0,0 +1,137 @@
...
+            ))}
+          </Table.Row>
+          {participants.map(participant => (
+            <ParticipantRow key={participant.name} participant={participant} />
```
ThiefMaster, line 131: names are not necessarily unique; maybe use their `id` instead

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -1,4 +1,5 @@
 import React, {useState, useCallback} from 'react';
+import Switch from '@material-ui/core/Switch';
```
ThiefMaster, line 2: I'm very surprised this works since we don't use `material-ui` in newdle. Maybe something pulling it in a a dependency (WTF).

Anyway, please use components from `semantic-ui-react` since that's our UI framework in newdle

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -27,6 +29,8 @@ import {usePageTitle} from '../../util/hooks';
 import RecipientList from '../RecipientList';
 
 import styles from './summary.module.scss';
+import FormGroup from '@material-ui/core/FormGroup';
+import FormControlLabel from '@material-ui/core/FormControlLabel';
```
ThiefMaster, line 33: same here (and please keep JS imports above CSS imports)

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -12,6 +13,7 @@ import {
   Label,
 } from 'semantic-ui-react';
 import ParticipantTable from '../ParticipantTable';
+import ParticipantTablev2 from '../ParticipantTablev2';
```
ThiefMaster, line 16: Can we find a better name for this new component? `ParticipantGrid` maybe?

newdle/client/src/components/ParticipantTablev2.js
```diff
@@ -0,0 +1,137 @@
...
+ParticipantRow.propTypes = {
+  participant: PropTypes.arrayOf(
+    PropTypes.shape({
+      answers: PropTypes.objectOf(PropTypes.oneOf(['unavailable', 'available', 'ifneedbe'])),
+      email: PropTypes.string,
+      name: PropTypes.string.isRequired,
+      status: PropTypes.string.isRequired,
+    })
+  ).isRequired,
+};
```
Leats, line 99:
```suggestion
ParticipantRow.propTypes = {
  participant: PropTypes.shape({
    answers: PropTypes.objectOf(PropTypes.oneOf(['unavailable', 'available', 'ifneedbe'])),
    email: PropTypes.string,
    name: PropTypes.string.isRequired,
  }).isRequired,
};
```
(status does not seem to be in this witht he nedwles where I tested this)

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -67,9 +71,20 @@ export default function SummaryPage() {
...
+      <FormGroup row>
+        <FormControlLabel
+          control={
+            <Switch
+              checked={checked}
+              onChange={event => setChecked(event.target.checked)}
+              name="toggle"
+            />
+          }
+          label="Toggle view"
```
Leats, line 85: The grid looks a little weird if nobody voted yet because then it only consists of the headers. Maybe it makes sense to only be able to switch when at least one person has voted already (or has been invited, just so that the table isn't completely empty)?

newdle/client/src/components/ParticipantTablev2.js
Leats, line 131: Participant rows are not sorted and can change when reloading. It would be great to have them somehow sorted (probably best alphabetically)!

ThiefMaster:
I think something went wrong when you tried to update to the latest master. Don't do a merge - instead, do a rebase so your commits are simply applied on top of the latest master.

newdle/client/src/components/ParticipantGrid.js
```diff
@@ -0,0 +1,126 @@
...
+        if (participant.answers[timeslot] === 'available') {
+          status = 'available';
+        } else if (participant.answers[timeslot] === 'unavailable') {
+          status = 'unavailable';
+        } else if (participant.answers[timeslot] === 'ifneedbe') {
+          status = 'ifneedbe';
+        }
+        return (
+          <Table.Cell key={timeslot} textAlign="center">
+            {!!status ? (
```
ThiefMaster, line 75: no need for `!!` here - we only need this if we need to cast to boolean, but since you use it in a ternary expression only truthiness matters.

if the code was `{!!status && (...)}` then the `!!` would help ensure that you don't render e.g. a `0` if status was 0. this isn't actually a problem here, since you have a well-defined list of possible statuses, and the only falsy one is already `null` which doesn't render anything.

newdle/client/src/components/ParticipantGrid.js
```diff
@@ -0,0 +1,126 @@
...
+      </Table.Cell>
+      {timeslots.map(timeslot => {
+        let status = null;
+        if (participant.answers[timeslot] === 'available') {
+          status = 'available';
+        } else if (participant.answers[timeslot] === 'unavailable') {
+          status = 'unavailable';
+        } else if (participant.answers[timeslot] === 'ifneedbe') {
+          status = 'ifneedbe';
+        }
```
ThiefMaster, line 72: Can't this whole block be replaced with `const status = participant.answers[timeslot];`?

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -70,6 +74,14 @@ export default function SummaryPage() {
 
   return (
     <Container text>
+      {hasParticipants > 0 ? (
```
ThiefMaster, line 77: If you do `{hasParticipants > 0 && (...)}` you can avoid the `: null` at the end

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -40,12 +43,13 @@ export default function SummaryPage() {
...
   const isCreator = userInfo !== null && newdle !== null && userInfo.uid === newdle.creator_uid;
   const dispatch = useDispatch();
+  const hasParticipants = useSelector(getNumberOfParticipants);
   const [_sendResultEmails, mailSending, mailError, sendMailResponse] = client.useBackendLazy(
     client.sendResultEmails
   );
   const [_setFinalDate, submitting] = client.useBackendLazy(client.setFinalDate);
   usePageTitle(newdle && `Summary: ${newdle.title}`, true);
-
+  const [checked, setChecked] = useState(false);
```
ThiefMaster, line 52: I don't think this is a good name - if it was in a component which renders a checkbox but nothing else then it'd be fine, but in a large component like this one the name should be more meaningful.

Maybe `gridViewActive` or something like this?

newdle/client/src/components/ParticipantGrid.js
```diff
@@ -0,0 +1,126 @@
...
+
+DateCell.propTypes = {
+  timeslot: PropTypes.string,
+};
+
+function ParticipantRow({participant}) {
+  const timeslots = useSelector(getNewdleTimeslots);
+  let participantName = participant.name;
+  let participantEmail = participant.email;
+  // eslint-disable-next-line react/prop-types
```
ThiefMaster, line 48: is this really needed? i don't see anything in the next line that looks like it'd trigger this rule...

newdle/client/src/components/ParticipantGrid.js
```diff
@@ -0,0 +1,126 @@
...
+}
+
+DateCell.propTypes = {
+  timeslot: PropTypes.string,
+};
+
+function ParticipantRow({participant}) {
+  const timeslots = useSelector(getNewdleTimeslots);
+  let participantName = participant.name;
+  let participantEmail = participant.email;
```
ThiefMaster, line 47: you never reassign so `const` is more appropriate here. but you can simply move this into the `function` line:

    function ParticipantRow({participant: {name, email}}) {

and then use `name` and `email` in here

newdle/client/src/components/ParticipantGrid.js
```diff
@@ -0,0 +1,126 @@
...
+  // eslint-disable-next-line react/prop-types
+  let matches = participantName.match(/\b(\w)/g);
+  let initials = matches.join('');
```
ThiefMaster, line 50: const initials = participantName.match(/\b(\w)/g).join('');

or better:

    const initials = participantName.match(/\b(\w)/g).slice(0, 2).join('');

(otherwise you get more than 2 chars in case the full name has more than two names)

newdle/client/src/components/ParticipantGrid.js
```diff
@@ -0,0 +1,126 @@
...
+
+  return (
+    <Table.Row textAlign="center">
+      <Table.Cell>
+        <div style={{display: 'flex'}}>
+          <UserAvatar
+            user={{email: participantEmail, name: participantName, initials}}
```
ThiefMaster, line 57: with the change suggested above this can be simplified to `user={{email, name, initials}}`

ThiefMaster (changes requested):

tomasr8:
I can't update the original PR so I put it in a comment here:

### Normal view
image

### Grid view
image

https://user-images.githubusercontent.com/8739637/124785868-86aba500-df47-11eb-9905-5ef3494d5762.mp4

tomasr8:
Added interactivity to the grid view:
### Not finalized
notfinalized
### Finalized
finalized

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -193,9 +229,35 @@ export default function SummaryPage() {
...
+                  {config.has_event_creation && (
+                    <Button
+                      className={styles['create-event-button']}
+                      onClick={createEvent}
+                      loading={loading}
+                    >
+                      <Trans>Create event</Trans>
+                    </Button>
+                  )}
+                </>
```
ThiefMaster, line 259: We have this exact same code twice in that file now. Can we avoid this duplication?

ThiefMaster: Moving the duplicate above the `return` block should be enough:

```jsx
  const actions = (
    <>
      {hasParticipantsWithEmail && (
        <Button
          icon
          color="blue"
          labelPosition="left"
          loading={mailSending}
          disabled={mailSending || mailSent}
          onClick={() => setMailModalOpen(true)}
        >
          <Icon name="mail" />
          <Trans>E-mail participants</Trans>
        </Button>
      )}
      {config.has_event_creation && (
        <Button className={styles['create-event-button']} onClick={createEvent} loading={loading}>
          <Trans>Create event</Trans>
        </Button>
      )}
    </>
  );
```

tomasr8: Thanks, I'll update it :+1:

ThiefMaster (approved):

PR #261, opened 2020-09-23
Add support for editing newdles

### LINKED ISSUE #185: Edit newdle / add new slots after creation
When a Newdle is created, the host sets a set of slots which can be selected by the potential participants.

It would be useful to also be able to add further slots after the Newdle has been created. Typical use cases is where many participants would not be able to make the initial proposals and the meeting organiser would like to add further options but still retain the other options in a single poll if not all the potential participants have participated.

If it would also be possible to send an update to those who had responded to the Newdle that new slots have been made available, this would be very helpful.

plourenco:
Closes #185 

Creates a newdle editing functionality re-using most of the creation components.

- [x] Any action that fails to notify through email, still succeeds the operation
- [x] ~Purge participant answers in timeslots that have been removed~ (no longer considered)
> In case of timeslot removal we have to check if the presence of answers for deleted timeslots is a problem; in that case we could purge them from all participants.


newdle/client/src/reducers/creation.js
```diff
@@ -66,6 +75,12 @@ export default combineReducers({
   },
   participants: (state = [], action) => {
     switch (action.type) {
+      case NEWDLE_RECEIVED:
+        // TODO: remove condition when initials get settled
+        // This condition exists because a no-permission returns null
+        return action.newdle.participants
+          ? action.newdle.participants.map(p => ({...p, initials: 'AB'}))
+          : [];
```
plourenco, line 83: RFC: Can we get rid of this?
User search results (when creating a newdle) return a user object that contains an `initials` attribute derived from the first and last name. Participants do not. What should we do here? Add a dynamic field in the schema that does it for us? Or, even better, generate one in the client?

Eventually this would become just: `return action.newdle.participants`.

ThiefMaster: Should we just ditch the two-char initials in favor of a simple `name[0]`? It's impossible to generate initials from a single name field since we don't know what's the first/last name.

ThiefMaster:
> Purge participant answers in timeslots that have been removed

I don't thinks this is needed unless there's some code which doesn't behave correctly if there are answers for non-existent timeslots in the DB.

BTW, I think we could actually make use of the fact that any time the keys in the user's answers are not exactly the available timeslots of the newdle it means a timeslot has been added/removed, so we could show a message telling the user that the timeslots have been changed and that they may want to  update their answers if needed

newdle/api.py
```diff
@@ -278,6 +279,51 @@ def create_newdle(title, duration, timezone, timeslots, participants, private, n
     return NewdleSchema().jsonify(newdle)
 
 
+@api.route('/newdle/<code>', methods=('PATCH',))
+@use_args(EditNewdleSchema(partial=True), locations=('json',))
+def edit_newdle(args, code):
```
ThiefMaster, line 284: We already have `update_newdle` for PATCH; let's reuse this and just add the missing code.

newdle/api.py
```diff
@@ -319,13 +365,20 @@ def delete_newdle(code):
 
 @api.route('/newdle/<code>/participants/')
 @allow_anonymous
-def get_participants(code):
+def get_participants(code):  # TODO: Do we need a flag to opt signatures out?
```
ThiefMaster, line 368: probably not

newdle/api.py
```diff
@@ -319,13 +365,20 @@ def delete_newdle(code):
...
         'Specified newdle does not exist'
     )
     if newdle.private and (g.user is None or newdle.creator_uid != g.user['uid']):
         raise Forbidden('You cannot view the participants of this newdle')
-    return RestrictedParticipantSchema(many=True).jsonify(newdle.participants)
+    participants = RestrictedParticipantSchema(many=True).dump(newdle.participants)
+    return jsonify(
+        [
+            sign_user({**u, 'uid': u['auth_uid']}, fields={'email', 'name', 'uid'})
+            for u in participants
```
ThiefMaster, line 378: `p` or `part`?

newdle/client/src/client.js
```diff
@@ -141,6 +141,14 @@ class Client {
     return this._request(flask`api.create_newdle`(), params);
   }
 
+  editNewdle(code, attrs) {
```
ThiefMaster, line 144: Let's rename + make `setFinalDate` generic and turn `setFinalDate` into a shortcut calling the generic one with just the final_dt update

newdle/client/src/components/NewdleTitle.js
```diff
@@ -41,12 +38,7 @@ export default function NewdleTitle({
                 content={!finished ? 'Answer newdle' : 'This newdle has already finished'}
                 position="bottom center"
                 trigger={
-                  <Button
-                    icon
-                    active={!isSummary}
-                    onClick={() => (isSummary ? history.push(answerURL) : null)}
-                    disabled={finished}
-                  >
+                  <Button icon exact disabled={finished} as={NavLink} to={answerURL}>
```
ThiefMaster, line 41: Without the custom onClick logic, does this spam entries in the history whenever you click? (we do not want to add a new history entry when you are already on the page)

newdle/client/src/components/creation/CreationHeader.js
```diff
@@ -6,6 +6,7 @@ import {useDispatch, useSelector} from 'react-redux';
 import {getStep, getFullTimeslots} from '../../selectors';
 import {setStep} from '../../actions';
 import styles from './CreationHeader.module.scss';
+import {STEPS} from './steps';
 
 export default function CreationHeader() {
   const steps = [
```
ThiefMaster, line 12: maybe this should go into `steps.js` (and be turned into a dict using the steps you defined there as an index)?

newdle/client/src/components/creation/EditPage.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+  if (!newdle) {
+    return <Loader active />;
+  }
+
+  return (
+    <Switch>
+      <Route exact path="/newdle/:code/edit" render={() => <TimeslotsStep isEditing />} />
+      <Route path="/newdle/:code/edit/participants" render={() => <ParticipantsStep isEditing />} />
+      <Route path="/newdle/:code/edit/options" render={() => <FinalStep isEditing />} />
```
ThiefMaster, line 43: can't these be exact as well?

newdle/client/src/components/creation/EditPage.js
```diff
@@ -0,0 +1,46 @@
...
+  if (!isUserLoggedIn) {
+    return <Redirect to="/" />;
+  }
+  if (!newdle) {
+    return <Loader active />;
+  }
+
+  return (
+    <Switch>
+      <Route exact path="/newdle/:code/edit" render={() => <TimeslotsStep isEditing />} />
```
ThiefMaster, line 41: maybe it'd be cleaner to have `.../timeslots` there?

newdle/client/src/components/TopHeader.js
```diff
@@ -32,6 +32,7 @@ export default function TopHeader() {
             <Switch>
               <Route exact path="/" component={HomeHeader} />
               <Route exact path="/new" component={CreationHeader} />
+              <Route path="/newdle/:code/edit" component={SummaryHeader} />
```
ThiefMaster, line 35:
```suggestion
              <Route path="/newdle/:code/edit/" component={SummaryHeader} />
```

newdle/client/src/components/creation/FinalStep.js
```diff
@@ -45,6 +50,15 @@ export default function FinalStep() {
     }
   }
 
+  async function editNewdle() {
+    const newdle = await _editNewdle(activeNewdle.code, {title, private: isPrivate, notify});
+
+    if (newdle) {
+      history.push('/new/success');
```
ThiefMaster, line 57: shouldn't we go back to the summary page after editing?

newdle/client/src/components/creation/FinalStep.js
```diff
@@ -123,35 +139,51 @@ export default function FinalStep() {
...
     </Container>
   );
 }
+
+FinalStep.propTypes = {
+  isEditing: PropTypes.bool,
+};
+
+FinalStep.defaultProps = {
+  isEditing: true,
```
ThiefMaster, line 188: that seems like a rather strange default tbh. I'd default to false or make it required

newdle/client/src/components/creation/timeslots/CandidateSlot.js
```diff
@@ -64,13 +64,19 @@ export default function CandidateSlot({
...
   const slot = (
     <Slot
       width={width}
       pos={pos}
       moreStyles={styles['candidate']}
-      tooltip={`${startTime} - ${endTime}`}
+      tooltip={
+        <div style={{textAlign: 'center'}}>
+          <span>{`${startTime} - ${endTime}`}</span>
+          <p>{text}</p>
```
ThiefMaster, line 77: do we really want to render  an empty `<p>` here in case there is no text?

newdle/client/src/components/creation/timeslots/Timeline.js
```diff
@@ -211,13 +213,15 @@ function TimelineInput({minHour, maxHour}) {
...
               const slotProps = getCandidateSlotProps(time, duration, minHour, maxHour);
+              const participants = availability?.find(a => a.startDt === `${date}T${time}`);
               return (
                 <CandidateSlot
                   {...slotProps}
                   key={time}
                   isValidTime={time => !candidates.includes(time)}
                   onDelete={() => handleRemoveSlot(time)}
                   onChangeSlotTime={newStartTime => handleUpdateSlot(time, newStartTime)}
+                  text={(participants?.availableCount || 'No') + ' participants registered'}
```
ThiefMaster, line 224: let's use proper i18n and pluralization here since it's available now :)

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -230,19 +233,30 @@ export default function SummaryPage() {
...
+                text="Edit"
+                floating
+                button
+                direction="right"
+                className={styles['edit-button']}
+              >
+                <Dropdown.Menu>
+                  <Dropdown.Item as={Link} to={`${editUrl}/participants`} text="Participants" />
+                  <Dropdown.Item as={Link} to={editUrl} text="Timeslots" />
+                  <Dropdown.Item as={Link} to={`${editUrl}/options`} text="Options" />
```
ThiefMaster, line 256: i18n

newdle/notifications.py
```diff
@@ -5,9 +5,15 @@
...
+    newdle,
+    subject,
+    text_template,
+    html_template,
+    get_context,
+    attachments=None,
+    participants=None,
 ):
-    participants = [p for p in newdle.participants if p.email]
+    participants = participants or [p for p in newdle.participants if p.email]
```
ThiefMaster, line 16:
```python
if participants is None:
    participants = ...
```

(right now calling the function with an empty list would notify everyone)

newdle/notifications.py
```diff
@@ -48,8 +54,11 @@ def notify_newdle_creator(
 
 
 def send_emails(emails):
-    with get_connection() as conn:
-        return conn.send_messages(emails)
+    try:
+        with get_connection() as conn:
+            return conn.send_messages(emails)
+    except ConnectionRefusedError:
```
ThiefMaster, line 60: I don't really like the idea of silently failing when sending emails fails...

plourenco: What would the alternative be? The reason I decided to silence emails is that answering a newdle results in a failure if, for example, SMTP fails. Which is a misconception since the answering actually succeeds.

ThiefMaster: since we don't have any background task for sending emails and I don't want to add this complexity just for that: at least when creating a newdle or notifying people about the final date I'd rather fail and have the user try submitting again (which may work in case of a temporary error). I guess for answering you could argue that losing a notification isn't a big deal.

newdle/client/src/components/creation/timeslots/Timeline.js
plourenco: Ah yes, we do have them now!

newdle/client/src/components/creation/FinalStep.js
plourenco: Agree, I expect false too (as I've made the others), this was a mistake.

newdle/client/src/components/creation/FinalStep.js
plourenco: In principle, we would and it worked but there is no feedback for the user. The options were:
- Create new bottom right dialogs for success like we do for errors
- Re-use the creation page.

ThiefMaster: Showing a message on the page where you saved the edit and only redirecting when closing it would be another option

newdle/client/src/components/creation/EditPage.js
plourenco: Yes, but seems to weird to have `edit` pointing nowhere. Perhaps defaulting to `timeslots`?

newdle/client/src/components/creation/EditPage.js
plourenco: Yes, they can. Didn't add since I expected it to be harmless.

newdle/client/src/components/creation/EditPage.js
ThiefMaster: yeah, you could redirect `.../edit` to `/edit/timeslots` (i guess that's what you meant with defaulting to timeslots?)

newdle/client/src/components/creation/CreationHeader.js
plourenco: Had the same idea before but then realized only `CreationHeader` needs this information.
Nonetheless, I was still considering to move it until I noticed that making the dict keys being step strings such as `STEPS.PARTICIPANTS` but with the values being an object with `title` and `description` causes us to not being able to use them like numbers (like it happens in `CreationHeader` just below):
```javascript
setStep(STEPS.PARTICIPANTS + 1)
```
If this still seems useful, I can change it.

newdle/client/src/components/NewdleTitle.js
plourenco: Yes, it does. Which means that when going back, it will return to the previous button link. I can change it to do the opposite.

newdle/client/src/reducers/creation.js
plourenco: We could divide the space and try to figure out the initials, or fallback to the first two chars. How does this sound?

newdle/client/src/components/creation/EditPage.js
plourenco: Exactly.

newdle/client/src/reducers/creation.js
ThiefMaster: single char is more reliable IMO - especially at CERN you have so many people with multiple first and last names...

newdle/client/src/client.js
plourenco: Alright. With renaming, you meant the same thing or also changing `editNewdle` to something as `updateNewdle`?

ThiefMaster: basically having one `updateNewdle` function where you pass the payload for the PATCH, and then call that one from `setFinalDate`

newdle/client/src/components/creation/timeslots/CandidateSlot.js
plourenco: I can ditch it if undefined.

newdle/client/src/reducers/creation.js
plourenco: Only realized we were talking about a single char, after replacing it to be the first two chars of the name. It seems good, nonetheless.

newdle/api.py
```diff
@@ -428,35 +469,38 @@ def update_participant(args, code, participant_code):
...
+                        for timeslot, answer in participant.answers.items()
+                        if answer != Availability.unavailable
+                    ],
+                    'summary_link': url_for(
+                        'newdle_summary', code=participant.newdle.code, _external=True
+                    ),
+                },
+            )
+        except ConnectionRefusedError:
+            logging.error('Failed notifying the newdle creator', exc_info=True)
```
ThiefMaster, line 502: `app.logger.exception(...)`

- we use the flask logger in other places as well
- `.exception()` already includes the exc_info by default

newdle/api.py
```diff
@@ -323,36 +324,68 @@ def create_newdle(title, duration, timezone, timeslots, participants, private, n
...
-
 @api.route('/newdle/<code>', methods=('PATCH',))
-@use_args(UpdateNewdleSchema(), locations=('json',))
+@use_args(NewNewdleSchema(partial=True), locations=('json',))
 def update_newdle(args, code):
     newdle = Newdle.query.filter_by(code=code).first_or_404(
         'Specified newdle does not exist'
     )
-    if newdle.creator_uid != g.user['uid']:
+    if g.user is None or newdle.creator_uid != g.user['uid']:
```
ThiefMaster, line 337: `g.user is None` should never happen here since the code in `before_request` aborts early for endpoints without `allow_anonymous`

newdle/api.py
```diff
@@ -323,36 +324,68 @@ def create_newdle(title, duration, timezone, timeslots, participants, private, n
...
         raise Forbidden
+    new_participants = []
+    if 'participants' in args:
+        participants = args.pop('participants')
+        # Filter the new participants to be created (left diff)
+        new_participants = [
+            Participant(**_p)
+            for _p in participants
+            if not any(_p['auth_uid'] == p.auth_uid for p in newdle.participants)
```
ThiefMaster, line 346: that nested loop with `p` and `_p` is a bit nasty ;) why not create a set with the existing uids and then check `_p['auth_uid'] in existing_uids`? (also rename `_p` to `p` in that case)

newdle/client/src/reducers/creation.js
```diff
@@ -139,6 +157,8 @@ export default combineReducers({
   },
   private: (state = false, action) => {
     switch (action.type) {
+      case NEWDLE_RECEIVED:
+        return action.newdle.private;
```
ThiefMaster, line 161: Don't you need something like this below for `notify` as well?

newdle/client/src/locales/es/messages.js
```diff
@@ -0,0 +1 @@
...:[["0","plural",{zero:"No participants registered",one:["#"," participant registered"],other:["#"," participants registered"]}]],"{0} will take place on:":[["0"]," will take place on:"],"{availableCount, plural, one {# available participant} other {# available participants}}":[["availableCount","plural",{one:["#"," available participant"],other:["#"," available participants"]}]],"{hours, plural, one {{hours} hour} other {{hours} hours}}":[["hours","plural",{one:[["hours"]," hour"],other:[["hours"]," hours"]}]],"{hours}h {minutes}min":[["hours"],"h ",["minutes"],"min"],"{numSlots, plural, one {<0>#</0> slot selected } other {<1>#</1> slots selected } =0 {You haven't selected any slots yet}}":[["numSlots","plural",{one:["<0>","#","</0> slot selected "],other:["<1>","#","</1> slots selected "],0:"You haven't selected any slots yet"}]],"{numberOfTimeslots, plural, one {{numberOfAvailable} out of # option chosen} other {{numberOfAvailable} out of # options chosen}}":[["numberOfTimeslots","plural",{one:[["numberOfAvailable"]," out of ","#"," option chosen"],other:[["numberOfAvailable"]," out of ","#"," options chosen"]}]],"{title} will take place on:":[["title"]," will take place on:"]};
```
ThiefMaster, line 1: Delete this file and run `make` to get the latest js dependencies - the latest linguijs update switched from .js to .jsm so I updated the gitignore...

newdle/api.py
```diff
@@ -323,36 +324,68 @@ def create_newdle(title, duration, timezone, timeslots, participants, private, n
...
+        ]
+        # Filter the existing participants so we don't reset them (intersection)
+        newdle.participants = {
+            p
+            for p in newdle.participants
+            if any(p.auth_uid == _p['auth_uid'] for _p in participants)
```
ThiefMaster, line 352: same here

newdle/api.py
```diff
@@ -323,36 +324,68 @@ def create_newdle(title, duration, timezone, timeslots, participants, private, n
...
+        'invitation_email.html',
+        lambda p: {
+            'creator': newdle.creator_name,
+            'title': newdle.title,
+            'answer_link': url_for(
+                'newdle', code=newdle.code, participant_code=p.code, _external=True
+            ),
+        },
+        participants=new_participants,
+    )
```
ThiefMaster, line 373: I guess this is very similar to what we have when creating a newdle? If yes, maybe we could add a `send_invitation_emails(newdle, participants=None)` util and call this in both places?

newdle/api.py
```diff
@@ -373,7 +406,15 @@ def get_participants(code):
...
         raise Forbidden('You cannot view the participants of this newdle')
-    return RestrictedParticipantSchema(many=True).jsonify(newdle.participants)
+    participants = RestrictedParticipantSchema(many=True).dump(newdle.participants)
+    return jsonify(
+        [
+            sign_user(p, fields={'email', 'name', 'auth_uid'})
+            if p['auth_uid'] is not None
+            else p
+            for p in participants
+        ]
```
ThiefMaster, line 416: This might better fit in a `post_dump` hook on the schema itself (or a subclass of it it we use it in other places as well)?

newdle/client/src/components/creation/timeslots/CandidateSlot.js
```diff
@@ -97,4 +103,5 @@ CandidateSlot.propTypes = {
   onDelete: PropTypes.func.isRequired,
   onChangeSlotTime: PropTypes.func.isRequired,
   isValidTime: PropTypes.func.isRequired,
+  text: PropTypes.string,
```
ThiefMaster, line 106:
```js
CandidateSlot.defaultProps = {
  text: '',
};
```

newdle/client/src/components/creation/timeslots/Timeline.js
```diff
@@ -211,13 +213,22 @@ function TimelineInput({minHour, maxHour}) {
...
                 <CandidateSlot
                   {...slotProps}
                   key={time}
                   isValidTime={time => !candidates.includes(time)}
                   onDelete={() => handleRemoveSlot(time)}
                   onChangeSlotTime={newStartTime => handleUpdateSlot(time, newStartTime)}
+                  text={
+                    participants &&
+                    plural(participants.availableCount, {
+                      zero: 'No participants registered',
```
ThiefMaster, line 227: English doesn't have a special plural form for `zero`, you need `_0` here.

https://js-lingui-git-next.lingui-js.now.sh/tutorials/react.html#beware-of-zeroes

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -229,88 +212,130 @@ export default function SummaryPage() {
...
+            </div>
+          )}
+          <DeleteModal open={deletionModalOpen} setOpen={setDeletionModalOpen} />
         </>
       )}
     </Container>
   );
 }
+
+export function DeleteModal({open, setOpen}) {
```
ThiefMaster, line 253: I think this component is big enough to go in its own .js file

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -229,88 +212,130 @@ export default function SummaryPage() {
...
+            {sendDeletionMail && (
+              <Input
+                type="text"
+                placeholder={t`Leave a comment (optional)`}
+                className={styles['deletion-comment']}
+                value={comment}
+                onChange={(__, {value}) => setComment(value)}
+              />
+            )}
+            {hasParticipantsWithoutEmail && (
```
ThiefMaster, line 315: Unrelated, but shouldn't we show this only if  `hasParticipantsWithoutEmail && sendDeletionMail`? Showing this warning when you didn't select to notify people is a bit pointless..

newdle/client/src/components/creation/FinalStep.js
```diff
@@ -132,35 +152,51 @@ export default function FinalStep() {
...
-          <Trans>Create your newdle!</Trans>{' '}
-          <span role="img" aria-label="Newdle">
-            🍜
-          </span>
+          {!isEditing ? (
+            <>
+              <Trans>Create your newdle!</Trans>{' '}
+              <span role="img" aria-label="Newdle">
+                🍜
+              </span>
```
ThiefMaster, line 163: Unrelated to your changes, but I think this is one of the cases where disabling the a11y warning instead of wrapping the emoji in a span with a label may be more appropriate; It's just decoration, a screen reader user would not have any benefit from it telling them something like "Create your newdle! newdle"...

newdle/client/src/components/App.js
```diff
@@ -37,6 +37,7 @@ export default function App() {
             <Route exact path="/mine" component={MyNewdles} />
             <Route exact path="/participating" component={NewdlesParticipating} />
             <Route path="/newdle/:code/summary" component={SummaryPage} />
```
ThiefMaster, line 39: unrelated to your changes, but this one should probably be `exact` since the summary doesn't have any sub-urls, does it?

newdle/api.py
plourenco: Makes sense, had the impression we had it before.

newdle/client/src/components/summary/SummaryPage.js
plourenco: And it is worth having it separated when it's only used here?

newdle/client/src/reducers/creation.js
plourenco: Good catch, this came from the latest rebase.

newdle/client/src/components/summary/SummaryPage.js
ThiefMaster: Avoiding huge files with multiple large components...

I mean I like putting small 'local' components in the same file because it's more convenient but when things get large I like e.g. the ability to open component A on the left side of my editor and component B on the right side (and opening the same file twice is awkward ;))

newdle/api.py
plourenco: Sounds like a good idea. I guess initially I just tried to avoid having two sets (the block below too).

newdle/api.py
plourenco: Should we include this together in the `notifications.py` file or create new util and move some of the email method-objects there?

ThiefMaster: `notifications.py` seems like a good place

newdle/api.py
plourenco: Sure, we're using it everywhere so, sounds good.

plourenco: In the same way, we can also add another `post_dump` to `UserSearchResultSchema`.

newdle/client/src/components/summary/SummaryPage.js
plourenco: I agree

newdle/client/src/components/App.js
plourenco: It doesn't, I usually omit it if there are no conflicts. But, we should remain consistent.

newdle/client/src/components/creation/timeslots/Timeline.js
ThiefMaster: Too bad - probably needs to be separated with a ternary operator then because `zero` probably won't work...  :/

I'd also open an issue in the lingui repo about this.

plourenco: lingui/js-lingui:770

plourenco: lingui/js-lingui#770

newdle/client/src/components/creation/FinalStep.js
```diff
@@ -65,31 +87,29 @@ export default function FinalStep() {
...
-        )}
-      </div>
+          )}
+        </div>
+      )}
       <div className={styles['advanced-options']}>
         <div
           className={styles['headerbar']}
-          onClick={() => setShowAdvancedOptions(!showAdvancedOptions)}
+          onClick={() => setShowAdvancedOptions(isEditing || !showAdvancedOptions)}
```
ThiefMaster, line 112: This prevents you from closing the options. I think the initial value for the `showAdvancedOptions` state is enough and we can simply revert the change in this line.

ThiefMaster: Alternatively, hide the button and pointer cursor and don't make it collapsible in edit mode

newdle/client/src/components/summary/DeleteModal.js
```diff
@@ -0,0 +1,105 @@
...
+
+  return (
+    <Modal onClose={handleClose} size="small" closeIcon open={open}>
+      <Modal.Header>
+        <Trans>Delete this newdle</Trans>
+      </Modal.Header>
+      <Modal.Content>
+        <Trans>
+          Do you really wish to delete this newdle? Please be aware that{' '}
+          <strong>this action cannot be undone.</strong>
```
ThiefMaster, line 61: should the `.` be inside or outside the strong?

newdle/client/src/components/summary/DeleteModal.js
```diff
@@ -0,0 +1,105 @@
...
+                  </Trans>
+                </strong>
+                <RecipientList recipients={participantsWithoutEmail} color="red" icon="close" />
+              </Message>
+            )}
+          </div>
+        )}
+      </Modal.Content>
+      <Modal.Actions>
+        <Button negative onClick={handleConfirm} loading={deletionMailSending || deleting}>
```
ThiefMaster, line 94: Probably the bug was there before, but this needs to set `disabled` using the same condition as we have for `loading`

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -229,86 +210,38 @@ export default function SummaryPage() {
...
-            </Modal.Content>
-            <Modal.Actions>
+            <div className={styles['button-row']}>
               <Button
-                negative
-                onClick={() => handleDeletionModalConfirm(true)}
-                loading={deletionMailSending || deleting}
+                type="submit"
+                loading={submitting}
+                disabled={!finalDate}
```
ThiefMaster, line 217: `!finalDate || submitting`

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -229,86 +210,38 @@ export default function SummaryPage() {
...
+                direction="right"
+                color="violet"
+                as={Button}
+                className={styles['edit-button']}
+              >
+                <Dropdown.Menu>
+                  <Dropdown.Item as={Link} text={t`Participants`} to={`${editUrl}/participants`} />
+                  <Dropdown.Item as={Link} text={t`Timeslots`} to={editUrl} />
+                  <Dropdown.Item as={Link} text={t`Options`} to={`${editUrl}/options`} />
+                  <Dropdown.Item
```
ThiefMaster, line 236: The dropdown is not visible for finalized newdles (since editing those makes no sense), so we need to show a standalone delete button in that case.

newdle/client/src/components/creation/timeslots/Timeline.js
```diff
@@ -164,6 +165,7 @@ function TimelineInput({minHour, maxHour}) {
   const date = useSelector(getCreationCalendarActiveDate);
   const candidates = useSelector(getTimeslotsForActiveDate);
   const pastCandidates = useSelector(getPreviousDayTimeslots);
+  const availability = useSelector(getParticipantAvailability);
   const [editing, setEditing] = useState(!!candidates.length);
```
ThiefMaster, line 169: Sometimes this component renders before the data is fully loaded, in which case editing may default to false.

This change, while not super pretty, avoids it:

```suggestion
  const [_editing, setEditing] = useState(false);
  const editing = _editing || !!candidates.length;
```

newdle/api.py
```diff
@@ -307,52 +308,53 @@ def create_newdle(title, duration, timezone, timeslots, participants, private, n
...
-    if newdle.creator_uid != g.user['uid']:
+    if g.user is None or newdle.creator_uid != g.user['uid']:
         raise Forbidden
+    new_participants = []
+    if 'participants' in args:
+        participants = args.pop('participants')
+        # Filter the new participants to be created (left diff)
+        uids = set(p.auth_uid for p in newdle.participants)
+        new_participants = [
+            Participant(**p) for p in participants if not p['auth_uid'] in uids
```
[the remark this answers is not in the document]
plourenco: Can't believe I didn't even notice this

newdle/client/src/components/creation/EditPage.js
```diff
@@ -0,0 +1,51 @@
...
+  const dispatch = useDispatch();
+  usePageTitle('Editing newdle');
+
+  useEffect(() => {
+    if (newdleCode) {
+      // TODO: We should re-use the existing newdle data instead. However, abortCreation will clear the state,
+      //  causing an edit -> summary -> edit to become empty if we remove the line below
+      //  Ideally, we should move abortCreation() - state flushing - to happen on mount instead of unmount.
+      dispatch(fetchNewdle(newdleCode, true));
+    }
```
ThiefMaster, line 25: When reloading the page on an edit URL this loads the newdle twice since the SummaryHeader also triggers a fetch. I see two options here to avoid this:

- Do not use `SummaryHeader` for editing but a new `EditHeader`. This component could simply render the `SummaryHeader`; since it's different nesting I'd expect it to be a fresh render (ie the `SummaryHeader` will fetch the newdle again
- While it was me originally suggesting to always load fresh data, an alternative would be to just remove it here and assume people do not keep the summary page open for an extended time *and* make edits to their newdle from another tab/browser/device in the meantime.

newdle/api.py
```diff
@@ -307,52 +308,53 @@ def create_newdle(title, duration, timezone, timeslots, participants, private, n
...
     )
-    if newdle.creator_uid != g.user['uid']:
+    if g.user is None or newdle.creator_uid != g.user['uid']:
         raise Forbidden
+    new_participants = []
+    if 'participants' in args:
+        participants = args.pop('participants')
+        # Filter the new participants to be created (left diff)
+        uids = set(p.auth_uid for p in newdle.participants)
+        new_participants = [
```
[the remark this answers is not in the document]
plourenco: This is the same as the first revision, right?

newdle/client/src/components/NewdleTitle.js
```diff
@@ -46,8 +45,9 @@ export default function NewdleTitle({
                 trigger={
                   <Button
                     icon
-                    active={!isSummary}
-                    onClick={() => (isSummary ? history.push(answerURL) : null)}
+                    active={!isSummaryRoute}
```
ThiefMaster, line 48: We need to add another `!isEditRoute` check here, since we need the following logic:

- on answer route (hard to match here): answer active, summary clickable
- on summary route: answer clickable, summary active
- on edit route: both clickable, none active

newdle/api.py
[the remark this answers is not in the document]
ThiefMaster: maybe flake8 would have noticed, but we currently don't run linters in case the tests fail

newdle/api.py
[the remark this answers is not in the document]
ThiefMaster: not sure what you mean - my suggestion is using a set instead of a list, so you can use `|=` below

newdle/client/src/components/creation/FinalStep.js
plourenco: I thought it wasn't supposed to be collapsible, but the caret is misleading. So perhaps an initial value is the simplest.

newdle/client/src/reducers/creation.js
```diff
@@ -35,6 +36,14 @@ export default combineReducers({
   },
   timeslots: (state = {}, action) => {
     switch (action.type) {
+      case NEWDLE_RECEIVED: {
```
Leats, line 39: Timezone and duration should be handled as well so it doesn't default to the standard values when editing a newdle

newdle/client/src/components/creation/EditPage.js
plourenco: Exactly, this is something I would like to improve as the notes above. I think the problem is that we're consistently refreshing the global state, while the purpose of redux here should be the exact opposite.

IMO, we should refactor the way we handle state just like you suggested in the second point and ideally find a better place to load it rather than the header (because we have less control there). However, since I'm not sure about the impact of this change, I would suggest having it in a separate PR - and keeping the loads in this one.

newdle/client/src/reducers/creation.js
plourenco: Thanks! These new fields eventually sneaked in after rebasing.

newdle/api.py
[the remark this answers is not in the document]
plourenco: Ah, I confused the lines. Sounds good.
`notify_newdle_participants` in the lines below expects `participants` to be a list, but since they are handled in the same way (at least for now), I guess this is fine.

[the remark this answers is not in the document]
ThiefMaster: I think it works fine with a set... at least I don't see any operation in `notify_newdle_participants` that wouldn't work with a set.

newdle/client/src/components/summary/DeleteModal.js
plourenco: I think we should have it outside.

newdle/client/src/components/summary/DeleteModal.js
plourenco: Good point. I actually expected this Button to prevent submitting while loading (disabled). It would make more sense.

newdle/client/src/components/summary/SummaryPage.js
plourenco: Now, that I just noticed this happens in at least more than one case. What if we created a wrapper that disables the button while loading? This seems like a proper use-case, right?

newdle/client/src/components/creation/timeslots/Timeline.js
plourenco: How did you manage to reproduce it?

plourenco: What do you think about an early `return <Loader active />`?

ThiefMaster: I think it makes sense, but doesn't fix the problem: you cannot return before calling all hooks, and once the state got the default value, it's gonna stay that way...

plourenco: Right, I also just realised I have to means of checking the intermediate loading state that works for creation and editing.
I'm going with this, way better than adding a new side-effect (useEffect) just for this.

newdle/client/src/components/NewdleTitle.js
plourenco: Good point, we also desperately need an alternative routing solution, I'm adding a TODO on that to prevent adding more changes here.

newdle/client/src/reducers/creation.js
```diff
@@ -99,6 +111,9 @@ export default combineReducers({
   },
   duration: (state = DEFAULT_DURATION, action) => {
     switch (action.type) {
+      case NEWDLE_RECEIVED:
+        // Newdle might be deleted and have no duration (null)
+        return action.newdle.duration || null;
```
plourenco, line 116: RFC: As an alternative, we could also modify `fetchNewdle` and return a newdle with:
```javascript
return {duration: null, ...newdle}
```
Not sure what is easier to maintain.

ThiefMaster: i'd keep it in the reducers

newdle/client/src/components/creation/timeslots/Timeline.js
ThiefMaster: Since it was just a bug in the typescript definitions in the end, let's use `0: ...` here

newdle/client/src/components/summary/SummaryPage.js
ThiefMaster: You added the new `Button` wrapper but don't actually use it in here...

newdle/api.py
```diff
@@ -307,52 +306,53 @@ def create_newdle(title, duration, timezone, timeslots, participants, private, n
...
-        'Specified newdle does not exist'
-    )
-    if newdle.deleted:
-        return DeletedNewdleSchema().jsonify(newdle)
-    return RestrictedNewdleSchema().jsonify(newdle)
-
-
 @api.route('/newdle/<code>', methods=('PATCH',))
-@use_args(UpdateNewdleSchema(), locations=('json',))
+@use_args(NewNewdleSchema(partial=True), locations=('json',))
```
ThiefMaster, line 316: When creating a newdle we currently do not allow anonymous participants (no email/auth_uid). But when editing, we do need to allow such participants since they may already be in the newdle - and right now saving the updated participants fails in that case:

image

ThiefMaster:
I pushed a few small fixes, so please fetch those before making more changes next week.

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -194,6 +175,13 @@ export default function SummaryPage() {
                   <Button className={styles['create-event-button']}>
                     <Trans>Create event</Trans>
                   </Button>
+                  <Button
+                    color="red"
+                    className={styles['delete-button']}
+                    onClick={() => setDeletionModalOpen(true)}
+                  >
+                    <Trans>Delete newdle</Trans>
+                  </Button>
```
ThiefMaster, line 184: I don't really like the placement of the delete button in case of a finalized newdle:

image

- It now requires wrapping the text in the buttons, which doesn't look great
- It looks like an action that's related to the slot, not the whole newdle

I would move it back to its original place below the list of slots (ie where it was before this PR).

newdle/client/src/components/creation/ParticipantsStep.js
```diff
@@ -1,26 +1,63 @@
...
+  const participants = useSelector(getParticipants);
+  const activeNewdle = useSelector(getCreatedNewdle);
+  const history = useHistory();
+  const [_editNewdle, submitting] = client.useBackendLazy(client.updateNewdle);
+
+  async function editNewdle() {
+    const newdle = await _editNewdle(activeNewdle.code, {participants});
+
+    if (newdle) {
+      dispatch(updateNewdle(newdle));
```
ThiefMaster, line 26: We also need to `dispatch(clearParticipantCodes());` here - otherwise, there's an edge case where, if you delete yourself, and then click the "answer newdle" button in the header, you get your old participant code in the URL which no longer exists.

newdle/client/src/components/creation/userSearch/UserSearch.js
```diff
@@ -71,7 +71,12 @@ export default function UserSearch() {
           {participants.length !== 0 ? (
             <List selection relaxed>
               {participants.map(participant => (
-                <List.Item key={participant.email} className={styles['participant-list-item']}>
+                <List.Item
+                  // when creating a new newdle, all participants have an email, but when editing
+                  // we may also have participants without one - but everyone has an ID there
+                  key={participant.email || participant.id}
```
plourenco, line 77: We could probably just use `id` for everyone instead, right?

ThiefMaster: we don't have one for newly added participants who aren't in the db yet

plourenco: I was just going to check that now too, ok sounds good!

newdle/client/src/components/creation/ParticipantsStep.js
plourenco: Good point! Even while reinforcing it while clearing the participant code, we can also add some content to show the user's no longer a participant since this edge case is also applicable to users that might be have been invited before or with active codes.

ThiefMaster: showing a nicer error message for invalid participant codes sounds good but please not in this PR :)

newdle/api.py
```diff
@@ -307,52 +307,55 @@ def create_newdle(title, duration, timezone, timeslots, participants, private, n
...
+    if g.user is None or newdle.creator_uid != g.user['uid']:
         raise Forbidden
+    new_participants = []
+    if 'participants' in args:
+        participants = args.pop('participants')
+        # Filter the new participants to be created (excluding anonymous)
+        new_participants = {
+            Participant(**p)
+            for p in participants
+            if 'id' not in p and 'auth_uid' in p and p['auth_uid'] is not None
```
ThiefMaster, line 331:
```suggestion
            if 'id' not in p and p.get('auth_uid') is not None
```

newdle/schemas.py
```diff
@@ -126,6 +156,13 @@ def validate_timeslots(self, v):
             raise ValidationError('Time slots are not unique')
 
 
+class UpdateNewdleSchema(NewNewdleSchema):
+    participants = fields.List(
+        fields.Nested(NewParticipantSchema, unknown=EXCLUDE), missing=[]
```
ThiefMaster, line 161: `missing=[]` seems very dangerous here... it works fine because just like required it's ignored with `partial=True` but if someone removed that, it would delete all participants when finalizing an newdle. I think just removing the `missing` kwarg here doesn't change the behavior but avoids this possible pitfall.

newdle/schemas.py
```diff
@@ -44,29 +53,47 @@ def sort_users(self, data, many, **kwargs):
...
 
 
+class NewParticipantSchema(NewKnownParticipantSchema):
+    email = fields.String(allow_none=True)
+    auth_uid = fields.String(allow_none=True)
+    signature = fields.String(allow_none=True)
+
+    @validates_schema
+    def validate_participant(self, data, **kwargs):
+        if any(x in data and data[x] is not None for x in ['email', 'auth_uid']):
```
ThiefMaster, line 93:
```suggestion
        if any(x in data and data[x] is not None for x in ('email', 'auth_uid')):
```

newdle/schemas.py
```diff
@@ -44,29 +53,47 @@ def sort_users(self, data, many, **kwargs):
...
-            raise ValidationError("Participant's user signature is invalid!")
+    def validate_participant(self, data, **kwargs):
+        validate_signature(data, **kwargs)
 
     @post_load
     def remove_signature(self, data, **kwargs):
         """Remove signature, which is not needed after validation."""
-        del data['signature']
+        if 'signature' in data:
+            del data['signature']
```
ThiefMaster, line 82:
```suggestion
        data.pop('signature', None)
```

newdle/schemas.py
```diff
@@ -44,29 +53,47 @@ def sort_users(self, data, many, **kwargs):
...
 
     @validates_schema
-    def validate_signature(self, data, **kwargs):
-        data = dict(data)
-        signature = data.pop('signature')
-        auth_uid = data.pop('auth_uid')
-        if not check_user_signature(dict(data, uid=auth_uid), signature):
-            raise ValidationError("Participant's user signature is invalid!")
+    def validate_participant(self, data, **kwargs):
+        validate_signature(data, **kwargs)
```
ThiefMaster, line 76: why the `**kwargs`?

newdle/schemas.py
```diff
@@ -44,29 +53,47 @@ def sort_users(self, data, many, **kwargs):
 
 
 class NewUnknownParticipantSchema(mm.Schema):
+    id = fields.Integer()
```
ThiefMaster, line 56: This only makes sense when editing, but not when *creating* a new unknown participant (in that case it's inserting a new row, so letting the user pass the ID would be problematic). Especially since we pass `**args` to the model in `create_unknown_participant`, so it would actually allow you to set a custom ID.

newdle/api.py
plourenco: I guess I was focused on the logic too much 🤦‍♂️

newdle/schemas.py
plourenco: That's completely right, this went through so many iterations that I eventually forgot to remove it when copying, good point.

newdle/schemas.py
plourenco: Good point, I was undecided on whether to remove it or exclude it (`exclude=('id',)`) for the particular case of `create_unknown_participant`. I eventually think it makes more sense to exclude it since, by definition, an unknown participant is non-user but may or not be registered already.

newdle/schemas.py
plourenco: It was the 'not sure if we might need it'. But there's absolutely no point now.

newdle/api.py
ThiefMaster: that check still isn't needed ;)

newdle/schemas.py
ThiefMaster: Wouldn't it be more logical to have an `UnknownParticipantSchema` that contains the ID, and a `NewUnknownParticipantSchema` (superclass?) that doesn't? Sure, the `exclude` in the API function may work (I guess you tested it?), but having this done on the schema level feels safer and cleaner...

newdle/notifications.py
```diff
@@ -1,13 +1,37 @@
...
+        f'Invitation: {newdle.title}',
+        'invitation_email.txt',
+        'invitation_email.html',
+        lambda p: {
+            'creator': newdle.creator_name,
+            'title': newdle.title,
+            'answer_link': url_for(
+                'newdle', code=newdle.code, participant_code=p.code, _external=True
+            ),
+        },
```
ThiefMaster, line 19: The email uses the participant name as well:

```suggestion
        lambda p: {
            'creator': newdle.creator_name,
            'title': newdle.title,
            'participant': p.name,
            'answer_link': url_for(
                'newdle', code=newdle.code, participant_code=p.code, _external=True
            ),
        },
```

newdle/schemas.py
plourenco: Yes, I just wanted to avoid the cardinality of participant schemas from growing, as we already have: `NewUnknownParticipantSchema`, `NewKnownParticipantSchema`, `NewParticipantSchema` (plus eventually `UnknownParticipantSchema` or `KnownParticipantSchema`). But apart from that, it feels safer for me too.

newdle/notifications.py
plourenco: Indeed! And the `summary_link` was lost too when comparing with master, not sure how exactly.

ThiefMaster: the summary link isn't needed in these templates

ThiefMaster (approved):

PR #263, opened 2020-09-24
SQL Migrations printer

### LINKED ISSUE #238: Add GitHub action to automatically comment with the SQL for alembic revisions
This action should check if a PR adds new Alembic revisions. If that's the case, it should run `flask db upgrade --sql PREV:LATEST` with `LATEST` being the `revision` of the last revision and `PREV` being the `down_revision` of the first revision added in the PR and comment on the PR with something like this:

    This PR contains database changes. Before merging it, make sure to apply the migration in production:
    
    ```sql
    <sql here>
    ```

    When reviewing the PR, make sure that the changes will not break the previously deployed
    version, i.e. any new column needs to have a `server_default` or be nullable.

saibot94:
Fixes #238 

See the workflow in action in the following PR on my local fork: https://github.com/saibot94/newdle/pull/2

### DIFF
--- a/.github/workflows/migration-sql.yml
+++ b/.github/workflows/migration-sql.yml
@@ -1,0 +1,84 @@
+name: Print alembic migrations SQL
+
+on:
+  pull_request:
+    branches: [master, sql-actions]
+
+jobs:
+  print_sql_upgrades:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.head.repo.full_name == github.repository
+
+    strategy:
+      matrix:
+        python-version: [3.8]
+
+    services:
+      postgres:
+        image: postgres:12
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
+        ports:
+          - 5432
+        options: --health-cmd pg_isready --health-interval 5s --health-timeout 5s --health-retries 10
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Setup venv and config file
+        run: |
+          make config
+          make .venv
+          pip install -q -e '.[dev]' 
+
+      - name: Activate virtualenv for later steps
+        run: |
+          echo "::set-env name=VIRTUAL_ENV::$(pwd)/.venv"
+          echo "::add-path::$(pwd)/.venv/bin"
+
+      - name: Setup Database
+        run: |
+          sudo apt-get install -y -qq postgresql-client libpq-dev
+          export PGHOST=localhost
+          export PGPORT=${{ job.services.postgres.ports[5432] }}
+          export PGUSER=postgres
+          createuser newdletest
+          createdb newdle_tests
+          conn_string="postgresql://newdletest@localhost:${{ job.services.postgres.ports[5432] }}/newdle_tests"
+          sed -i "s%SQLALCHEMY_DATABASE_URI = .*%SQLALCHEMY_DATABASE_URI = '$conn_string'%g" newdle/newdle.cfg
+
+      - name: Get list of created migrations
+        uses: ThiefMaster/changed-files-action@json-output
+        with:
+          repo-token: ${{ github.token }}
+          pattern: '^newdle/migrations/versions/.+\.py$'
+
+      - name: Print changed migrations
+        id: get-comment-body
+        run: |
+          flask db upgrade;
+          migration_file=$(jq -r '.[-1] // empty' ~/files_created.json);
+          sql_command=$(./ci/print_revision_sql.sh $migration_file);
+          if [[ ! -z "$sql_command" ]] ; then
+            echo ::set-output name=body::$sql_command 
+          fi
+
+      - name: Create comment
+        uses: peter-evans/create-or-update-comment@v1
+        if: ${{ steps.get-comment-body.outputs.body }}
+        with:
+          issue-number: ${{github.event.number}}
+          body: |
+            This PR contains database changes. Before merging it, make sure to apply the migration in production:
+
+            ```sql
+            ${{ steps.get-comment-body.outputs.body }}
+            ```
+
+            When reviewing the PR, make sure that the changes will not break the previously deployed
+            version, i.e. any new column needs to have a `server_default` or be nullable.
--- a/.github/workflows/migration-sql.yml
+++ b/.github/workflows/migration-sql.yml
@@ -1,8 +1,8 @@
-name: Print alembic migrations SQL
+name: Alembic consistency check
 
 on:
   pull_request_target:
-    branches: [master, sql-actions]
+    branches: [master]
 
 jobs:
   check-alembic:
--- a/.github/workflows/migration-sql.yml
+++ b/.github/workflows/migration-sql.yml
@@ -5,7 +5,7 @@ on:
     branches: [master, sql-actions]
 
 jobs:
-  print_sql_upgrades:
+  check-alembic:
     runs-on: ubuntu-latest
 
     services:
--- a/.github/workflows/migration-sql.yml
+++ b/.github/workflows/migration-sql.yml
@@ -25,18 +25,33 @@ jobs:
         with:
           python-version: '3.8'
 
+      - name: Get list of created migrations
+        id: get-changed-files
+        uses: ThiefMaster/changed-files-action@json-output
+        with:
+          repo-token: ${{ github.token }}
+          pattern: '^newdle/migrations/versions/.+\.py$'
+
+      - name: Check if any migrations changed
+        id: migrations-changed
+        run: |
+          echo ::set-output name=changed::${{ steps.get-changed-files.outputs.files_created != '[]' || steps.get-changed-files.outputs.files_updated != '[]' || steps.get-changed-files.outputs.files_deleted != '[]' }}
+
       - name: Setup venv and config file
+        if: steps.migrations-changed.outputs.changed == 'true'
         run: |
           make config
           make env
           pip install -q -e '.[dev]'
 
       - name: Activate virtualenv for later steps
+        if: steps.migrations-changed.outputs.changed == 'true'
         run: |
           echo "::set-env name=VIRTUAL_ENV::$(pwd)/.venv"
           echo "::add-path::$(pwd)/.venv/bin"
 
       - name: Setup Database
+        if: steps.migrations-changed.outputs.changed == 'true'
         run: |
           sudo apt-get install -y -qq postgresql-client libpq-dev
           export PGHOST=localhost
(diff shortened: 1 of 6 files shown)


.github/workflows/migration-sql.yml
```diff
@@ -0,0 +1,84 @@
...
+    branches: [master, sql-actions]
+
+jobs:
+  print_sql_upgrades:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.head.repo.full_name == github.repository
+
+    strategy:
+      matrix:
+        python-version: [3.8]
```
ThiefMaster, line 14: Not needed; you can just specify `3.8` directly below when setting up Python

.github/workflows/migration-sql.yml
```diff
@@ -0,0 +1,84 @@
+name: Print alembic migrations SQL
+
+on:
+  pull_request:
+    branches: [master, sql-actions]
+
+jobs:
+  print_sql_upgrades:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.head.repo.full_name == github.repository
```
ThiefMaster, line 10: Won't this check always succeed in case of a PR?

.github/workflows/migration-sql.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Setup venv and config file
+        run: |
+          make config
+          make .venv
```
ThiefMaster, line 36: IIRC you wanted to add a "nicer" `make env` (or `make venv`) makefile target for this

.github/workflows/migration-sql.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Get list of created migrations
+        uses: ThiefMaster/changed-files-action@json-output
+        with:
+          repo-token: ${{ github.token }}
+          pattern: '^newdle/migrations/versions/.+\.py$'
+
+      - name: Print changed migrations
+        id: get-comment-body
+        run: |
+          flask db upgrade;
```
ThiefMaster, line 64: most of the semicolons here  and in the bash script aren't needed

ci/print_revision_sql.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# Get the latest migration from the file received as a parameter
+# and print it with escaped newlines
+set -e;
+
+
+migration_file=$1
```
ThiefMaster, line 7: This won't work if a PR contains more than one new alembic revision :/

saibot94: I take the the latest one, always. But I think you're saying that I should take the first migration added in the PR and print the SQL from its `down_revision` to the latest `revision`. So the script would need 2 files as a parameter which are the same in case you add a single migration.

What's your suggestion?

saibot94: Alright, changed the script and this is the output for something old up the the present revision:

```bash
older_migration=newdle/migrations/versions/20200914_1149_679eab4aab01_add_comment_column_to_participants.py
migration_file=newdle/migrations/versions/20200922_1759_43f0bf77bf97_add_newdle_creator_email_notify.py    

./ci/print_revision_sql.sh $older_migration $migration_file 
```
Output:

```
INFO  [alembic.runtime.migration] Generating static SQL
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade 93a638b96375 -> 679eab4aab01, Add comment column to participants
INFO  [alembic.runtime.migration] Running upgrade 679eab4aab01 -> 239d7862e2fe, Add last update field to newdle model
INFO  [alembic.runtime.migration] Running upgrade 239d7862e2fe -> 43f0bf77bf97, Add creator email and notify columns to newdle model
BEGIN;%0A%0A-- Running upgrade 93a638b96375 -> 679eab4aab01%0A%0AALTER TABLE participants ADD COLUMN comment VARCHAR DEFAULT '' NOT NULL;%0A%0AUPDATE alembic_version SET version_num='679eab4aab01' WHERE alembic_version.version_num = '93a638b96375';%0A%0A-- Running upgrade 679eab4aab01 -> 239d7862e2fe%0A%0AALTER TABLE newdles ADD COLUMN last_update TIMESTAMP WITHOUT TIME ZONE DEFAULT (now() at time zone 'utc')::timestamp NOT NULL;%0A%0AUPDATE newdles%0A SET last_update = final_dt%0A WHERE final_dt IS NOT NULL AND final_dt < now() AT TIME ZONE 'utc';;%0A%0AUPDATE alembic_version SET version_num='239d7862e2fe' WHERE alembic_version.version_num = '679eab4aab01';%0A%0A-- Running upgrade 239d7862e2fe -> 43f0bf77bf97%0A%0AALTER TABLE newdles ADD COLUMN creator_email VARCHAR DEFAULT '' NOT NULL;%0A%0AALTER TABLE newdles ADD COLUMN notify BOOLEAN DEFAULT 'false' NOT NULL;%0A%0AALTER TABLE newdles ALTER COLUMN notify SET DEFAULT 'true';%0A%0AUPDATE alembic_version SET version_num='43f0bf77bf97' WHERE alembic_version.version_num = '239d7862e2fe';%0A%0ACOMMIT;
```

.github/workflows/migration-sql.yml
saibot94: Removed, thanks.

.github/workflows/migration-sql.yml
saibot94: Changed to `make env`, thanks for reminding me.

.github/workflows/migration-sql.yml
saibot94: Removed.

.github/workflows/migration-sql.yml
saibot94: Removed and added directly

.github/workflows/migration-sql.yml
```diff
@@ -0,0 +1,80 @@
...
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
+        ports:
+          - 5432
+        options: --health-cmd pg_isready --health-interval 5s --health-timeout 5s --health-retries 10
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Python ${{ matrix.python-version }}
```
ThiefMaster, line 23:
```suggestion
      - name: Setup Python
```

.github/workflows/migration-sql.yml
```diff
@@ -0,0 +1,80 @@
...
+        uses: ThiefMaster/changed-files-action@json-output
+        with:
+          repo-token: ${{ github.token }}
+          pattern: '^newdle/migrations/versions/.+\.py$'
+
+      - name: Print changed migrations
+        id: get-comment-body
+        run: |
+          first_migration=$(jq -r '.[0] // empty' ~/files_created.json)
+          latest_migration=$(jq -r '.[-1] // empty' ~/files_created.json)
```
ThiefMaster, line 60: Does the action's output guarantee any particular order? If it's sorted by file system order it should be fine - while it may be wrong in case someone rebased a PR without renaming revisions (after a revision with a higher timestamp was added to master, this would be something to detect separately and fail CI (if you want to add this as a bonus feel free to do so btw! the filesystem order should always correspond to the actual revision order - you could probably parse the `flask db history` output for this)

.github/workflows/migration-sql.yml
```diff
@@ -0,0 +1,80 @@
...
+          sql_command=$(./ci/print_revision_sql.sh $first_migration $latest_migration)
+          if [ ! -z "$sql_command" ] ; then
+            echo ::set-output name=body::$sql_command 
+          fi
+
+      - name: Create comment
+        uses: peter-evans/create-or-update-comment@v1
+        if: ${{ steps.get-comment-body.outputs.body }}
+        with:
+          issue-number: ${{github.event.number}}
```
ThiefMaster, line 71: we have a mix of `${{foo}}` and `${{ foo }}` in the file; let's go for one option

.github/workflows/migration-sql.yml
saibot94: I added a small python script that does the check between the FS history and the history found by running flask db history. Found it more friendly than bash, hope you agree.

saibot94:
@ThiefMaster I think this could get another look.

ci/check_history_consistency.py
```diff
@@ -0,0 +1,47 @@
...
+
+def _get_file_revisions():
+    root_path = 'newdle/migrations/versions'
+    for f in sorted(os.listdir(root_path)):
+        if f.endswith('.py'):
+            file_path = os.path.join(root_path, f)
+            op = (
+                subprocess.check_output(
+                    ['grep', '-Po', "(?<=^revision = ').*(?=')", file_path]
+                )
```
ThiefMaster: I think it'd be cleaner to `execfile()` the file and get the `revision` from its globals

ci/check_history_consistency.py
```diff
@@ -0,0 +1,47 @@
...
+        .strip()
+        .split('\n')
+    )
+    pattern = re.compile(r'^.* -> ([a-zA-Z0-9]+)( \(head\))?,')
+    for line in output:
+        for obj in pattern.finditer(line):
+            yield obj.group(1)
+
+
+def _get_file_revisions():
```
ThiefMaster, line 23: since you yield the revisions i'd rename it to `_iter_file_revisions()`

saibot94: Renamed

ci/check_history_consistency.py
saibot94: Done, getting it from `global_vars['revision']` now.

ci/check_history_consistency.py
```diff
@@ -0,0 +1,39 @@
...
+    for pyfile in sorted(root_path.glob('*.py')):
+        results = pattern.findall(pyfile.read_text())
+        yield results[0]
+
+
+def _check_history_consistency():
+    revisions = list(_iter_db_revisions())
+    file_revisions = list(_iter_file_revisions())
+    revisions.reverse()
+    assert revisions == file_revisions
```
ThiefMaster, line 35: Tthe error in case of an inconsistency could be a bit more detailed ;)

package.json
```diff
@@ -9,13 +9,13 @@
     "black": "black setup.py conftest.py newdle/ tests/",
```
ThiefMaster, line 9: add `ci/` here and in isort as well?

ci/check_history_consistency.py
saibot94: Raising an exception with more information now.

package.json
saibot94: Added.

ThiefMaster (approved):

PR #264, opened 2020-09-25
CI action: Run linters only once

### LINKED ISSUE #258: CI action: Run linters only once
Let's only run them for the python 3.8 entry in the CI matrix.

Otherwise you get this kind of noise:

image

ParthS007:
closes #258

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,4 +60,5 @@ jobs:
           make test
 
       - name: Lint
+        if: matrix.python-version == 3.8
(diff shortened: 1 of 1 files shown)


ThiefMaster:
I wouldn't duplicate all this - why not just use an `if: ...` condition to only run the lint step for a specific matrix entry?

ParthS007:
okay, trying that out. 👍

ParthS007:
@ThiefMaster Added the `if`, works fine. 🚀

PR #302, opened 2020-10-03
add i18n: Timeline

andramazo:


ThiefMaster (approved):
I added one string that was missing.

PR #303, opened 2020-10-03
i18n: DurationPicker

andramazo:
Added i18n for the two components - TimezonePicker and DurationPicker. I separated from the issue #284.

### DIFF
--- a/newdle/client/src/components/common/TimezonePicker.js
+++ b/newdle/client/src/components/common/TimezonePicker.js
@@ -1,5 +1,6 @@
 import React, {useMemo} from 'react';
 import PropTypes from 'prop-types';
+import {Trans} from '@lingui/macro';
 import {commonTimezones} from '../../util/timezones';
 import LazyDropdown from '../LazyDropdown';
 import styles from './TimezonePicker.module.scss';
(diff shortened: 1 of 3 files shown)


newdle/client/src/components/common/TimezonePicker.js
```diff
@@ -17,7 +18,7 @@ function TimezonePicker({onChange, currentTz, title, ...props}) {
   );
   return (
     <div>
-      {title}
+      <Trans>{title}</Trans>
```
ThiefMaster, line 21: This makes no sense, you cannot translate a string that is not known beforehand. This file has no translatable content (which is why there's no issue for it).

newdle/client/src/components/creation/timeslots/DurationPicker.js
```diff
@@ -20,9 +21,9 @@ function _minutesToHM(minutes) {
   }
   minutes = minutes % 60;
   if (!minutes) {
-    return hours === 1 ? `${hours} hour` : `${hours} hours`;
+    return hours === 1 ? t`${hours} hour` : t`${hours} hours`;
```
ThiefMaster, line 24: Use proper pluralization here

ThiefMaster (approved):
I fixed the issues I just reported via comments so I can merge it. Please check my changes nonetheless since I believe it's useful for you to understand *why* they had to be changed (feel free to ask for clarification in comments if necessary)

PR #314, opened 2020-10-06
Move the initials to the client side

plourenco:
Derive the initials from the participant name instead of serving it from the server-side.

### DIFF
--- a/newdle/client/src/components/UserAvatar.js
+++ b/newdle/client/src/components/UserAvatar.js
@@ -12,9 +12,8 @@ const defaultArgs = [
   true, // rounded
 ].join('/');
 
-function UserAvatar({user: {email, name, initials}, className, withLabel, size}) {
-  initials = initials.replace(/[/?#]/g, '');
-  const uri = `https://ui-avatars.com/api/${initials}/${size}/${defaultArgs}`;
+function UserAvatar({user: {email, name}, className, withLabel, size}) {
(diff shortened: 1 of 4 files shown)


newdle/client/src/components/UserAvatar.js
```diff
@@ -12,8 +12,8 @@ const defaultArgs = [
   true, // rounded
 ].join('/');
 
-function UserAvatar({user: {email, name, initials}, className, withLabel, size}) {
-  initials = initials.replace(/[/?#]/g, '');
+function UserAvatar({user: {email, name}, className, withLabel, size}) {
+  const initials = name && name.length ? name[0] : 'N';
```
ThiefMaster, line 16: Don't we always have a name here?

plourenco: I suppose we do, but by default I was trying to be defensive.

ThiefMaster: I don't think there's a decent fallback though :/ the random-ish char feels weird there.
Checking `name.length` isn't needed, empty strings are falsy.

plourenco: In the very unlikely event of having an empty string there, it will return `undefined`. The random-char was based on "N" from Newdle. But I can always ignore that case since I don't expect it to happen.

ThiefMaster: yeah i'd go for ignoring it

ThiefMaster (approved):

PR #315, opened 2020-10-07
Proxy gravatar requests via newdle

### LINKED ISSUE #139: Support using an image proxy for gravatars
https://github.com/atmos/camo/ is the most common one (GitHub uses it, Mattermost as well),

This should be fully optional of course - when disabled, gravatars are loaded directly, otherwise through the proxy.

Advantages?

- no tracking cookies
- gravatar never sees user IPs

Disadvantages?

- one more thing to deploy (but easy e.g. on openshift)
- to avoid having an open proxy, we need to either include the proxy url in all responses that will be used to show avatars, or just have an endpoint which redirects to the URL (could take the same args - ie email + initials - as gravatar, no need for md5 there since we have the email client-side anyway) and always use this in the img tags. but we can use this regardless of whether the proxy used or not; if not we'd simply redirect to gravatar

The redirect option has one extra request so caching in the browser may be a bit worse, while the option of returning an `avatar_url` in the API responses adds more "noise" to the API responses - but they are reasonable small so I think avoiding the extra redirect would be cleaner.

Running the imageproxy should be easily possible via Docker, but if you are an external contributor  you are also very welcome to simply implement the [url building](https://github.com/atmos/camo#url-formats) and we then test it with the PR. We'd prefer the `https://example.org/<digest>/<image-url>` URL style by the way.

mic4ael:
One thing to consider:

- [x] right now, all the requests to `/api/avatar/<email>` don't require authentication which is not ideal. It is not possible to attach the `Authorization` header when fetching the image, thus maybe it could be a good idea to put the token in a cookie?  

closes #139

### DIFF
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -1,23 +1,28 @@
+import hashlib
 import uuid
 from importlib import import_module
 
+import requests
 from faker import Faker
-from flask import Blueprint, current_app, g, jsonify, request, url_for
+from flask import Blueprint, Response, current_app, g, jsonify, request, url_for
 from itsdangerous import BadData, SignatureExpired
 from marshmallow import fields
 from marshmallow.validate import OneOf
 from pytz import common_timezones_set, timezone
 from sqlalchemy.orm import selectinload
 from werkzeug.exceptions import Forbidden, ServiceUnavailable, UnprocessableEntity
+from werkzeug.urls import url_encode
 
 from .calendar import create_calendar_event
 from .core.auth import search_users, user_info_from_app_token
 from .core.db import db
 from .core.util import (
     DATE_FORMAT,
+    avatar_info_from_payload,
     change_dt_timezone,
     format_dt,
     range_union,
+    render_user_avatar,
     sign_user,
 )
 from .core.webargs import abort, use_args, use_kwargs
@@ -75,6 +80,44 @@ def require_token():
     except BadData:
         return jsonify(error='token_invalid'), 401
     g.user = user
+
+
+@api.route('/avatar/<payload>')
+@allow_anonymous
+def user_avatar(payload):
+    user_info = avatar_info_from_payload(payload)
+    size = request.args.get('size')
+    email_hex = hashlib.md5(user_info['email'].lower().encode()).hexdigest()
+    # make gravatar return 404 HTTP code instead of a default image
+    query_args = {'d': '404'}
+    if size is not None:
+        query_args['s'] = size
+
+    gravatar_url = f'https://gravatar.com/avatar/{email_hex}?{url_encode(query_args)}'
+    request_headers = {}
+    if 'if-modified-since' in request.headers:
+        request_headers['if-modified-since'] = request.headers['if-modified-since']
+
+    resp = requests.get(gravatar_url, headers=request_headers)
+    if resp.status_code == 404:
+        return render_user_avatar(user_info['initial'], size)
+
+    # include only the headers we care about
+    forwarded_header_names = [
+        'content-type',
+        'last-modified',
+        'cache-control',
+        'expires',
+        'date',
+        'content-disposition',
+        'etag',
+    ]
+    headers = [
+        (name, value)
+        for (name, value) in resp.headers.items()
+        if name.lower() in forwarded_header_names
+    ]
+    return Response(resp.content, resp.status_code, headers)
 
 
 @api.route('/ping')
--- a/newdle/api.py
+++ b/newdle/api.py
@@ -86,6 +86,8 @@ def require_token():
 @allow_anonymous
 def user_avatar(payload):
     user_info = avatar_info_from_payload(payload)
+    if user_info is None:
+        abort(404)
     size = request.args.get('size')
     email_hex = hashlib.md5(user_info['email'].lower().encode()).hexdigest()
     # make gravatar return 404 HTTP code instead of a default image
--- a/newdle/core/util.py
+++ b/newdle/core/util.py
@@ -4,7 +4,7 @@ from enum import Enum
 from io import BytesIO
 
 from flask import current_app, render_template, send_file
-from itsdangerous import Signer, URLSafeSerializer, URLSafeTimedSerializer
+from itsdangerous import BadData, Signer, URLSafeSerializer, URLSafeTimedSerializer
 from werkzeug.local import LocalProxy
 
 
@@ -117,4 +117,7 @@ def avatar_payload_from_user_info(user_info):
 
 
 def avatar_info_from_payload(payload):
-    return secure_serializer.loads(payload, salt='avatar-payload')
+    try:
+        return secure_serializer.loads(payload, salt='avatar-payload')
+    except BadData:
+        return None
--- a/newdle/core/util.py
+++ b/newdle/core/util.py
@@ -1,13 +1,23 @@
 import re
 from datetime import datetime, time
 from enum import Enum
+from io import BytesIO
 
-from flask import current_app
-from itsdangerous import Signer
+from flask import current_app, render_template, send_file
+from itsdangerous import Signer, URLSafeSerializer, URLSafeTimedSerializer
+from werkzeug.local import LocalProxy
 
 
 DATE_FORMAT = '%Y-%m-%d'
 DATETIME_FORMAT = '%Y-%m-%dT%H:%M'
+
+secure_serializer = LocalProxy(
+    lambda: URLSafeSerializer(current_app.config['SECRET_KEY'], b'newdle')
+)
+
+secure_timed_serializer = LocalProxy(
+    lambda: URLSafeTimedSerializer(current_app.config['SECRET_KEY'], b'newdle')
+)
 
 
 class AutoNameEnum(Enum):
@@ -89,3 +99,22 @@ def find_overlap(day, start, end, tz):
 def dedent(value, *, _re=re.compile(r'^ +', re.MULTILINE)):
     """Remove leading whitespace from each line."""
     return _re.sub('', value)
+
+
+def render_user_avatar(initial, size):
+    avatar = render_template('avatar.svg', text=initial.upper(), size=size)
+    return send_file(BytesIO(avatar.encode()), mimetype='image/svg+xml')
+
+
+def avatar_payload_from_user_info(user_info):
+    return secure_serializer.dumps(
+        {
+            'email': user_info['email'],
+            'initial': user_info['first_name'][0].upper(),
+        },
+        salt='avatar-payload',
+    )
+
+
+def avatar_info_from_payload(payload):
+    return secure_serializer.loads(payload, salt='avatar-payload')
--- a/newdle/schemas.py
+++ b/newdle/schemas.py
@@ -12,7 +12,11 @@ from marshmallow_enum import EnumField
 from pytz import common_timezones_set
 
 from .core.marshmallow import mm
-from .core.util import DATETIME_FORMAT, check_user_signature
+from .core.util import (
+    DATETIME_FORMAT,
+    avatar_payload_from_user_info,
+    check_user_signature,
+)
 from .models import Availability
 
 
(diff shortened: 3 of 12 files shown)


newdle/api.py
```diff
@@ -77,6 +81,38 @@ def require_token():
...
+@allow_anonymous
+def user_avatar(email):
+    size = request.args.get('size')
+    email_hex = hashlib.md5(email.encode()).hexdigest()
+    # make gravat return 404 HTTP code instead of a default image
+    query_args = {'d': '404'}
+    if size is not None:
+        query_args['s'] = size
+    gravatar_url = (
+        f'https://gravatar.com/avatar/{quote(email_hex)}?' f'{urlencode(query_args)}'
```
ThiefMaster, line 94: Why two separate f-strings?

newdle/api.py
```diff
@@ -77,6 +81,38 @@ def require_token():
     g.user = user
 
 
+@api.route('/avatar/<email>')
+@allow_anonymous
+def user_avatar(email):
+    size = request.args.get('size')
+    email_hex = hashlib.md5(email.encode()).hexdigest()
+    # make gravat return 404 HTTP code instead of a default image
```
ThiefMaster, line 89: *gravatar

newdle/api.py
```diff
@@ -77,6 +81,38 @@ def require_token():
...
+    query_args = {'d': '404'}
+    if size is not None:
+        query_args['s'] = size
+    gravatar_url = (
+        f'https://gravatar.com/avatar/{quote(email_hex)}?' f'{urlencode(query_args)}'
+    )
+    resp = requests.get(gravatar_url)
+    if resp.status_code == 404:
+        user = search_users(None, email, 1)[1][0]
```
ThiefMaster, line 98: pass the initial via URL - doing a search here kills performance

newdle/api.py
```diff
@@ -77,6 +81,38 @@ def require_token():
...
+        'content-encoding',
+        'content-length',
+        'transfer-encoding',
+        'connection',
+    ]
+    headers = [
+        (name, value)
+        for (name, value) in resp.raw.headers.items()
+        if name.lower() not in excluded_headers
+    ]
```
ThiefMaster, line 112: I guess this is needed to handle last-modified-since etc. correctly? Otherwise I'd just use a whitelist and only forward relevant headers (e.g. content-type)

newdle/schemas.py
```diff
@@ -21,11 +21,14 @@ class UserSchema(mm.Schema):
     name = fields.Function(lambda u: f'{u["first_name"]} {u["last_name"]}')
     initials = fields.Function(lambda u: f'{u["first_name"][0]} {u["last_name"][0]}')
     uid = fields.String()
+    avatar_url = fields.Function(
+        lambda user: url_for('api.user_avatar', email=user['email'])
+    )
```
ThiefMaster, line 26: Do we still need this? We have access to both name and email on the client side anyway.

Anyway, the fact that we have it here means that the topic of authentication we discussed earlier suddenly becomes super easy: You can just include a signature in the URL and use this to avoid creating an unrestricted image proxy.

newdle/core/util.py
```diff
@@ -89,3 +90,10 @@ def find_overlap(day, start, end, tz):
 def dedent(value, *, _re=re.compile(r'^ +', re.MULTILINE)):
     """Remove leading whitespace from each line."""
     return _re.sub('', value)
+
+
+def render_user_avatar(user, size):
+    avatar = render_template(
+        'avatar.svg', text=user['first_name'][0].upper(), size=size
+    )
+    return send_file(BytesIO(avatar.encode('utf-8')), mimetype='image/svg+xml')
```
ThiefMaster, line 99: `.encode()` should be enough in python 3

newdle/api.py
```diff
@@ -1,8 +1,11 @@
+import hashlib
 import uuid
 from importlib import import_module
+from urllib.parse import quote, urlencode
```
ThiefMaster, line 4: I'd use the versions from `werkzeug.urls` (`url_quote` / `url_encode` I guess)

newdle/api.py
```diff
@@ -77,6 +81,38 @@ def require_token():
     g.user = user
 
 
+@api.route('/avatar/<email>')
```
ThiefMaster, line 84: Regarding my suggestion with the signed URLs below: You could use `/avatar/<payload>` and pass both email and initials in the signed string created by the `secure_serializer`.

newdle/api.py
ThiefMaster, line 94: an md5 hexdigest is guaranteed to be only `[a-f0-9]`; no need for `quote()` there

newdle/api.py
```diff
@@ -77,6 +81,38 @@ def require_token():
...
+    size = request.args.get('size')
+    email_hex = hashlib.md5(email.encode()).hexdigest()
+    # make gravat return 404 HTTP code instead of a default image
+    query_args = {'d': '404'}
+    if size is not None:
+        query_args['s'] = size
+    gravatar_url = (
+        f'https://gravatar.com/avatar/{quote(email_hex)}?' f'{urlencode(query_args)}'
+    )
+    resp = requests.get(gravatar_url)
```
ThiefMaster, line 96: what about forwarding headers from the client in order to handle client-side caching (if-modified-since) nicely?

newdle/api.py
mic4ael: Oh.. I think it is because I had some warning about max line length. I split it into two lines but then black must have had changed it anyway.

newdle/api.py
mic4ael: I was thinking about it as well. Would it be ok to something like `/api/avatar/<email>/<initial>`?

newdle/core/util.py
mic4ael: Oh.. true! I forgot to remove it.

newdle/api.py
mic4ael: Out of curiosity, why? Do they provide something more than the ones from stdlib?

newdle/api.py
mic4ael: This is also to prevent a case in which we get `transfer-encoding` == gzip. Flask decompresses the response body for us, so we either have to gzip the response again or get rid of the headers that might cause some trouble.

newdle/api.py
mic4ael: yeah, would you forward all of them?

ThiefMaster: forwarding headers relevant for caching should probably be enough... (i'd have a look at the headers your browser sends when you just open (no f5 reload, since both that and a hard ctrl+f5 reload change the behavior) a page embedding a gravatar that's already cached)

newdle/api.py
ThiefMaster: The question was more about why we need to forward response headers in the first place, not why we filter some out ;)

newdle/core/util.py
```diff
@@ -89,3 +95,22 @@ def find_overlap(day, start, end, tz):
...
+def render_user_avatar(initials, size):
+    avatar = render_template('avatar.svg', text=initials.upper(), size=size)
+    return send_file(BytesIO(avatar.encode()), mimetype='image/svg+xml')
+
+
+def avatar_payload_from_user_info(user_info):
+    return secure_serializer.dumps(
+        {
+            'email': user_info['email'],
+            'initials': user_info['first_name'][0].upper(),
```
ThiefMaster, line 109: `initial` vs `initials`?

newdle/core/util.py
```diff
@@ -1,14 +1,20 @@
...
+from itsdangerous import Signer, URLSafeSerializer
+from werkzeug.local import LocalProxy
 
 
 DATE_FORMAT = '%Y-%m-%d'
 DATETIME_FORMAT = '%Y-%m-%dT%H:%M'
 
+secure_serializer = LocalProxy(
+    lambda: URLSafeSerializer(current_app.config['SECRET_KEY'], b'newdle')
+)
```
ThiefMaster, line 16: What about moving the one from auth.py in here as well (and renaming it to `secure_timed_serializer` to make clear that we have to different ones?

newdle/core/util.py
```diff
@@ -89,3 +95,22 @@ def find_overlap(day, start, end, tz):
...
+    return secure_serializer.dumps(
+        {
+            'email': user_info['email'],
+            'initials': user_info['first_name'][0].upper(),
+        },
+        salt='avatar-payload',
+    )
+
+
+def user_info_from_payload(payload):
```
ThiefMaster, line 115: `avatar_info_from_payload`? since it's not really the whole user info, just the stuff relevant to the avatar..

newdle/core/util.py
mic4ael: yeah, actually did that locally but didn't push it

newdle/api.py
mic4ael: Ok then. Let's then agree on some necessary headers we might want to forward. 'Content-Type', 'Last-Modified', 'Cache-Control', 'Expires', 'Date', 'Content-Disposition'?

ThiefMaster: etag (or e-tag?) as well

newdle/api.py
```diff
@@ -77,6 +82,45 @@ def require_token():
     g.user = user
 
 
+@api.route('/avatar/<payload>')
+@allow_anonymous
+def user_avatar(payload):
+    user_info = avatar_info_from_payload(payload)
+    size = request.args.get('size')
+    email_hex = hashlib.md5(user_info['email'].encode()).hexdigest()
```
ThiefMaster, line 90:
```suggestion
    email_hex = hashlib.md5(user_info['email'].lower().encode()).hexdigest()
```

newdle/api.py
```diff
@@ -77,6 +82,45 @@ def require_token():
...
+    query_args = {'d': '404'}
+    if size is not None:
+        query_args['s'] = size
+
+    gravatar_url = f'https://gravatar.com/avatar/{email_hex}?{url_encode(query_args)}'
+    request_headers = {
+        header: value
+        for header, value in request.headers.items()
+        if header in ['if-modified-since']
+    }
```
ThiefMaster, line 101:
```suggestion
    request_headers = {}
    if 'If-Modified-Since' in request.headers:
        request_headers['If-Modified-Since'] = request.headers['If-Modified-Since']
```

ThiefMaster: maybe `request_headers = {'If-Modified-Since': request.headers.get('If-Modified-Since')}` would work as well, not sure how requests handles a `None` there...

newdle/api.py
```diff
@@ -77,6 +82,45 @@ def require_token():
...
+    resp = requests.get(gravatar_url, headers=request_headers)
+    if resp.status_code == 404:
+        return render_user_avatar(user_info['initial'], size)
```
ThiefMaster, line 104: We can add some strong caching headers here, since the fallback never changes

newdle/api.py
```diff
@@ -77,6 +82,45 @@ def require_token():
...
+        'last-modified',
+        'cache-control',
+        'expires',
+        'date',
+        'content-disposition',
+        'etag',
+    ]
+    headers = [
+        (name, value)
+        for (name, value) in resp.raw.headers.items()
```
ThiefMaster, line 118: What's the difference between `resp.raw.headers` and `resp.headers`?

newdle/core/util.py
```diff
@@ -89,3 +99,22 @@ def find_overlap(day, start, end, tz):
 def dedent(value, *, _re=re.compile(r'^ +', re.MULTILINE)):
     """Remove leading whitespace from each line."""
     return _re.sub('', value)
+
+
+def render_user_avatar(initials, size):
+    avatar = render_template('avatar.svg', text=initials.upper(), size=size)
```
ThiefMaster, line 105:
```suggestion
def render_user_avatar(initial, size):
    avatar = render_template('avatar.svg', text=initial.upper(), size=size)
```

newdle/api.py
mic4ael: we could also just change the `send_file` cache_timeout from the default value to something more appropriate

ThiefMaster: ah, I wasn't aware this is using `send_file`. then I think it may already add cache-related headers..

mic4ael: it only adds cache-control ;)

ThiefMaster: caching seems to work fine in the end, so I don't think we need to change anything :)

ThiefMaster (approved):

PR #322, opened 2020-10-09
Stop using deprecated APIs in GitHub actions

### LINKED ISSUE #321: Update GitHub actions to use new env/path utils
We currently use `set-env` and `add-path` in our GitHub actions and those have been deprecated. Let's switch to their successor.

- https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
- https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

nagy3n:
closes #321

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,8 +51,8 @@ jobs:
 
       - name: Activate virtualenv for later steps
         run: |
-          echo "::set-env name=VIRTUAL_ENV::$(pwd)/.venv"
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Thanks! GitHub seems to be having some issues atm, but I'll merge it as soon as the CI runs successfully here.

ThiefMaster (approved):

PR #324, opened 2020-10-11
Add i18n for MyNewdles

### LINKED ISSUE #278: Add i18n: MyNewdles
We'd like to make the newdle frontend i18n-capable.
This issue is about adding i18n support to the `MyNewdles` component (usually this is in a file with the same name).

Please test your changes, especially if you update a string that needs pluralization (we did not add pluralization so far because we didn't have i18n and doing it without an i18n library is not pretty).

Useful resources:

- The project's [README](https://github.com/indico/newdle#readme) explains how to run newdle
- **[Here's an example PR](https://github.com/indico/newdle/pull/267/files) where we added i18n to some components.** It contains [React usage](https://github.com/indico/newdle/pull/267/files#diff-0cc54461cea891fa78c0e85c7d6df434R172) (`<Trans>`), [string usage](https://github.com/indico/newdle/pull/267/files#diff-e5661acb3a7ba762bb8141855a9ccf73R191) (template literal tagged with `t`) and [pluralization](https://github.com/indico/newdle/pull/267/files#diff-0cc54461cea891fa78c0e85c7d6df434R153-R157).
- There's also the [Lingui documentation](https://js-lingui-git-next.lingui-js.now.sh/) but be aware that this is not fully up to date for the version we are using (Lingui 3)
- We do not use artificial IDs (like `newdle.created.message`); the source strings are always the English version
- When adding the lingui-related import (from `'@lingui/macro'`) make sure to put this import before any relative `./something` imports
- You do not need to update any .po files manually; you can run `make i18n` or simply `git commit` - we have a pre-commit hook which automatically updates the po file and stages it to be committed as well.

In case the component in this issue is rather small, feel free to take care of multiple components in a single PR. Don't forget to add `closes #xxx, closes #yyy` to your PR so all relevant issues are closed.

Thanks in advance for making a useful contribution during this [Hacktoberfest](https://hacktoberfest.digitalocean.com/)! 🎉

saibot94:
closes #278

### DIFF
--- a/newdle/client/src/components/MyNewdles.js
+++ b/newdle/client/src/components/MyNewdles.js
@@ -2,6 +2,7 @@ import React, {useEffect} from 'react';
 import {useDispatch} from 'react-redux';
 import {useHistory} from 'react-router';
 import {Link} from 'react-router-dom';
+import {Trans, t} from '@lingui/macro';
 import PropTypes from 'prop-types';
 import {Container, Icon, Label, Placeholder} from 'semantic-ui-react';
 import {clearParticipantCodes} from '../actions';
@@ -42,7 +43,7 @@ export default function MyNewdles() {
           <div className={styles['new-newdle']}>
             <h3 className={styles.title}>
               {/* eslint-disable-next-line jsx-a11y/accessible-emoji */}
-              {newdles.length === 0 ? 'Create your first newdle! 🍜' : 'Create a newdle! 🍜'}
+              {newdles.length === 0 ? t`Create your first newdle! 🍜` : `Create a newdle! 🍜`}
             </h3>
             <div>
               <Icon name="arrow right" size="large" color="purple" />
@@ -79,7 +80,7 @@ function MyNewdle({newdle: {code, title, participants, duration, final_dt: final
           {title}
         </a>
         <Label color={finalDT ? 'blue' : 'green'} size="tiny" className={styles.state}>
-          {finalDT ? 'finished' : 'ongoing'}
+          {finalDT ? t`finished` : t`ongoing`}
         </Label>
       </h3>
       {finalDT && (
@@ -100,11 +101,15 @@ function MyNewdle({newdle: {code, title, participants, duration, final_dt: final
         <Icon name="users" />
         <label>
           <em>
-            {!participants.length
-              ? 'There are no participants yet'
-              : answers.length !== participants.length
-              ? `${answers.length} out of ${participants.length} participants answered`
-              : 'All participants answered'}
+            {!participants.length ? (
+              t`There are no participants yet.`
+            ) : answers.length !== participants.length ? (
+              <Trans>
+                {answers.length} out of {participants.length} participants answered
+              </Trans>
+            ) : (
+              t`All participants answered.`
+            )}
           </em>
         </label>
       </div>
--- a/newdle/client/src/components/MyNewdles.js
+++ b/newdle/client/src/components/MyNewdles.js
@@ -2,7 +2,7 @@ import React, {useEffect} from 'react';
 import {useDispatch} from 'react-redux';
 import {useHistory} from 'react-router';
 import {Link} from 'react-router-dom';
-import {Trans, t} from '@lingui/macro';
+import {Trans, t, Plural} from '@lingui/macro';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


newdle/client/src/components/MyNewdles.js
```diff
@@ -100,11 +101,15 @@ function MyNewdle({newdle: {code, title, participants, duration, final_dt: final
...
-              ? 'There are no participants yet'
-              : answers.length !== participants.length
-              ? `${answers.length} out of ${participants.length} participants answered`
-              : 'All participants answered'}
+            {!participants.length ? (
+              t`There are no participants yet.`
+            ) : answers.length !== participants.length ? (
+              <Trans>
+                {answers.length} out of {participants.length} participants answered
+              </Trans>
```
ThiefMaster, line 109: This can use `<Plural ... />`

newdle/client/src/components/MyNewdles.js
```diff
@@ -42,7 +43,7 @@ export default function MyNewdles() {
           <div className={styles['new-newdle']}>
             <h3 className={styles.title}>
               {/* eslint-disable-next-line jsx-a11y/accessible-emoji */}
-              {newdles.length === 0 ? 'Create your first newdle! 🍜' : 'Create a newdle! 🍜'}
+              {newdles.length === 0 ? t`Create your first newdle! 🍜` : `Create a newdle! 🍜`}
```
ThiefMaster, line 46: I'd move the emoji outside the translated string

newdle/client/src/components/MyNewdles.js
```diff
@@ -100,11 +101,19 @@ function MyNewdle({newdle: {code, title, participants, duration, final_dt: final
...
-              ? `${answers.length} out of ${participants.length} participants answered`
-              : 'All participants answered'}
+            {!participants.length ? (
+              t`There are no participants yet.`
+            ) : answers.length !== participants.length ? (
+              <Trans>
+                <Plural
+                  value={participants.length}
+                  one={`${answers.length} out of # participant answered`}
+                  other={`${answers.length} out of # participants answered`}
```
ThiefMaster, line 111: I think it's fine now, but if we ever end up a language where the number of people who answered also affects the translation we'd probably need to nest yet another plural inside each of these two options

newdle/client/src/components/MyNewdles.js
```diff
@@ -100,11 +101,19 @@ function MyNewdle({newdle: {code, title, participants, duration, final_dt: final
...
-            {!participants.length
-              ? 'There are no participants yet'
-              : answers.length !== participants.length
-              ? `${answers.length} out of ${participants.length} participants answered`
-              : 'All participants answered'}
+            {!participants.length ? (
+              t`There are no participants yet.`
+            ) : answers.length !== participants.length ? (
+              <Trans>
+                <Plural
```
ThiefMaster, line 108: you can add `_0="There are no participants yet."` to cover the no-participants case

saibot94: Hmm if I add that and change the logic around, you either get `All participants answered.` for 0 participants or 2 messages when everybody answered. (`x out of x answered` and `All participants answered`). To keep the display like this it's easier to leave the current logic.

ThiefMaster: ah, i didn't realize there's yet another condition.. you are right, then it won't work... [Select](https://js-lingui-git-next.lingui-js.now.sh/ref/macro.html#jsxmacro-Select) could be an option but probably overkill here since the separate strings work fine

newdle/client/src/components/MyNewdles.js
```diff
@@ -100,11 +101,19 @@ function MyNewdle({newdle: {code, title, participants, duration, final_dt: final
...
+                  value={participants.length}
+                  one={`${answers.length} out of # participant answered`}
```
ThiefMaster, line 110: Actually this case will never happen because you never get in this branch in case of `answers.length === participants.length`. Not sure if omitting the `one` prop works, but if yes we can probably do that

saibot94: If I remove it then newdles with 1 participant that don't have any answer yet get no text at all.

ThiefMaster: Indeed, I thought we had a "Nobody has answered yet" case but we don't

ThiefMaster (approved):

Issue #309, opened 2020-10-05
Switching to "View Summary" view resets the answers


avivace:
If have selected the answers but not submitted them, and click on "View summary" on the top, the answers get reset without any warning.

![image](https://user-images.githubusercontent.com/14352721/95084684-7fd4d500-071e-11eb-883e-121424e3bde1.png)

ThiefMaster:
Maybe we can use the `<UnloadPrompt />` component on this page? PR welcome :)

Caetan95:
I can do it. Just an alert when changing view _"Are you sure you want to leave this page without saving?"_?

ThiefMaster:
Yes, that's what `UnloadPrompt` does ;)

PR #325, opened 2020-10-13
Switching to "View Summary" view resets the answers #309

Caetan95:
For the first answer an alert will be displayed in all this cases:
- Alert display when we change from Unavailable to other option and we try to change the view.
- Alert display when something is typed in the comment textbox and we try to change the view.

For the following answers (update answer) an alert will be displayed in all this cases:
- Alert display when we change to a different option and we try to change the view.
- Alert display when comment is changed and we try to change the view.

Alert is not displayed when we submit.


newdle/client/src/components/NewdleTitle.js
```diff
@@ -51,6 +52,7 @@ export default function NewdleTitle({
                     disabled={finished}
                   >
                     <Icon name="calendar plus outline" />
+                    <UnloadPrompt />
```
ThiefMaster, line 55: This component has an `active` prop which is required - and it should only be set to `true` when there are changes where you do want the prompt to occur (ie if there are any changed that would actually be lost when changing pages).

It may also be better to move this to `AnswerPage` since it's only needed there. And you already have `participantAnswersChanged` in there to check whether the unload prompt should be active or not. Best to check for `comment !== participant.comment` as well so starting to write a comment also activates the unload prompt.

ThiefMaster (changes requested):

newdle/client/src/components/NewdleTitle.js
Caetan95: I was working on it, but I think that there are some problems. The first time you use it (with non previous answers), const ```participantAnswersChanged``` is always false because there were no previous answers, this const is not the "real time" value, but the stored one in the previous answer.
Also is not possible to access ```comment``` const ```Cannot read property 'comment' of null``` since there are not previous comments.
In this case this piece of code:
```
  const submitDisabled =
    submitting ||
    !canSubmit ||
    (participantCode && !participant) ||
    (participantHasAnswers && !participantAnswersChanged && comment === participant.comment);
````
works the first time because ```comment``` is never evaluated (since it is the last part it never reach it), just changing the order ```(comment === participant.comment && participantHasAnswers && !participantAnswersChanged);``` the code breaks with the same error (```Cannot read property 'comment' of null```).
For these reasons I think that the first time we have to choose between prompt the alert in all cases or allow to switch and submit without any alert. The following answers it is possible to do what we want, but as I said, I think that the first one is impossible.

ThiefMaster: Like I said, move it into the AnswerPage component ;)

And I mean we use the logic to enable the submit button, so I don't think it's impossible.... `participantAnswersChanged` should work in any case, because if you look at the selector the initial answers are an empty dict if there's no participant.

ThiefMaster (approved):

PR #326, opened 2020-10-15
Add footer

### LINKED ISSUE #242: Add footer

VirtualEvan:
Closes #242 

This commit adds a footer to the project including
- Configurable links
- Info about the project
- A language selector

Still missing the feature to add the links directly from the config file, but I think we can start having a look into this meanwhile.
Let me know if any change is needed

Preview:
image
image


newdle/client/src/components/Footer.js
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <footer className={styles.footer}>
+      <Grid columns={3}>
+        <Grid.Row verticalAlign="middle">
+          <Grid.Column computer={4} tablet={4} only="computer tablet">
+            <ProjectInfo />
+          </Grid.Column>
+          <Grid.Column mobile={16} tablet={8} computer={8}>
+            <p className={styles.links}>
+              {Object.keys(FOOTER_LINKS).map((key, pos, arr) => (
```
ThiefMaster, line 24: `Object.entries()` gives you access to keys and values at the same time

newdle/client/src/components/Footer.js
```diff
@@ -0,0 +1,72 @@
...
+                <React.Fragment key={key}>
+                  <a
+                    href={FOOTER_LINKS[key]}
+                    className={styles.link}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                  >
+                    {key}
+                  </a>
+                  {pos < arr.length - 1 ? <span className={styles.separator}>{' | '}</span> : null}
```
ThiefMaster, line 34: an alternative to this could be a `<ul>` that's styled as an inline list with `content: ' | '` in the elements' `::before` (using a `li + li` selector to avoid it on the first element)

newdle/client/src/components/common/LanguageSelector.js
```diff
@@ -11,25 +11,19 @@ export default function LanguageSelector() {
   const language = useSelector(getLanguage);
   const dispatch = useDispatch();
 
-  if (window.location.hash !== '#i18n') {
```
ThiefMaster, line 14: There's another check like this somewhere else. You need to remove that one as well or keep it here. I would prefer the latter, since we currently do not have any complete translations, so showing it would just add noise...

newdle/client/src/components/Footer.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function ProjectInfo() {
+  return (
+    <div className={styles.info}>
+      <p className={styles.newdle}>
+        <Trans>
+          Powered by{' '}
+          <a href="https://github.com/indico/newdle" target="_blank" rel="noopener noreferrer">
+            newdle
```
Leats, line 58: Is writing `Powered by newdle` useful when we have the `newdle` name and logo in the title already? I feel like it does not really say much unless the header would be configurable (which isn't needed).
Maybe something like `newdle is Open Source` instead? What do others think about this?

newdle/client/src/components/Footer.js
VirtualEvan: I just did this, pretty neat trick! thanks for the suggestion

VirtualEvan:
@ThiefMaster When you mentioned in #242 about the links

> populated from a list/dict in the config file

Were you refering to somehow use `newdle.cfg` or use some other config file specific for this in the client `/src` folder via webpack?

ThiefMaster:
yep, newdle.cfg (ie you need to add another API that returns this)

newdle/client/src/components/Footer.js
```diff
@@ -0,0 +1,76 @@
...
+  return (
+    <footer className={styles.footer}>
+      <Grid columns={3}>
+        <Grid.Row verticalAlign="middle">
+          <Grid.Column computer={4} tablet={4} only="computer tablet">
+            <ProjectInfo />
+          </Grid.Column>
+          <Grid.Column mobile={16} tablet={8} computer={8}>
+            <ul className={styles.links}>
+              {Object.entries(links).map(([key, value]) => (
```
ThiefMaster, line 27: By using a dict for the link we don't keep any specific order. Maybe better to use a list with `[title, link]` pairs?

newdle/client/src/components/Footer.js
```diff
@@ -0,0 +1,76 @@
...
+
+export default function Footer() {
+  const [configuredLinks, loading] = client.useBackend(() => client.getFooterLinks(), []);
+
+  let links;
+  if (loading || configuredLinks === null) {
+    links = {};
+  } else {
+    links = configuredLinks;
+  }
```
ThiefMaster, line 16: Can't this be replaced with something as simple as this?

```js
const links = configuredLinks || {};
```

newdle/client/src/setupProxy.js
```diff
@@ -6,6 +6,7 @@ const proxy = require('http-proxy-middleware');
 module.exports = app => {
   const config = {target: process.env.FLASK_URL || 'http://127.0.0.1:5000'};
   app.use(proxy('/api/', config));
+  app.use(proxy('/config/', config));
```
ThiefMaster, line 9: Just put your new endpoint in `/api/config` - no need to change the proxy settings that way

newdle/config.py
```diff
@@ -0,0 +1,12 @@
+from flask import Blueprint, current_app
+
+
+config = Blueprint('config', __name__, url_prefix='/config')
+
+
+@config.route('/footer_links/')
```
ThiefMaster, line 7: Underscores in URLs are ugly ;) Let's use dashes instead.

newdle/config.py
```diff
@@ -0,0 +1,12 @@
...
+def footer_links():
+    if 'FOOTER_LINKS' in current_app.config:
```
ThiefMaster, line 9: Put an empty default value (plus documentation like we have for the other config settings) in `newdle.cfg.example` - that file is always loaded before the real config, so we know for sure that any config key is defined.

ThiefMaster, line 9: BTW, in case this check was necessary, `current_app.config.get('FOOTER_LINKS', {})` would be a more pythonic alternative to these 4 lines here

newdle/config.py
```diff
@@ -0,0 +1,12 @@
+from flask import Blueprint, current_app
+
+
+config = Blueprint('config', __name__, url_prefix='/config')
```
ThiefMaster, line 4: I don't think we need a new file/blueprint for this; this is just another API after all...

newdle/client/src/components/Footer.module.scss
```diff
@@ -0,0 +1,60 @@
...
+    li {
+      display: inline;
+
+      .link {
+        font-size: 14px;
+        color: $white;
+        opacity: 1;
+      }
+
+      + li ::before {
```
ThiefMaster, line 40: Shouldn't this be `& + li::before`?

`::before` is not a child element of the `li` but  but actually part of it

ThiefMaster (changes requested):

newdle/client/src/components/Footer.module.scss
VirtualEvan: Right!
About the ampersand, AFAIK it can be omitted, is there any reason to add it? readability maybe?

ThiefMaster: you are right, there's no need for `&` here, since the rule is `foo + bar` and not `foo+bar`

newdle/api.py
```diff
@@ -82,6 +82,12 @@ def require_token():
     g.user = user
 
 
+@api.route('/config/footer-links/')
```
ThiefMaster, line 85:
```suggestion
@api.route('/config/footer-links')
```

newdle/client/src/components/Footer.js
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <footer className={styles.footer}>
+      <Grid columns={3}>
+        <Grid.Row verticalAlign="middle">
+          <Grid.Column computer={4} tablet={4} only="computer tablet">
+            <ProjectInfo />
+          </Grid.Column>
+          <Grid.Column mobile={16} tablet={8} computer={8}>
+            <ul className={styles.links}>
```
ThiefMaster, line 21: does everything look alright when there are no links?

newdle/client/src/components/Footer.js
```diff
@@ -0,0 +1,71 @@
...
+              {links.map(([key, value]) => (
+                <React.Fragment key={key}>
```
ThiefMaster, line 23: Not needed anymore, you can use `<li key={key}>`

newdle/client/src/components/Footer.js
ThiefMaster: 👀  @VirtualEvan

newdle/newdle.cfg.example
```diff
@@ -97,3 +97,7 @@ LAST_ACTIVITY_CLEANUP_DELAY = None
 FINAL_DATE_CLEANUP_DELAY = None
 # Days after which a newdle that was deleted by its creator is cleaned up.
 DELETED_CLEANUP_DELAY = None
+
+# Footer content configuration
+# Specify which links are shown in the footer using ['title', 'link'] pairs
```
ThiefMaster, line 102:
```suggestion
# Specify which links are shown in the footer using `('title', 'url')` pairs
```

ThiefMaster:
Responsive view looks rather broken:

image

If there are no links defined, there's also some odd space above:

image

VirtualEvan:
I would say the first image is expected. It has sense if the i18n is enabled

image

I am looking on hidding the space when there are no links

ThiefMaster:
oh right, forgot about i18n 😳

newdle/client/src/components/Footer.js
VirtualEvan: My opinion is that "Powered by newdle" looks prettier :/
These are some of the ideas I tested

image
image
image

newdle/client/src/locales/es/messages.po
```diff
@@ -471,7 +471,7 @@ msgid "{hours}h {minutes}min"
 msgstr ""
 
 #: src/components/creation/timeslots/TimeslotsStep.js:21
-msgid "{numSlots, plural, one {<0>#</0> slot added} other {<1>#</1> slots added} =0 {You haven't added any slots yet}}"
+msgid "{numSlots, plural, one {<0>#</0> slot added } other {<1>#</1> slots added } =0 {You haven't added any slots yet}}"
```
ThiefMaster, line 474: If you get this change here, your lingui version is outdated. Running `make` should install the correct one (after that, run `make i18n` to force an update to this file).

newdle/client/src/components/Footer.js
```diff
@@ -0,0 +1,69 @@
...
+          <Grid.Column
+            {...(links.length === 0 ? {only: 'computer tablet'} : {mobile: 16})}
+            tablet={8}
+            computer={8}
+          >
+            <ul className={styles.links}>
+              {links.map(([key, value]) => (
+                <li key={key}>
+                  <a href={value} className={styles.link} target="_blank" rel="noopener noreferrer">
+                    {key}
```
ThiefMaster, line 29:
```suggestion
              {links.map(([title, href]) => (
                <li key={title}>
                  <a href={href} className={styles.link} target="_blank" rel="noopener noreferrer">
                    {title}
```

newdle/client/src/components/Footer.js
Leats: I like the "newdle is Open Source" one!

ThiefMaster: after checking in the team, everyone agrees that the "newdle is open source" one is the preferred message - so let's go for that one :)

PS: I think the `github.svg` icon can be removed sicne we no longer use it (unless there's a nice way of keeping it in the footer)

VirtualEvan: I added the change and rebased. Let me know if some other change is needed :)

btw: @ThiefMaster I just noticed your message is in all lowercase. Is it intended or should it be "newdle is Open Source" ?

ThiefMaster: `newdle is Open Source` is fine. I was just too lazy to use the shift key :D

ThiefMaster (approved):

PR #332, opened 2020-11-11
Fix removing participants

ThiefMaster:
Now it works for both existing and new participants

### DIFF
--- a/newdle/client/src/reducers/creation.js
+++ b/newdle/client/src/reducers/creation.js
@@ -84,7 +84,12 @@ export default combineReducers({
       case ADD_PARTICIPANTS:
         return [...state, ...action.participants];
       case REMOVE_PARTICIPANT:
-        return state.filter(p => p.id !== action.participant.id);
+        if (action.participant.id) {
+          return state.filter(p => p.id !== action.participant.id);
+        } else if (action.participant.auth_uid) {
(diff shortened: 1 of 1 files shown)


newdle/client/src/reducers/creation.js
```diff
@@ -84,7 +84,12 @@ export default combineReducers({
       case ADD_PARTICIPANTS:
         return [...state, ...action.participants];
       case REMOVE_PARTICIPANT:
-        return state.filter(p => p.id !== action.participant.id);
+        if (action.participant.id) {
+          return state.filter(p => !p.id || p.id !== action.participant.id);
```
plourenco, line 88: Why not just?
```suggestion
          return state.filter(p => p.id !== action.participant.id);
```

ThiefMaster: newly added participants from user search don't have one

plourenco: But wouldn't in that case `p.id` still be different from `action.participant.id`? As we never have a `undefined` `action.participant.id`.

ThiefMaster: It would be, and thus remove all those participants

ThiefMaster: ok nevermind, i was somehow thinking the other way round. you were right that the first half of that check isn't needed.

PR #358, opened 2021-06-11
Use Python 3.9 everywhere

ThiefMaster:


tomasr8:
Just a day after I had to figure out how to use python 3.8 alongside 3.9 :D

ThiefMaster:
`PYTHON=python3.9` is what I did to avoid installing 3.8 (even though it's as easy as `pyenv install 3.8.x`)

tomasr8:
> `PYTHON=python3.9` is what I did to avoid installing 3.8 (even though it's as easy as `pyenv install 3.8.x`)

I went the `pyenv` route in the end ;)

PR #360, opened 2021-06-21
Reword emails to avoid unnecessary pronoun usage

### LINKED ISSUE #348: Reword mails to avoid unnecessary pronoun usage
Email from a user:

> I've noticed a couple of times that Newdle uses "them" with an identified person, e.g.
>
> Tony Hawking would like you to help them choose a time...
>
> I appreciate why you are doing this, but more people are likely to be upset
> by this than thank you-for most people, you are imposing the wrong pronoun. Could you not reword the message as
>
>                 Please help Tony Hawking choose a time...
>
> And so avoid the pronoun issue completely?

While I think using "them" for cases where you do not know the gender/pronoun of someone (which is clearly the case when you only have a name) should be fine (it *is* supposedly neutral), rewording the message there to completely avoid pronouns is even better.

tomasr8:
Closes #348

### DIFF
--- a/tests/notifications_test.py
+++ b/tests/notifications_test.py
@@ -65,7 +65,7 @@ def test_deletion_email_plaintext(snapshot, comment, snapshot_name):
 def test_invitation_email_plaintext(snapshot):
     text = render_template(
         'invitation_email.txt',
-        creator='A random cat',
+        creator='Katy Purry',
         participant='Arthas Menethil',
         title='Sitting on keyboards',
         answer_link=url_for(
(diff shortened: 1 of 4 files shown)


tests/notifications_test.py
```diff
@@ -65,7 +65,7 @@ def test_deletion_email_plaintext(snapshot, comment, snapshot_name):
 def test_invitation_email_plaintext(snapshot):
     text = render_template(
         'invitation_email.txt',
-        creator='A random cat',
+        creator='a random cat',
```
ThiefMaster, line 68: Since names usually start with a capital letter I'd pick some name where this can be kept without looking wrong. "Catty McCatface"? :D

ThiefMaster (approved):

tests/notifications_test.py
tomasr8: Katy Purry it is

ThiefMaster: That name is purrfect. 😹

PR #364, opened 2021-07-21
Add a togglable grid view to the answer page

tomasr8:
Adds a separate grid view which can be used to answer the newdle.
The grid view is similar to that of the summary page but allows selecting answers.
It shows answers of all other participants and the availability of the current user.

On mobile, the grid view is disabled and the toggle button is not shown.

Screenshot from 2021-07-23 11-22-15
image


newdle/client/src/components/answer/AnswerGrid.js
```diff
@@ -0,0 +1,703 @@
...
+} from '../../selectors';
+import {serializeDate, toMoment} from '../../util/date';
+import AvailabilityRing from '../AvailabilityRing';
+import styles from './AnswerGrid.module.scss';
+
+function formatMeetingTime(startTime, duration) {
+  const endTime = moment(startTime)
+    .add(duration, 'm')
+    .format('HH:mm');
+  return `${serializeDate(startTime, 'HH:mm')} - ${endTime}`;
```
ThiefMaster, line 24: wouldn't it be cleaner to use serializeDate for both?

newdle/client/src/components/answer/AnswerGrid.js
```diff
@@ -0,0 +1,703 @@
...
+ * The red strip has absolute positioning with respect to the parent table cell and
+ * is styled using a combination of 'width' and 'left' or 'right' CSS rules.
+ *
+ */
+function getBusyTimesPositions(timeslot, busyTimes, duration) {
+  const start = toMoment(timeslot, moment.HTML5_FMT.DATETIME_LOCAL);
+  const styles = [];
+
+  Object.entries(busyTimes).forEach(([date, times]) => {
+    if (start.isSame(date, 'day')) {
```
ThiefMaster, line 41: What about using something like `busyTimes[start]` instead of the loop? or `.find()` on the entries if you need the momentjs logic (but just converting `start` to a date string would probably work as well)

newdle/client/src/components/answer/AnswerGrid.js
```diff
@@ -0,0 +1,703 @@
...
+    </div>
+  );
+}
+
+ParticipantGrid.propTypes = {
+  finalDate: PropTypes.string,
+  setFinalDate: PropTypes.func.isRequired,
+  isCreator: PropTypes.bool.isRequired,
+  finalized: PropTypes.bool.isRequired,
+};
```
ThiefMaster, line 703: 700 lines of code... maybe worth splitting this into multiple files?

newdle/client/src/components/answer/AnswerGrid.js
```diff
@@ -0,0 +1,703 @@
...
+          timeslots={timeslots}
+          interactive={!finalized}
+          finalDate={finalDate}
+          setFinalDate={setFinalDate}
+          isCreator={isCreator}
+          hoveredColumn={hoveredColumn}
+          setHoveredColumn={setHoveredColumn}
+        />
+        <Table.Body>
+          {_.sortBy(participants, ['name']).map(participant => (
```
ThiefMaster, line 673:
```suggestion
          {_.sortBy(participants, 'name').map(participant => (
```

AFAIK you do not need an array when sorting by a single field

ThiefMaster:
The bug you fixed in newdle#357 seems to be back w/ your changes here (view some newdle, go to my newdles, click create newdle)

ThiefMaster:
When a newdle is set to private, we now always try loading the participants (even in non-grid-view), even though the user cannot view them (causing an error to be displayed).

newdle/client/src/components/answer/AnswerGrid.js
tomasr8: Refactored to remove the loop :+1:

newdle/client/src/components/answer/AnswerGrid.js
tomasr8: I split it into `AnswerGrid` and `ParticipantGrid` with an extra file `GridCommon` containing shared components.

tomasr8:
> When a newdle is set to private, we now always try loading the participants (even in non-grid-view), even though the user cannot view them (causing an error to be displayed).

In this case, should we disable the grid view altogether? Since we have nothing to show there..

ThiefMaster:
Either that, or just show only one row... maybe the latter is less confusing of someone sets the grid view as their default so they don't suddenly get a different view?

PR #370, opened 2021-11-10
Add signature to UserSchema

tomasr8:
Fixes missing signature when using `Add myself`

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -62,6 +62,7 @@ def test_stats(flask_client):
     }
 
 
+@pytest.mark.usefixtures('mock_sign_user')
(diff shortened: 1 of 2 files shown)


tests/api_test.py
```diff
@@ -69,6 +69,8 @@ def test_me(flask_client, dummy_uid):
         'email': 'example@example.com',
         'name': 'Guinea Pig',
         'uid': 'user123',
+        'auth_uid': 'user123',
+        'signature': 'C5jy3r5ykNuVrBioWLSXYzjxIAo',
```
ThiefMaster, line 73: I'd use `mock_sign_user` instead like in the other tests involving APIs that return signed content

PR #393, opened 2022-04-20
Generate timezones dynamically to reflect CET/CEST changes

### LINKED ISSUE #369: The list of timezones should reflect summer times
The list of timezones currently does not reflect that Europe is not in CEST anymore. See screenshot below taken on 5.11. Such switches should be taken into account in the UI to make it less confusing to people.

Screenshot 2021-11-05 at 10 08 13

It seems that currently the list of timezones is static. Maybe it should come from server and be cached for some time.

tomasr8:
Closes #369 

The precompiled timezones did not take into account summer/winter time changes, always showing CEST.
This change adds ~20ms to the initial page load when the timezones have to be computed.

### DIFF
--- a/newdle/client/src/util/timezones.js
+++ b/newdle/client/src/util/timezones.js
@@ -1,13 +1,9 @@
+import _ from 'lodash';
+import moment from 'moment';
 import preval from 'preval.macro';
 import 'moment-timezone';
 
-// This will just generate a pre-compiled list with all timezones and their
-// respective offsets.
-export const commonTimezones = preval`
-  require('moment-timezone');
-  const moment = require('moment');
-  const _ = require('lodash');
-
+const commonTzSet = preval`
   const tzData = require('moment-timezone/data/meta/latest.json');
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Would it be possible to still precompute evreything but include DST and non-DST offsets (so we'd only need to pick the right one at runtime)?

tomasr8:
Since the DST time change differs for different timezones (and sometimes different years as I found out), the best I could come up with is this, which still takes ~10ms to run. What do you think?

```js
const timezones = precomputed.map(({ tz, name, offset, caption }) => {
  if(moment.tz(tz).isDST()) {
    return {
      name,
      offset: offset.dst,
      caption: caption.dst,
    };
  } else {
    return {
      name,
      offset: offset.std,
      caption: caption.std,
    };
  }
});
```

ThiefMaster:
Looks fine to me

ThiefMaster (approved):

PR #395, opened 2022-04-25
Add option to clone newdles

tomasr8:
Adds a simple form where you can choose to keep participants and/or timeslots. You can also move the timeslots to some other date, that is, the timeslots will start on the selected date. This form then takes you to a prefilled newdle creation page where you can proceed as usual when creating a newdle.

image
image
image

### DIFF
--- a/newdle/client/src/components/creation/ClonePage.js
+++ b/newdle/client/src/components/creation/ClonePage.js
@@ -25,6 +25,7 @@ export default function EditPage() {
   const isUserLoggedIn = useSelector(isLoggedIn);
   const newdle = useSelector(getCreatedNewdle);
   const history = useHistory();
+  const [keepSettings, setKeepSettings] = useState(true);
   const [keepParticipants, setKeepParticipants] = useState(true);
   const [keepTimeslots, setKeepTimeslots] = useState(true);
   const [option, setOption] = useState('keep');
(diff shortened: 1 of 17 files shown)


newdle/client/src/components/creation/ClonePage.js
```diff
@@ -0,0 +1,160 @@
...
+                  />
+                </div>
+              )}
+            </div>
+          )}
+        </div>
+      </div>
+      <div className={styles['create-button']}>
+        <Button color="violet" type="submit" disabled={!canSubmit} onClick={cloneNewdle}>
+          <Trans>Review newdle</Trans> 🍜
```
ThiefMaster, line 138: hm... "Review cloned newdle"? and probably worth showing some message on the next page that the new-dle hasn't been saved yet so people do not think they are already done? but we could also see if people realize it on their own first

newdle/client/src/components/summary/SummaryPage.js
```diff
@@ -211,11 +212,10 @@ export default function SummaryPage() {
           {isCreator && (
             <div className={styles['button-row']}>
               {gridViewActive && actions}
-              <Button
-                color="red"
-                className={styles['delete-button']}
```
ThiefMaster, line 216: Was this removed on purpose?

tomasr8: I just removed the class (which doesn't exist)

newdle/client/src/components/creation/ClonePage.js
tomasr8: I added an info message to the first step

ThiefMaster (approved):

PR #398, opened 2022-05-09
Add option so that each slot can only be selected once

### LINKED ISSUE #337: Option to allow only one choice of slot per participant
Pretty self-explanatory, and other similar systems seem to provide this option.

### LINKED ISSUE #385: Option so each slot can only be selected once
This is a feature some people miss from Doodle etc. It would cover cases where one does not want to find a single final date but rather let their participants each pick a single time slot where they are available, and then not letting anyone else pick the same time slot.

This would probably be good to combine with #337 since selecting multiple (or ifneedbe in general) slots would not make sense in such a mode.

When enabled, the following behaviors would change:

- ifneedbe is not used; only available/unavailable
- selecting a slot would deselect other slots (and maybe show a notification that this happened)
- "select final date" would be disabled and maybe replaced with a "finalize" button that (if enabled) notifies everyone with an available slot about their selected slot...

tomasr8:
Closes #337 
Closes #385 

This PR adds a new type of newdle in which every participant can only choose a single timeslot and cannot choose
a timeslot which has already been selected by someone else.

Quick summary:
- `ifneedbe` is not used with this type - only `[un]available`
- selecting a slot automatically deselects other slots
- slots which are taken are disabled (and marked as such) and participants cannot select them
- changing a newdle to a newdle with limited slots is not allowed to prevent situations where participants have already answered with multiple slots, `ifneedbe`'s, etc.. 
- instead of selecting a final date, you simply _finalize_ the newdle
- when sending emails, every participant gets and email with their chosen slot
- added labels to the newdle header which indicate that the newdle has limited slots (maybe we could find a better name for that?) and that it is private
- availability rings are {0,1}/1 since every timeslot can be chosen by only one participant
- the summary view shows only participants who selected the given timeslot (since everyone else is implicitly unavailable)

image

image
image

image

### DIFF
--- a/newdle/answers.py
+++ b/newdle/answers.py
@@ -1,0 +1,58 @@
+from collections import Counter
+
+from newdle.core.util import format_dt
+from newdle.models import Availability
+
+from .core.webargs import abort
+
+
+def validate_answers(newdle, participant, answers):
+    invalid = answers.keys() - set(newdle.timeslots)
+    if invalid:
+        abort(
+            422,
+            messages={
+                'answers': {
+                    format_dt(key): {'key': ['Invalid timeslot']} for key in invalid
+                }
+            },
+        )
+
+    if newdle.limited_slots:
+        # 'ifneedbe' is not allowed for newdles with limited slots
+        if Availability.ifneedbe in answers.values():
+            abort(
+                422,
+                messages={
+                    'answers': [
+                        "'if-need-be' is not allowed in newdles with limited slots"
+                    ]
+                },
+            )
+
+        # Must have at most one 'available'
+        counts = Counter(answers.values())
+        if counts[Availability.available] > 1:
+            abort(
+                422,
+                messages={'answers': ["At most one 'available' may be present"]},
+            )
+
+        # The slot must not be already taken
+        timeslot = next(
+            (
+                slot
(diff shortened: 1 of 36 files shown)


newdle/answers.py
```diff
@@ -0,0 +1,59 @@
...
+                    'answers': [
+                        "'if-need-be' is not allowed in newdles with limited slots"
+                    ]
+                },
+            }
+
+        # Must have at most one 'available'
+        counts = defaultdict(int)
+        for availability in answers.values():
+            counts[availability] += 1
```
ThiefMaster, line 34: I think using a `Counter` for this would be easier: `counts = Counter(answers.values())`

newdle/answers.py
```diff
@@ -0,0 +1,59 @@
...
+        timeslot = next(
+            slot
+            for slot, availability in answers.items()
+            if availability == Availability.available
+        )
+        for p in newdle.participants:
+            if p.code == participant.code:
+                continue
+
+            if p.answers.get(timeslot) == Availability.available:
```
ThiefMaster, line 51: `any(p.answers.get(timeslot) == Availability.available for p in newdle.participants if p.code != participant.code)`?

PR #401, opened 2022-05-23
Add option to export answers

### LINKED ISSUE #399: Export Newdle results in an Excel file
Would be nice to download the results of the Newdle in an Excel file for further processing.
My use case : I need to find 2 dates for the same training, and want to maximize the attendance.
So I need to be able to see which combination of dates gives the best results.

tomasr8:
Closes #399 

- [x] Implement csv/xlsx export in flask
- [x] Add option to export to the summary page
- [x] Add tests 

image
image


newdle/api.py
```diff
@@ -595,3 +605,33 @@ def send_deletion_emails(args, code):
...
+
+    if format == 'csv':
+        buffer = export_answers_to_csv(newdle)
+    else:
+        buffer = export_answers_to_xlsx(newdle)
+
+    mimetype = {
+        'csv': 'text/csv',
+        'xlsx': 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+    }[format]
```
ThiefMaster, line 630: I'd either but both in a mapping (`{'csv': ('text/csv', export_anwers_to_csv), ...}`) or set the mimetype in the if/else above.

plourenco (approved):
Looks good! Interesting approach with the blob...

PR #414, opened 2022-07-13
Update to react-router 6

ThiefMaster:
Draft since react-router 6 currently lacks some functionality we need: https://github.com/remix-run/react-router/issues/8139

Another open point (see the TODO in the code) is finding an alternative for the previously used `history.listen()` in the ErrorMessage component.


avivace:
I believe this should be unblocked now?

PR #428, opened 2022-10-10
Preserve user-selected timezone between creation of new newdles

### LINKED ISSUE #256: Default to the timezone from localStorage when creating a newdle
Just like when answering it makes sense to remember the last used timezone during creation as well.

We can either use the same localStorage key for it, or separate ones (not sure what's the better option)

mic4ael:
closes #256


newdle/client/src/components/common/TimezonePicker.js
```diff
@@ -37,7 +37,7 @@ function TimezonePicker({onChange, currentTz, title, ...props}) {
 TimezonePicker.propTypes = {
   onChange: PropTypes.func.isRequired,
   currentTz: PropTypes.string.isRequired,
-  title: PropTypes.string,
+  title: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
ThiefMaster, line 40: would `PropTypes.node` cover this?

mic4ael: Yup! That is much better :)

newdle/client/src/components/creation/timeslots/Timeline.js
```diff
@@ -370,9 +396,13 @@ export default function Timeline({date, availability, defaultMinHour, defaultMax
                 <div className={styles['config-box']}>
                   <DurationPicker />
                   <TimezonePicker
-                    onChange={value => dispatch(setTimezone(value))}
+                    onChange={value => {
+                      dispatch(setTimezone(value));
+
```
ThiefMaster, line 401: so much blank space here! 👾

ThiefMaster (commented):
Functionality-wise it works fine. But for that revert button we really need to find a different color ;)

newdle/client/src/components/creation/timeslots/Timeline.js
mic4ael: Haha :D I thought it would be better to separate these calls but I don't mind changing it :D

newdle/client/src/components/creation/timeslots/Timeline.js
```diff
@@ -355,6 +355,32 @@ export default function Timeline({date, availability, defaultMinHour, defaultMax
     setHourSpan(getHourSpan(input));
   }, [candidates, defaultHourSpan, defaultMaxHour, defaultMinHour, duration]);
 
+  const creationTimezone = localStorage.getItem('creationTimezone');
+  const revertIcon = (
+    <Popup
+      trigger={
+        <Icon
+          name="redo"
+          color="olive"
```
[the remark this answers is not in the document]
mic4ael: Honestly, I thought that it was better than the one I used at the beginning (green). How about something like this?

image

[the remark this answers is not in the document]
ThiefMaster: way too prominent IMHO... this is something people will probably rarely use. could we just stick with white/gray?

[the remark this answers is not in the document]
mic4ael: Indeed, that is a good point.

image

newdle/client/src/components/creation/timeslots/Timeline.js
```diff
@@ -355,6 +355,32 @@ export default function Timeline({date, availability, defaultMinHour, defaultMax
     setHourSpan(getHourSpan(input));
   }, [candidates, defaultHourSpan, defaultMaxHour, defaultMinHour, duration]);
 
+  const creationTimezone = localStorage.getItem('creationTimezone');
+  const revertIcon = (
+    <Popup
+      trigger={
+        <Icon
+          name="redo"
+          color="grey"
```
ThiefMaster, line 364:
```suggestion
          color="white"
```

The contrast with grey is pretty bad :/

Issue #447, opened 2023-06-25
[builder 6/7] RUN HUSKY_SKIP_INSTALL=1 make


salagula:
Hi, I have tried to run it on my W11 machine with 'docker compose -f docker-compose.development.yml up'

But I only get

```shell
[+] Running 0/2
 - flask-server Warning                                                                                            3.4s
 - react-server Error                                                                                              3.4s
[+] Building 15.1s (11/18)
 => [internal] load build definition from Dockerfile                                                               0.0s
 => => transferring dockerfile: 32B                                                                                0.0s
 => [internal] load .dockerignore                                                                                  0.0s
 => => transferring context: 35B                                                                                   0.0s
 => [internal] load metadata for docker.io/library/python:3.11                                                    15.0s
 => [builder 1/7] FROM docker.io/library/python:3.11@sha256:fe68f3194a1a6df058901085495abca83d8841415101366c3a4c6  0.0s
 => [internal] load build context                                                                                  0.1s
 => => transferring context: 14.84kB                                                                               0.1s
 => CACHED [stage-1 2/8] RUN set -ex &&  groupadd -r newdle &&  useradd -r -g newdle -m -d /newdle newdle          0.0s
 => CACHED [stage-1 3/8] RUN pip install uwsgi                                                                     0.0s
 => CACHED [builder 2/7] RUN curl -sL https://deb.nodesource.com/setup_16.x | bash -                               0.0s
 => CACHED [builder 3/7] RUN apt-get update && apt-get install -y nodejs                                           0.0s
 => CACHED [builder 4/7] ADD . /build/                                                                             0.0s
 => CACHED [builder 5/7] WORKDIR /build                                                                            0.0s
 => ERROR [builder 6/7] RUN HUSKY_SKIP_INSTALL=1 make                                                             10.0s
------
 > [builder 6/7] RUN HUSKY_SKIP_INSTALL=1 make:
#0 0.442 SETUP  Creating virtualenv
#0 9.880 SETUP  Installing top-level node packages
#0 9.882 /bin/bash: line 1: npm: command not found
#0 9.882 make: *** [Makefile:46: node_modules/.lastmake] Error 127
------
failed to solve: executor failed running [/bin/sh -c HUSKY_SKIP_INSTALL=1 make]: exit code: 2
```

PR #449, opened 2023-07-01
Fix docker setup

GovernmentPlates:
Closes: #447 

~~I did some initial testing with this, and found that whilst this fixes the issue, it also causes prettier (and estlint) to complain about the formatting of the react stuff (probably related to the nodejs update I did...). I'm open to testing this further.~~

Issues have been fixed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,8 @@ RUN set -ex && \
 	groupadd -r newdle && \
 	useradd -r -g newdle -m -d /newdle newdle
 
+# required packages for uwsgi to build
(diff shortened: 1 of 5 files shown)


GovernmentPlates:
Putting this on hold for now as this requires more testing.

newdle/client/.prettierrc.yaml
tomasr8, line 1: empty file?

GovernmentPlates: This was probably added-in by accident when I was testing some fixes - it has been removed.

tomasr8 (approved):

PR #456, opened 2023-11-09
Add CSV writer for answer exports

micsucmed:


newdle/export.py
```diff
@@ -27,10 +28,14 @@ def _generate_answers_for_export(newdle):
 
 def export_answers_to_csv(newdle):
     rows = _generate_answers_for_export(newdle)
-    csv = '\n'.join([','.join(row) for row in rows])
     buffer = BytesIO()
-    buffer.write(csv.encode('utf-8-sig'))
+    output = StringIO()
+    writer = csv.writer(output, dialect='unix')
```
tomasr8, line 33: A minor thing, but I'd prefer to only quote when necessary while the unix dialect quotes everything regardless.. I think you can use `csv.QUOTE_MINIMAL` for that

tests/api_test.py
```diff
@@ -1374,7 +1374,7 @@ def test_answer_export(snapshot, monkeypatch, flask_client, dummy_uid):
     snapshot.snapshot_dir = Path(__file__).parent / 'export'
     p1 = Participant.query.filter_by(code='part1').first()
     p1.answers = {datetime(2019, 9, 11, 14, 0): Availability.available}
-    p1.comment = 'Available comment'
+    p1.comment = 'Available, comment'
```
ThiefMaster, line 1377: I suggest using something more random where one looking at the test/snapshot won't think first why there's a status indicator there. `Hello,World` or something like that...

newdle/export.py
```diff
@@ -27,10 +28,14 @@ def _generate_answers_for_export(newdle):
...
+    writer.writerows(rows)
+    buffer.write(output.getvalue().encode('utf-8-sig'))
     buffer.seek(0)
+    output.close()
```
ThiefMaster, line 37: Check the `csv_text_io_wrapper` we use in Indico to avoid this mess with two separate buffers.

PR #477, opened 2024-04-03
Sticky participants column for answers and participants grids

### LINKED ISSUE #471: Participant column should sticky-scroll
When there are a lot of participants and dates it is hard to see who is available when in the grid view (see attached screenshot). While the calendar widget is already sticky it would also be nice if the list of participants (first column) would be sticky on scroll.

Screenshot 2024-03-12 at 14 10 20

micsucmed:
closes #471

### DIFF
--- a/newdle/client/src/components/GridCommon.module.scss
+++ b/newdle/client/src/components/GridCommon.module.scss
@@ -40,6 +40,19 @@
   overflow-x: auto;
   max-width: max-content;
 
+  th:first-child {
+    position: sticky !important;
+    left: 0;
+    z-index: 1;
+  }
+
+  tr:not(.spacer) td:first-child {
+    position: sticky !important;
+    left: 0;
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Looks nice, can we do the same for the heading when scrolling down with many participants?

ThiefMaster:
I'm not a big fan of the way it looks like now with a rather small scrollable area...

Latest commit from your branch:
image

First commit from your branch:
image


To me this feels like a downgrade from what we have before: At least to me scrolling the whole page feels nice - it's just that having the headers stick on top would make it nicer. But small scrollable areas on an otherwise big page are not great (except at wasting valuable screen space that's now unused)...

tomasr8:
^ We've discussed this with @micsucmed yesterday. Unfortunately it doesn't seem that easy to do :confused:

PR #494, opened 2024-10-22
Use mouse click to add a new time slot

### LINKED ISSUE #490: Make it easier to create timeslots
Currently, to create a time slot one must use the time picker which is not very efficient when you need to create
lots of time slots. Indico's Room Booking allows you to to drag your mouse to create a booking:

image

I think it'd be great if we had something like this in Newdle to be able to quickly create time slots.

renefs:
This issue resolves #490 .

1. When the timeline is clicked, a new slot with the default duration is added.
2. A tooltip is displayed on mouse hover to indicate the position in the timeline.
3. The granularity is set to 15 minutes, but it can be changed using the time picker.

https://github.com/user-attachments/assets/263ae6a0-4d8d-40ec-9d65-cc7462481874

### DIFF
--- a/newdle/client/src/components/answer/Calendar.js
+++ b/newdle/client/src/components/answer/Calendar.js
@@ -38,7 +38,7 @@ function calculateHeight(start, end, minHour, maxHour) {
     endMins = end.hours() * 60 + end.minutes();
   } else if (end.day() !== start.day()) {
     // the end of the slot is on another day so we will allow the slot to overflow
-    endMins = maxHour * 60 + end.minutes();
+    endMins = MAX_HOUR * 60 + end.hour() * 60 + end.minutes();
   }
 
   if (startMins < minHour * 60) {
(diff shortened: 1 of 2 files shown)


tomasr8:
Overall looks good! I just have some points regarding UX:

- Could we reuse the existing tooltip when hovering over the timeline? There are now two different tooltips which might look a bit off
obrazek

- The green area looks a bit different now, was that intentional?
before:
obrazek
after:
obrazek
Related to that, it's also more difficult to create parallel timeslots because of the reduced padding (there's only a tiny spot at the bottom where you can click):
obrazek


- Maybe we could use `cursor: pointer` to make it more obvious that you can click on the timeline?

- The tooltip renders when you hover the `+` button:
obrazek

As I mentioned [here](https://github.com/indico/newdle/issues/490#issuecomment-2402320014), maybe we could render a silhouette when simply hovering to make it even more obvious you're supposed to click on the timeline?

PR #500, opened 2024-12-12
Add left margin to timeline and fix colors in edit mode

renefs:
Screenshot 2024-12-12 at 13 50 35

1. Fix the timeline width in edit mode to avoid participant names to overlap the calendar.
2. Display the green background and the green border only in the editable area of the timeline, not in the participant availability area.


tomasr8:
I'd add some left padding/margin to the participant avatars when the screen is in tablet layout, otherwise it looks good :)

image

tomasr8 (approved):

tomasr8:
@ThiefMaster do you think we could merge this? I don't want to keep the ugly overflow in prod for longer than needed :smile:

PR #504, opened 2025-04-08
Add the Italian locale in `lingui` :it: 

tomasr8:


newdle/client/package.json
```diff
@@ -51,7 +51,6 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "add-locale": "lingui add-locale",
```
tomasr8, line 54: This command is now deprecated, it's enough to just add the locale in the lingui config

PR #511, opened 2025-05-22
Handle empty user and participant names

### LINKED ISSUE #502: Do not allow participants w/ empty name
Not sure how it happened in the first place and I'm pretty sure we reject it AT LEAST on the frontend, but if you end up with such a participant in the DB, it fails when trying to generate the avatar url:

```python
File "/usr/local/lib/python3.12/site-packages/newdle/schemas.py", line 130, in <lambda>
'api.user_avatar', payload=avatar_payload_from_participant(participant)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/site-packages/newdle/core/util.py", line 123, in avatar_payload_from_participant
'initial': participant.name[0].upper(),
~~~~~~~~~~~~~~~~^^^
IndexError: string index out of range
```

renefs:
- Handles empty names on avatars for users and participants.
- Adds a `name_not_empty` constraint in the participant model.

Fixes #502


ThiefMaster:
DB constraint is maybe a bit too much for it, how about just rejecting it in the schema?

```py
def not_empty(value):
    if not value:
        raise ValidationError('This field cannot be empty.')
```

newdle/migrations/versions/20250701_c1_set_interrogation_participant_names.py
```diff
@@ -0,0 +1,28 @@
...
+
+Revision ID: c1
+Revises: 30b5fe7b5949
+Create Date: 2025-07-01 12:14
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = 'c1'
```
tomasr8, line 11: Small nit, but the revision id should be generated by alembic. It seems like we don't have a script for this like we do in Indico, but you can run `alembic -c newdle/migrations/alembic.ini revision` which generates a new file and you can take the ID from there (also make sure to update this file's name). 

If it complains about missing script location (not sure why that is), you can add `script_location = newdle/migrations` to `alembic.ini`

newdle/migrations/versions/20250704_1127_cab5d47c1152_add_interrogation_participant.py
```diff
@@ -0,0 +1,28 @@
+"""Add interrogation participant
```
ThiefMaster, line 1: "interrogation"?

I think something like "Ensure participant names" would be clearer... (update the filename accordingly as well)

newdle/migrations/versions/20250704_1127_cab5d47c1152_add_interrogation_participant.py
```diff
@@ -0,0 +1,28 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        """
+        UPDATE participants
+        SET name = '?'
+        WHERE name IS NULL OR name = '';
```
ThiefMaster, line 22:
```suggestion
        WHERE name = '';
```

name is not nullable, so you cannot have NULL names there.

renefs:
I wonder if it's worth to add anything to the downgrade code like:
```python
def downgrade():
    op.execute(
        """
        UPDATE participants
        SET name = NULL
        WHERE name = '?';
        """
    )
```

Since this is fixing a "bug"... probably not?

newdle/schemas.py
```diff
@@ -131,6 +131,11 @@ class ParticipantSchema(mm.Schema):
         )
     )
 
+    @validates('name')
+    def name_not_empty(self, value):
+        if not value or not value.strip():
```
ThiefMaster, line 136:
```suggestion
        if not value:
```

We already strip surrounding whitespace [here](https://github.com/indico/newdle/blob/master/newdle/core/webargs.py#L19-L20), so you will never have whitespace-only names

ThiefMaster:
For the broken comment about the alembic revision here in this PR, can you try removing these 3 lines?

https://github.com/indico/newdle/blob/master/ci/print_revision_sql.sh#L14-L16

Not sure why they were needed...

tomasr8:
Thanks @renefs!!

